PR #8, opened 2016-11-05
Parse variables and operationName for server example

amozoss:
Using the mutation and variables from http://graphql.org/learn/queries/#mutations in GraphiQL wasn't working in the example server. 

Variables (logging the type, key, and value from json.Unmarshal):
```
type: string k: ep v: JEDI
type: map[string]interface {} k: review v: map[stars:5 commentary:This is a great movie!]
--type: float64 k: stars v: 5
--type: string k: commentary v: This is a great movie!
```

`json.Unmarshal`defaults to float64 for the type of stars, so it was causing a panic:

```
2016/11/05 15:54:14 http: panic serving [::1]:62941: interface conversion: interface is float64, not int
```

So I adjusted `internal/exec/scalar.go` to cast `float64` to `int32`, so the example will work with the way I've unmarshaled the json. Casting from float64 to int32 seems harmless since the type was specified as `Int` in graphql, but perhaps there is a better way.

### DIFF
--- a/example/starwars/server/server.go
+++ b/example/starwars/server/server.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"io/ioutil"
 	"log"
 	"net/http"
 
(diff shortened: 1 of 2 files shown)


neelance:
Hey @amozoss, I've implemented the necessary changes on `master`. Does that look good to you?

amozoss:
Looks great, thanks!

PR #10, opened 2016-11-06
lexer.TestLexer and lexer.ConsumeFloat fix

piotrkowalczuk:
Small contribution for a good start.

Btw, any chance that this project will be versioned using http://semver.org ?

### DIFF
--- a/example/starwars/starwars.go
+++ b/example/starwars/starwars.go
@@ -25,7 +25,6 @@ var Schema = `
 		character(id: ID!): Character
 		droid(id: ID!): Droid
 		human(id: ID!): Human
-		humans(height: Float): [Human]!
 		starship(id: ID!): Starship
 	}
 	# The mutation type, represents all updates we can make to our data
@@ -346,16 +345,6 @@ func (r *Resolver) Human(args *struct{ ID graphql.ID }) *humanResolver {
 		return &humanResolver{h: h}
 	}
 	return nil
-}
-
-func (r *Resolver) Humans(args *struct{ Height *float64 }) []*humanResolver {
-	var res []*humanResolver
-	for _, h := range humans {
-		if args.Height != nil && h.Height == *args.Height {
-			res = append(res, &humanResolver{h: h})
-		}
-	}
-	return res
 }
 
 func (r *Resolver) Droid(args *struct{ ID graphql.ID }) *droidResolver {
--- a/example/starwars/starwars.go
+++ b/example/starwars/starwars.go
@@ -25,6 +25,7 @@ var Schema = `
 		character(id: ID!): Character
 		droid(id: ID!): Droid
 		human(id: ID!): Human
+		humans(height: Float): [Human]!
 		starship(id: ID!): Starship
 	}
 	# The mutation type, represents all updates we can make to our data
@@ -197,6 +198,15 @@ var humans = []*human{
 		Height:    1.8,
 		Mass:      0,
 	},
+	{
+		ID:        "1005",
+		Name:      "Augwynne Djo",
+		Friends:   []graphql.ID{"1002", "1003"},
+		AppearsIn: []string{"THELASTJEDI"},
+		Height:    2, // dono ;p
+		Mass:      60,
+		Starships: []graphql.ID{},
+	},
 }
 
 var humanData = make(map[graphql.ID]*human)
@@ -336,6 +346,16 @@ func (r *Resolver) Human(args *struct{ ID graphql.ID }) *humanResolver {
 		return &humanResolver{h: h}
 	}
 	return nil
+}
+
+func (r *Resolver) Humans(args *struct{ Height *float64 }) []*humanResolver {
+	var res []*humanResolver
+	for _, h := range humans {
+		if args.Height != nil && h.Height == *args.Height {
+			res = append(res, &humanResolver{h: h})
+		}
+	}
+	return res
 }
 
 func (r *Resolver) Droid(args *struct{ ID graphql.ID }) *droidResolver {
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -40,6 +40,79 @@ func (r *timeResolver) AddHour(args *struct{ Time time.Time }) time.Time {
 	return args.Time.Add(time.Hour)
 }
 
+type valueCoercionResolver struct {
+	asBool   *bool
+	asFloat  *float64
+	asInt    *int32
+	asString *string
+}
+
+func (r *valueCoercionResolver) AsInt() *int32 {
+	return r.asInt
+}
+
+func (r *valueCoercionResolver) AsString() *string {
+	return r.asString
+}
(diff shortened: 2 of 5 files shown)


internal/lexer/lexer.go
```diff
@@ -80,7 +80,12 @@ func (l *Lexer) ConsumeInt() int {
 
 func (l *Lexer) ConsumeFloat() float64 {
 	text := l.sc.TokenText()
-	l.ConsumeToken(scanner.Int)
+	switch l.next {
```
neelance: I think this should simply be `l.ConsumeToken(scanner.Float)`, since the token kind already gets checked before this method is called.

internal/lexer/lexer_test.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+	"reflect"
+	"strings"
+	"testing"
+	"text/scanner"
+
+	"github.com/neelance/graphql-go/internal/lexer"
+)
+
+func TestLexer(t *testing.T) {
```
neelance: This test looks overly complicated. I also think that we should test the public API of the `graphql` package instead of this internal API. Could you come up with some failing test there that gets fixed by your change to `ConsumeFloat`?

piotrkowalczuk: 1: completely agree, but wanted to play a little bit around
2: `array of floats` was failing before

internal/lexer/lexer.go
piotrkowalczuk: When replaced by `scanner.Float` gives an error anyway:

```
--- FAIL: TestLexer_ConsumeFloat (0.00s)
    --- FAIL: TestLexer_ConsumeFloat/zero (0.00s)
    	lexer_test.go:52: unexpected error: graphql: syntax error: unexpected "0", expecting Float (line 1, column 1)
    --- FAIL: TestLexer_ConsumeFloat/max-int64 (0.00s)
    	lexer_test.go:52: unexpected error: graphql: syntax error: unexpected "9223372036854775807", expecting Float (line 1, column 1)
    --- FAIL: TestLexer_ConsumeFloat/max-float64 (0.00s)
    	lexer_test.go:52: unexpected error: graphql: syntax error: unexpected "179769313486231570814527423731704356798070567525844996598917476803157260780028538760589558632766878171540458953514382464234321326889464182768467546703537516986049910576551282076245490090389328944075868508455133942304583236903222948165808559332123348274797826204144723168738177180919299881250404026184124858368", expecting Float (line 1, column 1)
```

neelance:
Sorry, but now you're duplicating https://github.com/golang/go/blob/master/src/strconv/atof_test.go. Let's only have a simple addition to https://github.com/neelance/graphql-go/blob/master/graphql_test.go that covers this fix.

piotrkowalczuk:
It was not possible to reproduce completely this issue by testing public API. Function is guarded by `common.ParseValue`.

neelance:
If it would not be possible to reproduce the issue with the public API, then it would be not a valid issue. Only the public API is the contract to consumers of the library, so only bad behavior by the public API qualifies as a bug. In this case however I think it should be possible to construct such a test on the public API (but not within the Star Wars example, since it has no input value with type `Float`.

neelance:
About semver: Maybe after the library has reached a stable release.

piotrkowalczuk:
Maybe I should be more precise. Core issue has been reproduced. But situation where `lever.ConsumeFloat` is called on `Int` was not.

neelance:
> But situation where lexer.ConsumeFloat is called on Int was not.

There is no bug in this situation. `ConsumeFloat` is allowed to panic in this case.

example/starwars/starwars.go
```diff
@@ -25,6 +25,7 @@ var Schema = `
 		character(id: ID!): Character
 		droid(id: ID!): Droid
 		human(id: ID!): Human
+		humans(height: Float): [Human]!
```
neelance: `[Human!]!` would be better. Also I think it would be better to make this test separate from the Star Wars example for two reasons:
1. I expect there will be some more value coercion tests coming up.
2. I don't want to bloat the star wars schema with them, the schema should stay in line with what is used in the GraphQL documentation.

piotrkowalczuk:
Any feedback?

neelance:
This PR is not yet ready to be merged. We can either iterate on it or I can implement fix and test separately if you prefer that.

piotrkowalczuk:
I don't mind to iterate here.

graphql_test.go
```diff
@@ -1135,3 +1207,76 @@ func TestTime(t *testing.T) {
 		},
 	})
 }
+
+func TestValueCoercion(t *testing.T) {
```
neelance: Those tests look better, but why so complicated with the `Result` type and an array of it? Wouldn't top level fields suffice to test what we want to test?
```
type Query {
	coerceFloat(arg: Float): Float
	[...]
}
```

example/starwars/starwars.go
```diff
@@ -197,6 +197,15 @@ var humans = []*human{
 		Height:    1.8,
 		Mass:      0,
 	},
+	{
```
neelance: This character is not necessary any more, right?

internal/exec/scalar.go
```diff
@@ -36,15 +36,24 @@ var intScalar = &scalar{
 			}
 			return coerced, nil
 		default:
-			return nil, fmt.Errorf("wrong type")
+			return nil, fmt.Errorf("wrong type: %T", input)
```
neelance: Why do you add the type here? It is already included in the error message by the code that calls this. Have you ever tested it?

internal/exec/scalar.go
```diff
@@ -36,15 +36,24 @@ var intScalar = &scalar{
...
 		}
 	},
 }
 var floatScalar = &scalar{
 	name:        "Float",
 	reflectType: reflect.TypeOf(float64(0)),
 	coerceInput: func(input interface{}) (interface{}, error) {
-		return input, nil // TODO
+		switch in := input.(type) {
```
neelance: Those 4 options look fine. 👍

internal/lexer/lexer.go
neelance: This still has no need of `l.ConsumeToken(scanner.Int)`.

internal/lexer/lexer_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"strings"
+	"testing"
+	"text/scanner"
+
+	"github.com/neelance/graphql-go/internal/lexer"
+)
+
+func TestLexer_ConsumeFloat(t *testing.T) {
```
neelance: Please remove this test.

neelance:
I have implemented a fix for this bug, see 86f0f14544112d824efa3d586f47ccc1b04a1d26. Closing this PR.

Issue #12, opened 2016-11-07
Pagination panics in example


amozoss:
The panic seems to happen when the last node is requested from the connection
```
diff --git a/graphql_test.go b/graphql_test.go
index 7b64c6e..8853a63 100644
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -666,6 +666,23 @@ func TestConnections(t *testing.T) {
                                                                }
                                                        }
                                                }
+                                       },
+                                       moreFriends: hero {
+                                               name
+                                               friendsConnection(first: 1, after: "Y3Vyc29yMg==") {
+                                                       totalCount
+                                                       pageInfo {
+                                                               startCursor
+                                                               endCursor
+                                                               hasNextPage
+                                                       }
+                                                       edges {
+                                                               cursor
+                                                               node {
+                                                                       name
+                                                               }
+                                                       }
+                                               }
                                        }
                                }
                        `,
@@ -689,6 +706,25 @@ func TestConnections(t *testing.T) {
                                                                }
                                                        ]
                                                }
+                                       },
+                                       "moreFriends": {
+                                               "name": "R2-D2",
+                                               "friendsConnection": {
+                                                       "totalCount": 3,
+                                                       "pageInfo": {
+                                                               "startCursor": "Y3Vyc29yMw==",
+                                                               "endCursor": "Y3Vyc29yMw==",
+                                                               "hasNextPage": false
+                                                       },
+                                                       "edges": [
+                                                               {
+                                                                       "cursor": "Y3Vyc29yMw==",
+                                                                       "node": {
+                                                                               "name": "Leia Organa"
+                                                                       }
+                                                               }
+                                                       ]
+                                               }
                                        }
                                }
                        `,
```
```
2016/11/07 12:26:15 graphql: graphql: panic occured: runtime error: makeslice: len out of range
goroutine 89 [running]:
github.com/neelance/graphql-go/internal/exec.(*request).handlePanic(0xc4201d8840)
        /Users/dan/go/src/github.com/neelance/graphql-go/internal/exec/exec.go:384 +0x123
panic(0x2ce460, 0xc4201ecab0)
        /usr/local/Cellar/go/1.7/libexec/src/runtime/panic.go:458 +0x243
github.com/neelance/graphql-go/example/starwars.(*friendsConnectionResolver).Edges(0xc4201f2c00, 0x0)
        /Users/dan/go/src/github.com/neelance/graphql-go/example/starwars/starwars.go:591 +0x65
```

[...]

PR #13, opened 2016-11-07
starwars example: fix pagination panic (#12)

amozoss:


example/starwars/starwars.go
```diff
@@ -572,8 +572,8 @@ func newFriendsConnectionResolver(ids []graphql.ID, args *friendsConenctionArgs)
 	if args.First != nil {
 		to = from + int(*args.First)
 	}
-	if to > len(ids)-from {
-		to = len(ids) - from
+	if to > len(ids) {
```
neelance: Test and bugfix look good. Minor suggestion: Can we move this `if` one line up? Because it is only relevant if `args.First` is set.

PR #16, opened 2016-11-10
Fix for ServeHTTP not handling strings for 'variable' variable

saward:
Hi,

Testing out a graphql endpoint using neelance/graphql-go produced errors when using https://github.com/skevy/graphiql-app, but not another Chrome graphiql extension.  The difference was that the former would send "" for variables.  I discovered that even when I added variables in, the server would produce errors.

With this commit, I'm able to leave variables empty (with "" or null) or include them, and the server will parse them correctly.

Also added checks for request type, and grab the values from the appropriate place depending on whether the request is GET or not.

### DIFF
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -50,16 +50,28 @@ type Handler struct {
 
 func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	var params struct {
-		Query         string                 `json:"query"`
-		OperationName string                 `json:"operationName"`
-		Variables     map[string]interface{} `json:"variables"`
+		Query         string `json:"query"`
+		OperationName string `json:"operationName"`
+		Variables     string `json:"variables"`
 	}
+
 	if err := json.NewDecoder(r.Body).Decode(&params); err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
-	response := h.Schema.Exec(r.Context(), params.Query, params.OperationName, params.Variables)
+	var parsedVariables map[string]interface{}
+
+	if len(params.Variables) > 0 {
+		if err := json.Unmarshal([]byte(params.Variables), &parsedVariables); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	} else {
+		parsedVariables = make(map[string]interface{})
+	}
+
+	response := h.Schema.Exec(r.Context(), params.Query, params.OperationName, parsedVariables)
 	responseJSON, err := json.Marshal(response)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -55,12 +55,18 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Variables     string `json:"variables"`
 	}
 
-	if err := json.NewDecoder(r.Body).Decode(&params); err != nil {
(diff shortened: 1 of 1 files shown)


neelance:
Hi, first thanks for contributing!

The `graphql/relay` package is called `relay` because it is meant to work with the Relay client, which it does. If some other client does not speak the same protocol, then it is not the Relay protocol. Thus I would say that those other protocols don't belong into a package called `relay`. Also, as far as I can see your change is even breaking the Relay client, since it does not accept the non-string version any more.

I am aware that the default setup of GraphiQL is sending the variables as a string. That's why I have added this line to make it work: https://github.com/neelance/graphql-go/blob/master/example/starwars/server/server.go#L46

I see a common pitfall here. What about adding documentation about what exactly the `relay` package expects and also adding a hint to the error message on a JSON parse error that one should take a look at such documentation?

saward:
Ah yes, I hadn't realised that the relay package was for the Relay client.  I haven't yet looked at the Relay client, but I will be soon.  I think best to close this for now until I've had a better look at Relay.

I think at the least a hint on the JSON parse error message is a good idea.

bsr203:
@neelance is it possible to push the relay example, in a separate branch if still not working yet. It will save lot of time when trying to port my app. thanks in advance.

shurcooL:
> What about adding documentation about what exactly the relay package expects and also adding a hint to the error message on a JSON parse error that one should take a look at such documentation?

That sounds like a great idea! 👍 Currently, the package has no documentation:

https://godoc.org/github.com/neelance/graphql-go/relay

neelance:
@shurcooL fyi: I've started to write some documentation in the readme: https://github.com/neelance/graphql-go/blob/master/README.md

saward:
@bsr203 if you found my changes useful for your needs, you can just put the modified ServeHTTP function somewhere in your main application, and use that instead of the relay one.  You can find the full function here: https://github.com/saward/graphql-go/blob/master/relay/relay.go

bsr203:
Hi @saward thanks for the pointers. We discussed about a complete relay example as you see at https://github.com/neelance/graphql-go/issues/2 . It is quite impressive how succinct the query and resolvers in the examples, and I am sure @neelance would have better idea on how to structure some of the quirks to support relay (global id, node interface, connection, ..). I just have the feeling it may better to see the example before I port my app from https://github.com/graphql-go/graphql . I will look at your changes, and thanks for pointing it out. Cheers.

saward:
@bsr203 I haven't tested this with relay yet, but maybe something like this will work for the global id and node interface parts:

```go
var Schema = `
schema {
	query: Query
}

interface Node {
	id: ID!
}

type Query {
	node(id: ID!): Node
}

type Property implements Node {
	id: ID!
}
`

type NodeResolver interface {
	ID() graphql.ID
	ToProperty() (*PropertyResolver, bool)
}

type nodeBase struct{}

func (nb nodeBase) ToProperty() (*PropertyResolver, bool) {
	return nil, false
}

type PropertyResolver struct {
	nodeBase
	Property someOtherPropertyStructReturnedByGetProperty
}

func (r *PropertyResolver) ID() graphql.ID {
	return graphql.ID("property:" + r.Property.ID)
}

func (r *Resolver) Property(args *struct{ ID graphql.ID }) *PropertyResolver {

	parts := strings.Split(string(args.ID), ":")

	property, err := GetProperty(parts[1])

	if err != nil {
		return nil
	}

	return &PropertyResolver{Property: property}
}
```

bsr203:
Thank you @saward. going to try out whenever I get around it.

PR #21, opened 2016-11-17
fixes #19

### LINKED ISSUE #19: Array as argument fails
Looking at the graphql spec, this is no indication that this is officially supported or not but I have a use case where I would like to pass an array as an argument to a resolver. Currently, I'm getting a TODO panic error when I do this:
https://github.com/neelance/graphql-go/blob/3b7efd5cb6e73337e26a67ca05750b6e1b02320e/internal/exec/exec.go#L315

As far as I can tell, that type switch needs to be able to handle a `*common.List` item.

Here is my use case:
```graphql
{
  approvals(status: [PENDING, APPROVED]) {
    type
    title
    amount
    attachments
    submitter {
      firstName
      lastName
      title
    }
  }
}
```

nicksrandall:


nicksrandall:
This fix is not complete. I'm still getting an error message when I have an input type that has a field that is an array (in my case a `[]graphql.ID`). The specific panic message that I get is:

```
graphql: graphql: panic occured: reflect.Set: value of type []interface {} is not assignable to type []graphql.ID
```

neelance:
packer == function that packs the generic data types (`map[string]interface{}` and `[]interface{}`) into concrete Go types like structs and slices of concrete types.

I think you can not use a `valuePacker` for the list, you'll need a new packer type instead that properly packs each value.

Hope this helps. :)

nicksrandall:
Closing this because I know it won't work

Issue #28, opened 2016-11-24
Use field as resolver instead of method. 


nicksrandall:
Would it be possible to use a struct field as a resolver instead of always using methods? Most of my methods are just simple getters so it would be convenient to allow resolver to default to field. Then I could implement method only when I need special logic.

lpalmes:
That would be really useful, maybe it could use introspection and struct tags like `json:"name"`, and a resolver method should override the field.

bsr203:
not sure how much performance penalty we pay for using reflection on this. I do codegen to write these simple resolvers.

lpalmes:
We should test and get some data, that would get us some information. I will do my best to try to get some benchmarks.

nicksrandall:
The amount of reflection that this implementation uses is probably my biggest concern. I do love how minimal the API is though. 

@bsr203 I'd love to see your code gen tools if you'd be willing to share.

neelance:
I have two concerns about this:

1. Simplicity.
I think that the following happened a lot during the design of the Go language:
Is it possible? Yes.
Would it give the option to write less code? Yes.
Should we add it? Probably not.
Adding this feature does not enable us to do something that we could not do before. It is just a second way to get to the same outcome. Imho this is something that is usually avoided in the Go world. Simplicity is preferred.

2. Tight coupling between resolver and underlying data structures.
Imho resolvers are a view on the underlying data structures. Those data structures can change without changing the GraphQL schema. The resolver can provide legacy fields. By directly accessing struct fields, this distinction becomes less obvious. This can cause the data structures to become more rigid, because people don't want to affect the schema. I think keeping this strong separation makes your GraphQL codebase more healthy in the long term.

bsr203:
@nicksrandall I agree with neelance fully, and used codegen during the initial migration. The code is tied with our internal stuff, and not intend to use for general purpose. I could highlight how we do it.

for a type `Widget` keep a convention that `WidgetResolver` is the graphql resolver for it.
there are many codegen example available in github, and there are some generic one. Use one of it, and get the Struct and Field info for `Widget`. Now, you may skip some of the fields with struct tags, and rest, you can generate simple mappers. HTH

```
// Writer is a structure to hold accumulated output
type Writer struct {
	Buf bytes.Buffer
}

// Printf writes the formatted string to writer
func (g *Writer) Printf(format string, args ...interface{}) {
	g.Buf.WriteString(fmt.Sprintf(format, args...))
}

[...]

Issue #17, opened 2016-11-11
No way to get requested fields inside resolver


DenisNeustroev:
**Problem:**
Inefficient microservice/database queries.
If I need to return 100 orders with their clients, I have to make 101 requests to microservices (1 for batch of orders, 100 for their clients (one per each order)).

**Scheme (simplified):**
```graphql
type Query {
  orders: [Order]
}

type Order {
  id: ID
  client: Client
}

type Client {
  id: ID
  name: String
}
```

**Query:**
```graphql
query GetOrders {
  orders {
    client {
      name
    }
  }
}
```

**Resolvers:**
```go
func (r *Resolver) Orders() []*orderResolver {
	data, err := Request(orderService, "Order.List", &Args{}) // returns batch
	...
	return // batch of 100 order resolvers
}

// 100 client resolvers - 100 requests
func (r *orderResolver) Client() *clientResolver {
	data, err := Request(clientService, "Client.Get",&Args{) // returns one 
	...
	return // a client resolver
}
```

Instead, I'd like to know (in the order resolver) that the field with clients has been requested.
So I could make one batch request at the order resolver level.

Something like this:
```go
func (r *Resolver) Orders(ctx context.Context) []*orderResolver {
	data, err := Request(orderService, "Order.List", &Args{}) // returns batch
	if ctx.Value("fields")... {
		  clients, err := Request(clientService, "Client.List",&Args{) // batch
        }
	...
	return // batch of 100 order resolvers
}
```

So I get 2 request.

neelance:
This is definitely a feature we should think about. Is there anything like that in the graphql-js implementation?

I am a bit worried that this can cause people to write bad resolvers. One nice attribute of GraphQL are all those consistency guarantees it provides.

nicksrandall:
I agree that this could cause people to write bad code resolvers, but it could also help write efficient resolvers. For example, one could use that information to prevent an unnecessary "join" in his/her db query if the user didn't ask for any of those fields. Also, it could give user the ability to log which fields are being requested most frequently and then perform optimizations on those fields (like pulling from redis-cache rather than db). 

The last variable passed to graphql resolvers in graphql-js has lots of info about the schema, fields, ect... 

see: https://github.com/graphql/graphql-js/blob/master/src/type/definition.js#L489

neelance:
@nicksrandall Do you know any example code that uses this feature with graphql-js?

nicksrandall:
[...]

PR #40, opened 2016-12-08
implement sql.Scanner and driver.Value for graphql.ID type

nicksrandall:
This makes `graphql.ID` type implement [sql.Scanner](https://golang.org/pkg/database/sql/#Scanner) and [driver.Valuer](https://golang.org/pkg/database/sql/driver/#Valuer). This simplifies the data structures that I'm using so that they can implement the types that graphql likes but can also be marshaled into the DB.

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -6,6 +6,7 @@ import (
 	"database/sql/driver"
 	"encoding/json"
 	"fmt"
+	"strconv"
 
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/opentracing/opentracing-go/ext"
@@ -40,6 +41,12 @@ func (id *ID) Scan(src interface{}) error {
 	switch t := src.(type) {
 	case string:
 		*id = ID(t)
+	case int64:
+		*id = ID(strconv.FormatInt(t, 10))
+	case []byte:
+		*id = ID(t)
+	case nil:
+		// do nothing
 	default:
 		return fmt.Errorf("wrong type")
 	}
--- a/graphql.go
+++ b/graphql.go
@@ -2,6 +2,8 @@ package graphql
 
 import (
 	"context"
+	"database/sql"
+	"database/sql/driver"
 	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


nicksrandall:
@neelance @bsr203 What do you think of this small change? Do you think it would be useful to others?

bsr203:
@nicksrandall sorry, I don't use sql backend but could see it may useful for others.

neelance:
What about this: Have your own `type ID string` and give it those methods. Use it in your struct. Then cast to `graphql.ID` in the resolver of the `id` field.

nicksrandall:
Having to always cast to graphql.ID is what I'm trying to avoid. Currently my structs use strings for ID prop and I have to cast to graphql.ID in resolvers. I use strings for the DB. This change would allow me to not have to worry about casting in all the right places. 

This also allows me to use the same struct for resolvers as input type because input type requires ID type (to be idiomatic). 

I imagine that anybody using this lib with a SQL backend would find this helpful. Am I wrong? 

What are your concerns with adding this?

shurcooL:
> I imagine that anybody using this lib with a SQL backend would find this helpful. Am I wrong?

Would everyone who uses this library and a SQL backend want this exact implementation, or would their needs differ slightly?

nicksrandall:
Anybody using the sql package in the standard lib would need this exact implementation.

neelance:
I still think that it is beneficial to have a clear separation between resolver API and underlying data structures. The resolvers are a view/projection on the underlying data. To me it feels like your change would mix this up a bit.

On the other hand I see that there currently is a strict 1-to-1 coupling between the GraphQL type `ID` and the Go type `graphql.ID`. This is inflexible and contradicts the freedom you have with resolver types.

One option would be to accept any named `string` type as a GraphQL `ID`. But that would also allow a pure `string`, which I dislike.

There might be some good solution around custom resolvers. You could use any type as the result type for `ID`, but that type would need to provide a method to turn the value into something GraphQL understands. The type would also check at startup if it is compatible with the given GraphQL type. The same method could be used for `Time` and for custom behavior like #28 and #17. I have to think about it some more.

nicksrandall:
What I have ended up doing is adding my own custom scalar type `ID` that is a named string that implements the sql methods that I need.

neelance:
@nicksrandall Please take a look at 2b513d7e6c46b01de0cbcaeb1b68424dd3c043be and https://github.com/neelance/graphql-go/issues/3#issuecomment-267831350. This should be a proper solution here.

nicksrandall:
I like it! Thanks @neelance. I close this PR

bedecost:
@nicksrandall @neelance First of all, thank you for your work!!

Sorry to dig in the past, but could you please explain me how I can apply the proposed solution to my case? (I'm using SQL with IDs defined as integers)

As a new gopher, I'm not quite sure where and how I should implement this..

bedecost:
I ran into the same situation regarding the scalar Time. I solved it at the moment by editing `graphql-go/time.go` (BUT this should be **avoided** of course!!):
```
// Value implements the driver Valuer interface
func (t Time) Value() (driver.Value, error) {
	return t.Time, nil
}

// Scan implements the Scanner interface.
func (t *Time) Scan(src interface{}) error {
	return t.UnmarshalGraphQL(src)
}
```
This offers me the possibility to use this same struct for GraphQL and GORM:
```
type Question struct {
	ID       int32  `gorm:"primary_key"`
	Question string `gorm:"not null"`
	...
	CreatedAt graphql.Time
}
```
However, I'd really like to apply your proposition (using `UnmarshalGraphQL` I suppose), but I don't understand how to..

PR #45, opened 2016-12-20
fix wrong import statement

### LINKED ISSUE #44: Weird dependency
Is this line correct? I feel like you meant to inject the internal github package? Its currently breaking my build.

https://github.com/neelance/graphql-go/blob/master/internal/exec/packer.go#L11

nicksrandall:
@neelance I could be wrong but I think that you meant to do this...

`go-import`s is awesome but sometimes it's wrong 😃 

fixes #44

### DIFF
--- a/internal/exec/packer.go
+++ b/internal/exec/packer.go
@@ -6,9 +6,9 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/neelance/graphql-go/errors"
 	"github.com/neelance/graphql-go/internal/common"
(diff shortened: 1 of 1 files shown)


neelance:
Yes, thanks a lot! No idea how that happened.

shurcooL:
Both packages have a func named `Errorf` (see [here](https://godoc.org/github.com/pkg/errors#Errorf) and [here](https://godoc.org/github.com/neelance/graphql-go/errors#Errorf)), and `goimports` prefers shorter import paths when picking between one of many possible matches.

Issue #37, opened 2016-12-08
Ability to return errors with extra fields


F21:
According to the [spec](http://facebook.github.io/graphql/#sec-Errors), errors must contain a `message` key. However, the developer can add other fields if he wants.

In my case, for each error, I want to return both a `message` which is developer friendly, and an `error_code` such as `USERNAME_UNAVAILABLE`. It would be nice if returning errors with extra user defined fields is possible.

neelance:
This got resolved by #49, especially see https://github.com/neelance/graphql-go/pull/49#issuecomment-271729496.

F21:
Hey @neelance 

I am a bit confused on how to use the `QueryError`. In my case, I want to return errors with  message and code, for example: 

```json
{
   ...
  errors: [
    {"key": "some_error", "msg": "some message"},
    ...
  ]
}
```

Is there an example for using `QueryError`?

PR #49, opened 2017-01-10
POC: custom error handlers

nicksrandall:
This proof of concept will likely need changes before it can be merged but I hope it can serve as a starting point for a conversation around how we can elegantly handle errors in this library.

This change gives users an optional hook to handle errors. Access to this hook opens up the door for several use cases:

-  Send more information rich errors to the client (related to #37) like sending error codes or other data related to error.
- Single place to filter out "non-user-safe" errors from being sent to the client.
- Integrate/send errors created in my resolves or graphql to a logging/monitoring service.

One benefit of this approach is that it is non-breaking. Users are able to "opt-in" when needed.

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -554,7 +554,9 @@ func (e *fieldExec) execField(ctx context.Context, r *request, f *query.Field, r
 	result, err := e.execField2(spanCtx, r, f, resolver, span)
 
 	if err != nil {
-		r.addError(errors.Errorf("%s", err))
+		queryError := errors.Errorf("%s", err)
+		queryError.ResolverError = err
+		r.addError(queryError)
 		addResult(f.Alias, nil) // TODO handle non-nil
 
 		ext.Error.Set(span, true)
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -3,8 +3,9 @@ package errors
 import "fmt"
 
 type QueryError struct {
-	Message   string      `json:"message"`
-	Locations []*Location `json:"locations,omitempty"`
+	Message       string      `json:"message"`
+	Locations     []*Location `json:"locations,omitempty"`
+	ResolverError error       `json:"-"`
 }
 
 type Location struct {


nicksrandall:
@jargv What are you're thoughts on this?

neelance:
`Schema.Exec` returns the list of errors in the `Errors` field of its `Response` value. There is the chance to modify them before returning them to the client. Maybe the `QueryError` struct should have an additional `ResolverError error` field that contains the original error so you can post-process.

For adding custom fields to the errors I'd suggest creating a new error type that has `QueryError` embedded and a new `Response` type that allows those custom errors to be encoded as JSON. This should only require a few lines of code and seems like the most lightweight solution to me.

What do you think?

nicksrandall:
@neelance I've made that change. What do you think?

neelance:
Yes, that's the direction I wanted to go. But I think we should only set `ResolverError` to errors actually returned by a resolver. Otherwise it can be `nil`. No need to always wrap an error.

neelance:
So I wouldn't even change the signature of `Errorf` but simply set the `ResolverError` field in the single instance where we call the resolver.

nicksrandall:
Ok, I made that change. What do you think? Thanks again for your help!

internal/exec/exec.go
```diff
@@ -554,7 +557,9 @@ func (e *fieldExec) execField(ctx context.Context, r *request, f *query.Field, r
 	result, err := e.execField2(spanCtx, r, f, resolver, span)
 
 	if err != nil {
-		r.addError(errors.Errorf("%s", err))
+		queryError := errors.Errorf("%s", err)
+		queryError.ResolverError = err
+		r.addError(queryError)
 		addResult(f.Alias, nil) // TODO handle non-nil
```
nicksrandall: This may be a discussion for another thread but shouldn't this be `addResult(f.Alias, result)` if the handler was able to resolve a partial amount of data, shouldn't we still send that to client and allow them to figure out what to do with it?

neelance: The spec says that `null` should be returned on error: http://facebook.github.io/graphql/#sec-Errors-and-Non-Nullability

Actually the `null` should even propagate to the parent if the current field has a non-null type. This is not yet implemented.

nicksrandall: Interesting. Ok! Thanks!

internal/exec/exec.go
```diff
@@ -554,7 +557,9 @@ func (e *fieldExec) execField(ctx context.Context, r *request, f *query.Field, r
 	result, err := e.execField2(spanCtx, r, f, resolver, span)
 
 	if err != nil {
-		r.addError(errors.Errorf("%s", err))
+		queryError := errors.Errorf("%s", err)
+		queryError.ResolverError = err
```
neelance: Nice and simple. ❤️

internal/exec/exec.go
```diff
@@ -303,6 +303,9 @@ func (r *request) addError(err *errors.QueryError) {
 func (r *request) handlePanic() {
 	if err := recover(); err != nil {
 		execErr := errors.Errorf("graphql: panic occured: %v", err)
+		if err, ok := err.(error); ok {
+			execErr.ResolverError = err
```
neelance: Hmm, is there a good scenario on why we should set the field on a panic? A panic should always be an unexpected internal error, so there should be nothing to tell the user except "internal error".

Right now the error message includes the panic's error value, but eventually I want to change it to only say "internal error" and not leak internals. However right now I don't want to do that yet because there are still some graphql-go error conditions that cause a panic and they would be hard to understand if the message was omitted.

nicksrandall: My thought here is that this could help your use case of not leaking internals. The error message could say something like "internal error" but more information (not available to client) could be included on the ResolverError (maybe thats a bad name) which won't get sent to client because it wont be serialized to json.

Also, if something panics in a resolver, won't it eventually be caught by this func? I don't have a use case for this yet but I could see myself wanting to have access to the originating error? 

I don't feel very strongly about this so if you'd like me to remove it then I will.

neelance: If you want access to the originating error to extract information from it, then the error is not unexpected and thus it does not qualify for a panic, right? ;-)

nicksrandall: Thats true.

nicksrandall: Removed.

nicksrandall:
I removed changes from the panic Handler. Are we good to merge now?

neelance:
Yes, we are!

neelance:
@nicksrandall Thanks for contributing!

F21:
Is there any example for using this?

felixfbecker:
Is there an example of how to use this?

DenisNeustroev:
we should write some docs 📝

felixfbecker:
@DenisNeustroev in lieu of docs, do you have a minimal example? 🙇

jpascal:
@DenisNeustroev I explored all code and I can't find way to provide custom errors to client.

sanae10001:
@jpascal What about this?
https://gist.github.com/sanae10001/9ad4c2ebfe640751f7e3278f5c6eb02a

PR #62, opened 2017-02-22
Respect Context Cancellation and Deadline

### LINKED ISSUE #61: Context Deadline Should Be Respected
Exec should complete when a context deadline is exceeded or the context is cancelled. Resolve methods that haven't been started should not be started after a context has already expired.

RaniSputnik:
Points to note:
* `ErrContextDeadlineExceeded` is exposed so that the error message can be customised
* Do we need to distinguish between a context cancel and a context deadline exceeded?
* Increases test suite run time by 100ms - perhaps this test should be run in parallel?
* Currently doesn't stop nested resolve methods from being started, even after the context is cancelled, I will look to improve this either now or in another PR.

Feedback much appreciated.

Fixes #61

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -25,6 +25,10 @@ const OpenTracingTagTrivial = "graphql.trivial"
 const OpenTracingTagArgsPrefix = "graphql.args."
 const OpenTracingTagError = "graphql.error"
 
+// ErrContextDeadlineExceeded is the error that will be returned when a context
+// is cancelled before an Exec command is completed
+var ErrContextDeadlineExceeded = errors.Errorf("context deadline was exceeded")
+
 type ID string
 
 func (_ ID) ImplementsGraphQLType(name string) bool {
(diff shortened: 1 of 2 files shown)


neelance:
What are you trying to achieve? Imho this code does not give you anything. It does not prevent any work from happening and it does not give the caller of `Exec` any information that it does not yet have (meaning that the caller cloud listen for `ctx.Done()` if it is interested in an early return). The idea of cancelling a context is to avoid unnecessary work. This often results in returning early, but imho that's a consequence, not the purpose.

RaniSputnik:
When I started using the library, it made intuitive sense to me that I would cancel a context and Exec would stop blocking that moment. I agree with what you're saying though, that makes more sense. If a consumer of the library wants that functionality it's easy enough to manage themselves.

I understood the more valuable change would be preventing nested resolvers from starting after the context is cancelled, I missed the fact you'd already done it, 😊  doh!

PR #63, opened 2017-02-22
Added gitignore

RaniSputnik:


neelance:
Thanks for wanting to contribute. However I'm not a fan of such generic `.gitignore` files. Please use http://stackoverflow.com/questions/7335420/global-git-ignore if your system arbitrarily pollutes directories.

PR #66, opened 2017-03-14
remove default param on deprecated directive fixes #65

### LINKED ISSUE #65: Graphiql is broken.
If you view graphiql in the starwars example of of this repo, you'll see that you are greeted with a confusing error. This was introduced when support for deprecated fields was added. 

I'm investigating the root cause of the error.

nicksrandall:
I'm not sure why, but for some reason the Graphiql parser did not like the default value here.

### DIFF
--- a/introspection/introspection.go
+++ b/introspection/introspection.go
@@ -232,8 +232,10 @@ func (r *Field) DeprecationReason() *string {
 	if !ok {
 		return nil
 	}
-	reason := args["reason"].(string)
-	return &reason
+	if reason, ok := args["reason"].(string); ok {
+		return &reason
+	}
+	return nil
 }
 
 type InputValue struct {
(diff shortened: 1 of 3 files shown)


nicksrandall:
Not sure why the build failed? Let me know what I can do to help.

shurcooL:
> Not sure why the build failed?

What makes you unsure? Have you looked at the build output? It says there was a panic in `TestDeprecatedDirective`, during an interface conversion.

introspection/introspection.go
```diff
@@ -232,8 +232,10 @@ func (r *Field) DeprecationReason() *string {
 	if !ok {
 		return nil
 	}
-	reason := args["reason"].(string)
-	return &reason
+	if reason, ok := args["reason"].(string); ok {
+		return &reason
+	}
+	return nil
```
shurcooL: Style suggestion. Keep the normal path at lesser indentation, indent the `return nil` paths, like so:

```Go
args, ok := r.field.Directives["deprecated"]
if !ok {
	return nil
}
reason, ok := args["reason"].(string)
if !ok {
	return nil
}
return &reason
```

Reference: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow.

neelance:
Superseded by 36f8ba8ba769fe7d76bb6f5954cad2b4d721b3e9

PR #73, opened 2017-03-23
Add _test to testing file so it doesn't get included in builds

jargv:
the testing.go file was being included in regular (non-test) builds. This was causing problems because in other parts of the app, we detect if we're running in a test by doing:

```go
// check if we're running in a test
if flag.Lookup("test.v") != nil { ... }
```

It's also problematic, in general, that our app has all of the testing flags registered (and displayed with `-h`) by just importing graphql-go.

### DIFF
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -1,4 +1,4 @@
-package graphql
+package gqltesting
 
 import (
 	"bytes"
@@ -6,11 +6,13 @@ import (
 	"encoding/json"
 	"strconv"
 	"testing"
+
+	graphql "github.com/neelance/graphql-go"
 )
 
 // Test is a GraphQL test case to be used with RunTest(s).
 type Test struct {
-	Schema         *Schema
+	Schema         *graphql.Schema
 	Query          string
 	OperationName  string
 	Variables      map[string]interface{}
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/neelance/graphql-go"
 	"github.com/neelance/graphql-go/example/starwars"
+	"github.com/neelance/graphql-go/gqltesting"
 )
 
 type helloWorldResolver1 struct{}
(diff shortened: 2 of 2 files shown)


neelance:
Uhm, I don't think this should be `_test`, because it is not a test of the `graphql` package. It is a tool for other packages to use and I think it would be unavailable with the `_test` suffix.

What about putting it in a subpackage called `gqltesting`?

shurcooL:
> It's also problematic, in general, that our app has all of the testing flags registered (and displayed with -h) by just importing graphql-go.

Yeah, importing `testing` package in a library is not a good idea.

Moving it to a subpackage makes sense.

jargv:
> What about putting it in a subpackage called gqltesting?

I started down that path. It introduced a cyclic dependency issue between the graphql and gqltesting packages. I have a feeling this will cascade further.

What do you think of me changing it back, and introducing an interface of the used parts of the `*testing.T` type so that this package doesn't depend directly on "testing"?

jargv:
> What do you think of me changing it back, and introducing an interface of the used parts of the *testing.T type so that this package doesn't depend directly on "testing"?

I started down this path and encountered another issue: the `.Run` method of `testing.T` has the signature
```go
  func (t *T) Run(name, func(*T)) bool
```
so AFAICT, it's not possible to create an interface in order to avoid referring to the testing package.

shurcooL:
> I started down that path. It introduced a cyclic dependency issue between the graphql and gqltesting packages. I have a feeling this will cascade further.

I expect that `qltesting` should import `graphql`, but why does `graphql` need to import `qltesting`?

jargv:
> I expect that qltesting should import graphql, but why does graphql need to import qltesting?

Oh interesting. The test file `graphql_test.go` is *not* in the `graphql` package, I had assumed that it was in the same package.  I didn't know that the "one package per directory" rule was lifted in the case of tests. I've been moving my non-package tests to a different director for no reason apparently.

neelance:
Cool, merged. Thanks!

shurcooL:
> I didn't know that the "one package per directory" rule was lifted in the case of tests. I've been moving my non-package tests to a different director for no reason apparently.

@jargv It's not lifted. There are two type of test files, internal external. Internal ones are compiled together with the package, but external ones are compiled as a separate package.

This behavior is documented at https://golang.org/cmd/go/#hdr-Test_packages:

> Test files that declare a package with the suffix "_test" will be compiled as a separate package, and then linked and run with the main test binary.

jargv:
@shurcooL I see. Thanks for the clarification.

PR #76, opened 2017-03-31
implement user defined logger (with sensible defaults)

nicksrandall:
This allows users to implement their own logger for when a panic is received during query execution. My use case is that I have implemented a custom logger (i.e. sends logs to logging service) and I would like the  graphql panics to show there when they occur.

This PR introduces no breaking changes but allows users to "opt in" to implementing their own logger if they choose to. Otherwise, everything behaves as it did before.

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -17,6 +17,7 @@ import (
 	"github.com/neelance/graphql-go/internal/schema"
 	"github.com/neelance/graphql-go/internal/validation"
 	"github.com/neelance/graphql-go/introspection"
+	"github.com/neelance/graphql-go/log"
 	"github.com/neelance/graphql-go/trace"
 )
 
@@ -49,6 +50,7 @@ func ParseSchema(schemaString string, resolver interface{}, opts ...SchemaOpt) (
 		schema:         schema.New(),
 		maxParallelism: 10,
 		tracer:         trace.OpenTracingTracer{},
+		logger:         &log.DefaultLogger{},
 	}
 	for _, opt := range opts {
 		opt(s)
@@ -85,6 +87,7 @@ type Schema struct {
 
 	maxParallelism int
 	tracer         trace.Tracer
+	logger         log.Logger
 }
 
 // SchemaOpt is an option to pass to ParseSchema or MustParseSchema.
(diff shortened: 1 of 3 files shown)


graphql.go
```diff
@@ -104,6 +106,13 @@ func Tracer(tracer trace.Tracer) SchemaOpt {
 	}
 }
 
+// Logger is used to log panics durring query execution. It defaults to exec.DefaultLogger.
+func Logger(logger exec.Logger) SchemaOpt {
```
neelance: `exec` is an internal package. Don't expose it in the public API.

internal/exec/exec.go
```diff
@@ -18,36 +18,47 @@ import (
...
 }
 
 func makePanicError(value interface{}) *errors.QueryError {
-	err := errors.Errorf("graphql: panic occurred: %v", value)
-	const size = 64 << 10
-	buf := make([]byte, size)
-	buf = buf[:runtime.Stack(buf, false)]
-	log.Printf("%s\n%s", err, buf)
-	return err
+	return errors.Errorf("graphql: panic occurred: %v", value)
```
neelance: `makePanicError` is also called from `execFieldSelection` where the stack trace now gets lost.

internal/exec/exec.go
```diff
@@ -18,36 +18,47 @@ import (
...
 )
 
+type Logger interface {
+	Log(ctx context.Context, value interface{})
+}
+
+type DefaultLogger struct{}
+
+func (l *DefaultLogger) Log(_ context.Context, value interface{}) {
+	err := makePanicError(value)
```
neelance: The name `Log` does not sound like this would handle a panic.

nicksrandall:
@neelance thanks for reviewing this! I appreciate the feedback. I have made a new, public log package and renamed the interface method to `HandlePanic`. I have also ensured that `execFieldSelection` doesn't loose the stack trace that it originally had.

neelance:
This looks good to merge. At some point we may want to merge this with the `trace` package, I'm not sure yet.

nicksrandall:
Thanks! Sounds good.

internal/exec/exec.go
shurcooL: Was this addressed?

PR #77, opened 2017-04-04
Add supports for snake case

subosito:
Most API from Rails backend use snake_case for the fields and arguments naming. This PR automatically map snake_case name to proper name in Go.

### DIFF
--- a/internal/exec/resolvable/packer.go
+++ b/internal/exec/resolvable/packer.go
@@ -130,8 +130,11 @@ func (b *execBuilder) makeStructPacker(values common.InputValueList, typ reflect
 	var fields []*structPackerField
 	for _, v := range values {
 		fe := &structPackerField{field: v}
+		fx := func(n string) bool {
+			return strings.EqualFold(n, v.Name.Name) || strings.EqualFold(camelize(n), camelize(v.Name.Name))
+		}
 
-		sf, ok := structType.FieldByNameFunc(func(n string) bool { return strings.EqualFold(n, v.Name.Name) })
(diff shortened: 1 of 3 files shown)


internal/exec/resolvable/packer.go
```diff
@@ -130,8 +130,11 @@ func (b *execBuilder) makeStructPacker(values common.InputValueList, typ reflect
 	var fields []*structPackerField
 	for _, v := range values {
 		fe := &structPackerField{field: v}
+		fx := func(n string) bool {
+			return strings.EqualFold(n, v.Name.Name) || strings.EqualFold(camelize(n), camelize(v.Name.Name))
```
subosito: @neelance Yeah, you're right :). Updated in https://github.com/neelance/graphql-go/pull/77/commits/6e8c113057e01a51ddc13f17702b4d3506c195e7.

shurcooL: `strings.EqualFold(n, v.Name.Name)` shouldn't be needed either, should it?

As far as I can tell, for all inputs where `strings.EqualFold(camelize(n), camelize(v.Name.Name))` is false, ``strings.EqualFold(n, v.Name.Name)` would be false too.

subosito: @shurcooL  Yup, I think that's not necessary, I was thinking to preserve original comparison. Removed in https://github.com/neelance/graphql-go/pull/77/commits/416031fa414f09c20d97281d9da93b90ab1e3e52.

PR #81, opened 2017-04-25
Add support for Get prefix on resolver Methods (gogoproto Getters).

hsyed:
I'm quite desperate to use this codebase but I am having quite a hard time. I have a sprawling tree of complex mutations where each node follows a very similar structure -- I can't leverage this till custom resolvers are in. It took me 700 lines and a lot of tedium just to bootstrap a simple toy/stub version of  our model . In reality the model is much more complex with 2-4 DTO's in use in the Input and Output position per node. 

I'm using the SnakeCase patch as a hack to have the DTOs work as input types and type resolvers --i.e.., `Name_()` for the types. This only aids marginally.

The Project has the Kubernetes go2idl code generation infrastructure in place to generate proto DTOs -- gogoproto can be configured to generate Get methods. If I had this I could generate and orchestrate proto DTOs this would make things more manageable. 

p.s., I am a fan of the hand written schema approach -- But I also really desperately need custom resolvers for this application.

### DIFF
--- a/internal/exec/resolvable/resolvable.go
+++ b/internal/exec/resolvable/resolvable.go
@@ -333,7 +333,7 @@ func (b *execBuilder) makeFieldExec(typeName string, f *schema.Field, m reflect.
 
 func findMethod(t reflect.Type, name string) int {
 	for i := 0; i < t.NumMethod(); i++ {
-		if strings.EqualFold(stripUnderscore(name), stripUnderscore(t.Method(i).Name)) {
+		if strings.EqualFold(stripUnderscore(name), stripGetPrefixAndUnderscore(t.Method(i).Name)) {
 			return i
 		}
 	}
@@ -345,6 +345,13 @@ func unwrapNonNull(t common.Type) (common.Type, bool) {
 		return nn.OfType, true
 	}
 	return t, false
+}
+
+func stripGetPrefixAndUnderscore(s string) string {
+	if strings.HasPrefix(s, "Get") {
+		s = s[3:]
+	}
+	return stripUnderscore(s)
 }
 
 func stripUnderscore(s string) string {
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -42,6 +42,27 @@ func (r *helloSnakeResolver2) SayHello(ctx context.Context, args *struct{ FullNa
 	return "Hello " + args.FullName + "!", nil
 }
 
+type helloGetResolver1 struct{}
+
+func (r *helloGetResolver1) GetHelloHTML() string {
+	return "Hello snake!"
+}
+
+func (r *helloGetResolver1) GetSayHello(args *struct{ FullName string }) string {
+	return "Hello " + args.FullName + "!"
+}
+
+type helloGetResolver2 struct {
+}
+
+func (r *helloGetResolver2) GetHelloHTML(ctx context.Context) (string, error) {
+	return "Hello snake!", nil
+}
+
+func (r *helloGetResolver2) GetSayHello(ctx context.Context, args *struct{ FullName string }) (string, error) {
+	return "Hello " + args.FullName + "!", nil
+}
+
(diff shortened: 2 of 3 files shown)


tonyghita:
A few questions: 

What is a DTO? 

Why do you have to prefix resolver methods with `Get`? 

To my understanding, that's an anti-pattern in golang (as is snake case). See https://golang.org/doc/effective_go.html#Getters and https://golang.org/doc/effective_go.html#mixed-caps

What is a "custom resolver"? How does it differ from any other resolver?

You can use the `introspection` package to iterate over your schema and generate resolvers automatically if you have many to write. I've used this together with golang's `text/template` package to generate resolver boilerplate.

neelance:
I guess what you are trying to do is to use your internal data structures directly as resolvers. Your struct already has a field `Foo` and you can't create a method now with the name `Foo` to resolve it, so you want to use `GetFoo`. My suggestion is to use the resolvers as a mapping between your internal data structure and your external GraphQL API by wrapping your types:
```
type MyDataResolver struct {
   d *somepkg.MyData
}

func (r MyDataResolver) Foo() int {
   return r.d.Foo
}
```

hsyed:
@tonyghita A DTO is a [data transfer object](https://en.wikipedia.org/wiki/Data_transfer_object). I am generating proto from structs in this project already. Getters might be an anti-pattern but it is essentially what this library is using for resolvers that just map a data carrying object. it is a a "Getter" without a "Get" prefix. [gogoproto can generate getters](https://godoc.org/github.com/gogo/protobuf/plugin/face) for protobuf objects. Gogoproto is used as a replacement for the standard go generator in most go projects using protobuf.

[Custom Resolvers](https://github.com/neelance/graphql-go/issues/58#issuecomment-289322294). The custom resolvers are another issue really I should not have mentioned it here. The project I am working on will need this support and it does have a side effect of eliminating a lot of boilerplate generation.

hsyed:
@neelance I am doing what you suggest at the moment, but I am also using protobuf in this project. I just have to change a setting and it would generate the accessors for the DTOs automatically. I wrote a toy representation of an aspect of our model by hand and it turned out to be 700 lines this model has  only 1-2 field per DTO -- if I started fleshing it out I would be spending most of my coding hours writing the getters and making sure they map correctly to the (multiple) GraphQL schemas. Instead I should be focusing my energy on making sure the schemas are correct and writing resolver logic that does more than just serve as an Accesor.

Code generation is the key to using GraphQL for large projects that have multiple RPC layers (protobuf / rest / graphql). Being able to use the gogoproto generated Getters is a small step towards this and it helps today. 

The kubernetes [go2idl](https://github.com/kubernetes/kubernetes/tree/master/cmd/libs/go2idl) tools are an effective model for maintaining an API layer. With custom resolvers similar strategies would be possible for GraphQL .

internal/exec/resolvable/resolvable.go
```diff
@@ -347,6 +347,13 @@ func unwrapNonNull(t common.Type) (common.Type, bool) {
 	return t, false
 }
 
+func stripGetPrefixAndUnderscore(s string) string {
+	if strings.HasPrefix(s, "Get") {
```
hsyed: should check if next field is "_" or `IsUpper` otherwise it would greedily match words starting with Get.

neelance:
Can't you just code generate the resolvers I suggested?

hsyed:
I could do that but my project already has code generation for DTOs -- the code generation tech already has Accessor interface and accessor method generation support  and it is a widely adopted in the go community (gogoproto). Adding a code generation pass for the accessors would be redundant work and others can't make use of it.

neelance:
Sorry, but I don't agree that accessor methods with `Get` are "widely adopted in the go community", see links in @tonyghita's comment above. Unless I see more popular demand here, I'd say that the better solution is to extend your codegen.

PR #83, opened 2017-05-09
Fix panic when resolver is not a pointer

tonyghita:
I ran into a panic when I had a non-nullable object on this line. Because the reflect type is struct and not `reflect.Ptr`, `resolver.IsNil()` panics.

I'm not sure why it matched the case, and I wasn't sure how to add a test for this case.

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -203,7 +203,7 @@ func (r *Request) execSelectionSet(ctx context.Context, sels []selected.Selectio
 	t, nonNull := unwrapNonNull(typ)
 	switch t := t.(type) {
 	case *schema.Object, *schema.Interface, *schema.Union:
-		if resolver.IsNil() {
+		if resolver.Kind() == reflect.Ptr && resolver.IsNil() {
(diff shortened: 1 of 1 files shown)


tonyghita:
I think I just misunderstood that resolvers must always be pointer types, even if the object they are resolving is non-nullable.

I've switched to return pointer-to-resolvers in my code and no longer hit the panic case.

shurcooL:
Is it worth it (and not prohibitively expensive) to still detect when the user does something that's not allowed, and print a helpful error message?

tonyghita:
I don't think this is an expensive operation, however I'll defer to @neelance to decide whether it should be valid for a resolver to not have a pointer type.

I'm not yet familiar enough with the codebase to say whether an error should be given to the user if the value is not a pointer-to-resolver, or if it's fine to use the value-of-resolver instead of the pointer.

neelance:
Resolvers don't always have to be pointers, only if they are nullable. I think this PR was good.

neelance:
Thx!

PR #89, opened 2017-05-25
fix panic when variable name not declared

nicksrandall:
I was getting a panic when I misnamed a variable (typo). This prevents that panic. 

Example query that would panic:
```graphql
query HeroNameAndFriends($e: Episode) {
     hero(episode: $episode) {
        name
    }
}
```

Now it just gives the error that the variable was never defined.

### DIFF
--- a/internal/validation/validation.go
+++ b/internal/validation/validation.go
@@ -508,7 +508,7 @@ func validateLiteral(c *opContext, l common.Literal) {
 func validateValueType(c *opContext, v common.Literal, t common.Type) (bool, string) {
 	if v, ok := v.(*common.Variable); ok {
 		for _, op := range c.ops {
-			if v2 := op.Vars.Get(v.Name); v != nil {
+			if v2 := op.Vars.Get(v.Name); v != nil && v2 != nil {
 				t2, err := common.ResolveType(v2.Type, c.schema.Resolve)
 				if _, ok := t2.(*common.NonNull); !ok && v2.Default != nil {
 					t2 = &common.NonNull{OfType: t2}


neelance:
Thanks!

internal/validation/validation.go
```diff
@@ -508,7 +508,7 @@ func validateLiteral(c *opContext, l common.Literal) {
 func validateValueType(c *opContext, v common.Literal, t common.Type) (bool, string) {
 	if v, ok := v.(*common.Variable); ok {
 		for _, op := range c.ops {
-			if v2 := op.Vars.Get(v.Name); v != nil {
+			if v2 := op.Vars.Get(v.Name); v != nil && v2 != nil {
```
shurcooL: Why didn't you just replace `v != nil`? It's not needed. I'm guessing it was originally a typo and @neelance meant to write `v2` instead. `v` is guaranteed not to be `nil` by the time the `v != nil` check executes, because if it were, `v.Name` would have already panicked. So, this code is now equivalent to:

```Go
if v2 := op.Vars.Get(v.Name); true && v2 != nil {
```

If you do want to check that `v != nil`, consider moving it to line 509:

```Go
if v, ok := v.(*common.Variable); ok && v != nil {
```

/cc @dominikh Do you think this could be detected by a static check?

dominikh: @shurcooL Once I finalize my VRP work, yes.

nicksrandall: Good point @shurcooL. I'll submit another PR in the morning to fix it.

PR #90, opened 2017-05-26
Allow custom scalars as input arguments

RXminuS:
Fixes https://github.com/neelance/graphql-go/issues/79 

This is super basic & probably needs some more work to 'type-check' custom scalars, but this at least gives the possibility of using things like Time as input arguments. Hopefully this PR can kickstart a more permanent fix.


neelance:
Thanks! This is a good start on the issue.

PR #91, opened 2017-05-26
Small fix based on feedback.

nicksrandall:


shurcooL:
As a fixup to #89, this has my LGTM.

On a _separate_ note, I want to ask if there's a guarantee that `v` (not `v2`) is not null because of preconditions elsewhere in the code? Because if `v` is ever null, the code will panic.

Issue #92, opened 2017-05-30
panic on negative int as argument


shnellpavel:
When I try to pass negative integer as argument for query field, I've got an error:
```
{
  "errors": [
    {
      "message": "syntax error: invalid value",
    }
  ]
}
```
During debug I've found that source of panic was [literals.go#L191](https://github.com/neelance/graphql-go/blob/master/internal/common/literals.go#L191).
Passing positive int processed properly.

Part of schema:
```
type User {
  ...
   photos_filtered(size: Int, page: Int, locale: String!): UserPhotos
  ...
}
```
Query: 
```
query ql {
  user {
    ...
    photos_filtered(locale: "ru", size: -1) {
    	...
    }
  }
}
```

PR #94, opened 2017-05-30
addition to #92: add test cases to check support of negatives

shnellpavel:
Add same changes to generation source:
https://github.com/neelance/graphql-js/pull/1

### DIFF
--- a/internal/tests/testdata/tests.json
+++ b/internal/tests/testdata/tests.json
@@ -6,6 +6,12 @@
     "errors": []
   },
   {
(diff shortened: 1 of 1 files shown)


neelance:
Thanks for wanting to contribute, but this needs to be brought upstream to https://github.com/graphql/graphql-js. Ping me after that is done, then I will regenerate the tests.

shurcooL:
For posterity, the upstream PR that was created is https://github.com/graphql/graphql-js/pull/891 (not 1, as the PR description incorrectly linked).

PR #98, opened 2017-06-25
Set content-type header to `application/json`

pavelnikolov:
This PR adds `Content-Type: application/json` header to the response of `relay.Handler`'s `ServerHTTP` method. 

I would like to submit more PRs to add some unit tests and improve the way `gqltesting` reports errors. For both of these I would like to add some dependencies like `github.com/stretchr/testify`. Do you have any dependency management tool preferences? I was thinking about `dep` or `glide`, WDYT?

### DIFF
--- a/relay/relay_test.go
+++ b/relay/relay_test.go
@@ -1,0 +1,36 @@
+package relay_test
+
+import (
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/neelance/graphql-go"
+	"github.com/neelance/graphql-go/example/starwars"
+	"github.com/neelance/graphql-go/relay"
+)
+
+var starwarsSchema = graphql.MustParseSchema(starwars.Schema, &starwars.Resolver{})
+
+func TestServeHTTP(t *testing.T) {
+	w := httptest.NewRecorder()
+	r := httptest.NewRequest("POST", "/some/path/here", strings.NewReader(`{"query":"{ hero { name } }", "operationName":"", "variables": null}`))
+	h := relay.Handler{Schema: starwarsSchema}
+
+	h.ServeHTTP(w, r)
+
+	if w.Code != 200 {
+		t.Fatalf("Expected status code 200, got %d.", w.Code)
+	}
+
+	contentType := w.Header().Get("Content-Type")
+	if contentType != "application/json" {
+		t.Fatalf("Invalid content-type. Expected [application/json], but instead got [%s]", contentType)
+	}
+
+	expectedResponse := `{"data":{"hero":{"name":"R2-D2"}}}`
+	actualResponse := string(w.Body.Bytes())
+	if expectedResponse != actualResponse {
+		t.Fatalf("Invalid response. Expected [%s], but instead got [%s]", expectedResponse, actualResponse)
+	}
+}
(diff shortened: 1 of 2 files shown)


neelance:
I appreciate that you want to contribute. However, `github.com/stretchr/testify` looks like a repo full of testing stuff that is unnecessary and especially in the Go world frowned upon. Please take a look at this related talk from the dotGo conference: https://www.youtube.com/watch?v=yi5A3cK1LNA

pavelnikolov:
Yeah, it makes sense. I get it why you don't want to introduce big dependences. I wanted to just use one method [`assert.JSONEq`](https://godoc.org/github.com/stretchr/testify/assert#JSONEq) - it gives you a very detailed message showing which fields are different and can save development time. Maybe I should just copy this method to the `gqltesting` package - do you think it is a good idea?
I also wanted to use the testify `mock` package, but I can add unit tests without it - I'll use your example schema and will not mock methods.
You do not have `CONTRIBUTING.md` document and this is why I ask what are your preferences before I spend time writing code, that might be declined later. 

I am using this project and would like to contribute to it as well.

What do you think about the `content-type: application/json` header? We have a 3rd party tool which consumes JSON data it requires the content type header to be present, though. This is why I added it.

relay/relay_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+	h.ServeHTTP(w, r)
+
+	if w.Code != 200 {
+		t.Errorf("Expected status code 200, got %d.", w.Code)
+	}
+
+	contentType := w.Header().Get("Content-Type")
+	if contentType != "application/json" {
+		t.Errorf("Invalid content-type. Expected [application/json], but isntead got [%s]", contentType)
```
neelance: Typo `isntead`.

relay/relay_test.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	expectedResponse := `{"data":{"hero":{"name":"R2-D2"}}}`
+	actualResponse := string(w.Body.Bytes())
+	if expectedResponse != actualResponse {
+		t.Errorf("Invalid response. Expected [%s], but isntead got [%s]", expectedResponse, actualResponse)
```
neelance: Same here.

relay/relay_test.go
pavelnikolov: 🤦‍♂️

neelance:
Cool. Thanks for contributing!

relay/relay_test.go
```diff
@@ -0,0 +1,36 @@
...
+		t.Fatalf("Expected status code 200, got %d.", w.Code)
+	}
+
+	contentType := w.Header().Get("Content-Type")
+	if contentType != "application/json" {
+		t.Fatalf("Invalid content-type. Expected [application/json], but instead got [%s]", contentType)
+	}
+
+	expectedResponse := `{"data":{"hero":{"name":"R2-D2"}}}`
+	actualResponse := string(w.Body.Bytes())
```
shurcooL: @pavelnikolov, `string(w.Body.Bytes())` could be simplified to `w.Body.String()`.

pavelnikolov: I'll fix it in the next PR

PR #99, opened 2017-06-29
Add `(*Schema).Validate`

Fugiman:
This adds a `Validate` method to the schema, which allows you to find out if a query is valid without actually running it. This is valuable when you have a client with static queries and want to statically determine whether they are valid.

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -122,7 +122,7 @@ type Response struct {
 	Extensions map[string]interface{} `json:"extensions,omitempty"`
 }
 
-// Validate validates the given query with the schema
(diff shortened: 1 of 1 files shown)


graphql.go
```diff
@@ -122,6 +122,16 @@ type Response struct {
 	Extensions map[string]interface{} `json:"extensions,omitempty"`
 }
 
+// Validate validates the given query with the schema
```
neelance: Missing dot at the end.

neelance:
Merged. Thanks for contributing!

PR #101, opened 2017-07-05
Update GraphiQL UI

pavelnikolov:
This PR updates the GraphiQL UI in the example server. The most significant change is that it adds support for query history.


neelance:
Cool! Thanks for contributing.

PR #107, opened 2017-09-08
cleaned up example to use MustParseSchema

kmarquardsen:


neelance:
Thanks for contributing!

PR #110, opened 2017-09-20
add built-in GraphiQL

klud1:
I've been using [graphql-go/graphql](github.com/graphql-go/graphql) along with [graphql-go/handler](github.com/graphql-go/handler) for a while, until I found out this library, which is way less verbose (thanks for that).

Thing is They recently added built-in GraphiQL support, based on those changes I adapted them to this library, I've tested them on a project as well as in the star-wars example and they're working flawlessly.

This avoids declaring two endpoints, one for GraphiQL and the other one for `/query` and now it works (for both GraphiQL and queries) on a single user-declared endpoint.

~~Plus It now shows pretty JSON if set so.~~

### DIFF
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -1,11 +1,14 @@
 package relay
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 
 	graphql "github.com/neelance/graphql-go"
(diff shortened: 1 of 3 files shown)


relay/relay.go
```diff
@@ -43,28 +46,189 @@ func UnmarshalSpec(id graphql.ID, v interface{}) error {
...
+		acceptHeader := r.Header.Get("Accept")
+		_, raw := r.URL.Query()["raw"]
+		if !raw && !strings.Contains(acceptHeader, ContentTypeJSON) && strings.Contains(acceptHeader, "text/html") {
+			renderGraphiQL(w, params)
+			return
+		}
+	}
+
+	// Use proper JSON Header
+	w.Header().Add("Content-Type", "application/json; charset=utf-8")
```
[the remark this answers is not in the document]
shurcooL: As far as I know, the default encoding for JSON is UTF-8, so it shouldn't be necessary to specify the charset in the header.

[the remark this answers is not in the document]
klud1: @shurcooL Roger that. Then I'll get rid of it. Thanks!

relay/relay.go
```diff
@@ -43,28 +46,189 @@ func UnmarshalSpec(id graphql.ID, v interface{}) error {
...
 
-	w.Header().Set("Content-Type", "application/json")
-	w.Write(responseJSON)
+	if h.pretty {
+		w.WriteHeader(http.StatusOK)
+		responseJSON, err := json.MarshalIndent(response, "", "\t")
```
pavelnikolov: Why would this handler pretty print? GraphiQL + many browser plugins + `jq` already do the pretty printing for you. I guess this handler is to be consumed by other apps only, no?

relay/relay.go
```diff
@@ -43,28 +46,189 @@ func UnmarshalSpec(id graphql.ID, v interface{}) error {
...
+
+// A workaround for getting `variables` as a JSON string
+type requestOptionsCompatibility struct {
+	Query         string `json:"query" url:"query" schema:"query"`
+	Variables     string `json:"variables" url:"variables" schema:"variables"`
+	OperationName string `json:"operationName" url:"operationName" schema:"operationName"`
+}
+
+func getFromForm(values url.Values) *RequestOptions {
+	query := values.Get("query")
```
pavelnikolov: +1 for accepting query string params and allowing `GET` requests. It deserves a separate PR, though.

klud1: @pavelnikolov this is not accepting a `GET` request, but getting the first value associated with the given key though. According to [golang spec](https://golang.org/pkg/net/url/#Values.Get)
> `func (v Values) Get(key string) string`
> Get gets the first value associated with the given key. If there are no values associated with the key, Get returns the empty string. To access multiple values, use the map directly.

relay/relay.go
klud1: Roger that! Getting rid of it!

LennyPenny:
Any updates on this @neelance

vergenzt:
I agree with @neelance on this one. I'd rather keep this library simple. I'd be in favor of adding _documentation_ for _how_ to serve GraphiQL alongside a GraphQL endpoint, but I'd rather it not be in the library.

Can we close this since @neelance has declined it?

klud1:
I'll close it, don't worry.

pavelnikolov:
You can check out the example included in this repo.

PR #117, opened 2017-10-11
Coerce integer ID scalar input

tonyghita:
This change enables ID scalar inputs to be coerced from integer values as mentioned in the [October 2016 version of the GraphQL specification](http://facebook.github.io/graphql/October2016/#sec-ID).

> Input Coercion
> 
> When expected as an input type, any string (such as "4") or integer (such as 4) input value should be coerced to ID as appropriate for the ID formats a given GraphQL server expects. Any other input value, including float input values (such as 4.0), must raise a query error indicating an incorrect type.

This PR has is made up of the following changes:

- Add test cases for string and integer ID input
- Add a case to `ID.UnmarshalGraphQL` switch statement to handle coercion/unmarshalling integer input
- Move ID scalar implementation from `graphql.go` to its own file for neater code organization
- Separate `scanner.Int` and `scanner.Float` argument literal parsing cases

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -6,8 +6,6 @@ import (
 
 	"encoding/json"
 
-	"strconv"
-
 	"github.com/neelance/graphql-go/errors"
 	"github.com/neelance/graphql-go/internal/common"
 	"github.com/neelance/graphql-go/internal/exec"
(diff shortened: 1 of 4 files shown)


nicksrandall:
LGTM

neelance:
LGTM, thanks!

PR #127, opened 2017-11-19
Fail to parse empty field lists for Object, Interface, and Input Object types

vergenzt:
Per the spec, [Object][1], [Interface][2], and [Input Object][3] types must define at least one field. This commit makes graphql-go fail to parse any of these types if no fields are defined.

[1]: http://facebook.github.io/graphql/October2016/#sec-Object-type-validation
[2]: http://facebook.github.io/graphql/October2016/#sec-Interface-type-validation
[3]: http://facebook.github.io/graphql/October2016/#sec-Input-Object-type-validation

---

The first three commits are some refactors that I thought may help with clarity. Only the last makes any observable behavior change.

Let me know what you think!

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -581,6 +581,10 @@ func TestIncludeDirective(t *testing.T) {
 
 type testDeprecatedDirectiveResolver struct{}
 
+func (r *testDeprecatedDirectiveResolver) Test() string {
+	return ""
+}
+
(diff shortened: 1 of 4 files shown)


pavelnikolov:
Thanks! I think this PR has to, actually, be split into 2 PRs. One for refactoring and another one to add validation for these types. This would make it much cleaner and would increase the chances of getting it merged. There are also conflicts to be resolved.

Issue #128, opened 2017-11-22
How could i resolve this?


francescjaume:
Hi!!

I have a question. I don't know how can I implements this schema:

```
schema{
    query: Query
    mutation: Mutation
}

type Query {
  admin: AdminQuery
}

type Mutation {
  admin: AdminMutation
}
```

Here the resolvers functions:

```
func (r* Resolver) Admin() *AdminMutationResolver {
	return &AdminMutationResolver{}
}
func (r* Resolver) Admin() *AdminQueryResolver {
	return &AdminQueryResolver{}
}

```

Thanks for all!!

lpalmes:
Hey @francescjaume, do you have more information about this example? Like what is you `AdminQuery` and `AdminMutation` type in your schema, or what you are trying to achieve, so we can be able to help

francescjaume:
Hi @lpalmes, we just need resolve admin in type Query and in type Mutation too.
But Go raises us that we have same function (Admin) duplicated. 

Our first solution is enable two diferent root resolvers like this:

```
func (r* MutationResolver) Admin() *AdminMutationResolver {
	return &AdminMutationResolver{}
}
func (r* QueryResolver) Admin() *AdminQueryResolver {
	return &AdminQueryResolver{}
}
```
We fork this aproach and it works good. Maybe you have some easier solution.

lpalmes:
Yep, the problem is that root resolvers, like the ones inside Mutation, Subscription or Query object are bound to have name collisions, given that they all resolved by `func (r *Resolver)`, i think what you are doing is good and a sensible solution, in my schema at least, i try to use verbs for mutation (eg: `updateUser`), and subjects for queries (`users: [User]!`), and i will take into account this solution if i ever need it, thanks!

PR #129, opened 2017-11-22
Accept two root resolvers

francescjaume:
#128

### DIFF
--- a/example/starwars/server/server.go
+++ b/example/starwars/server/server.go
@@ -12,7 +12,7 @@ import (
 var schema *graphql.Schema
 
 func init() {
-	schema = graphql.MustParseSchema(starwars.Schema, &starwars.Resolver{})
+	schema = graphql.MustParseSchema(starwars.Schema, &starwars.QueryResolver{}, &starwars.MutationResolver{})
 }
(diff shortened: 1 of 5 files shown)


vergenzt:
I like this in theory (and I was also confused originally why mutation and query resolvers got merged into the same place even though they're different types in the schema), but personally I'd rather not break backcompat at the moment, especially since there's no proper versioning scheme for this repo yet.

francescjaume:
Hi @vergenzt , the reason why i'm trying to do this is because i want to structure my schema in products, and then each product will have their own queries and mutations. For example:

```
schema{
    query:Query
    mutation:Mutation
}

type Query {
    apple: AppleQuery
    samsung: SamsungQuery
}

type Mutation {
     apple: AppleMutation
     samsung: SamsungMutation
}

type AppleQuery {
    mcbook: Mcbook
}

type AppleMutation {
    createMcbook: Mcbook
    updateMcbook: Mcbook    
    deleteMcbook: Mcbook
}
```
The only way to do this now is to have 2 different schemas. To merge all in the same IDL schema i need to apply my patch.

Issue #88, opened 2017-05-23
RFC: New way to declare resolvers


neelance:
I've just been to the GraphQL Europe conference and it got me thinking about how we specify resolvers. Using Go methods seemed nice at first, but is that really the best way? Please take a look at this alternative:
https://github.com/neelance/graphql-go/blob/new-resolvers/example/starwars/starwars.go#L291

Advantages:
- no resolver wrappers, especially good for list results
- API easier to learn
- clean shortcuts like `ResolverField`
- most type checking at startup is still possible
- explicit name mapping between GraphQL and Go
- automatic handling of GraphQL type assertions

Disadvanatages:
- no type checking on return value if type is a GraphQL interface (they map to Go's `interface{}`)

What do you think?

/cc @F21 @nicksrandall @acornejo @bsr203 @EmperorEarth

bsr203:
looks much simpler and easier for new users and looks great. I use relay "Node" interface a lot and not sure how much headache  would it be not having validation. But, I am ok with the tradeoff considering the overall simplification. Thanks for your continuing effort on improving the library. Cheers.

tonyghita:
This looks like a good way to lower the entry to barrier.

Would the existing way to define resolvers still be supported?

It could be that I'm used to the current method of writing resolvers, but I have a few concerns about doing away with the existing method and going only with this approach:

- would force the creation of API-specific models if the underlying data does not match the GraphQL schema.
- would make it harder to unit test, unless the functions are named.
- function names would have to be more verbose (maybe include the name of the object being resolved?) because we lose the ability to attach functions to types
- it becomes a bit unclear how to organize graphql-related code inside your project (what if I want the registration to be separate from the function definitions?)
- we lose some type checking
- ResolverField is convenient for simple use cases but in advanced use cases (i.e. lazy-loading of fields) maybe not so much

However, the only drawback I can think of if this method of specifying resolvers lives alongside the existing is that multiple ways to do the same things can be confusing.

neelance:
> Would the existing way to define resolvers still be supported?

No, but you can emulate it via https://golang.org/ref/spec#Method_expressions.

> * would force the creation of API-specific models if the underlying data does not match the GraphQL schema.

What do you mean by that?

> * would make it harder to unit test, unless the functions are named.
> * function names would have to be more verbose (maybe include the name of the object being resolved?) because we lose the ability to attach functions to types

I think you can still have them attached to a type by using https://golang.org/ref/spec#Method_expressions.

[...]

PR #132, opened 2017-11-28
Make subscriptions work

caseymrm:
There's some code towards supporting graphql subscriptions in queries, this fixes a few spots where they weren't handled yet.

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -1753,3 +1753,34 @@ func TestComposedFragments(t *testing.T) {
 		},
 	})
 }
+
+func TestSubscription(t *testing.T) {
+	gqltesting.RunTests(t, []*gqltesting.Test{
+		{
+			Schema: graphql.MustParseSchema(`
+				schema {
+					query: Query
+					subscription: Subscription
+				}
+
+				type Query {
+					hello: String!
+				}
+
+				type Subscription {
+					hello: String!
+				}
+			`, &helloWorldResolver1{}),
+			Query: `
+				{
+					hello
+				}
+			`,
+			ExpectedResult: `
+				{
+					"hello": "Hello world!"
+				}
+			`,
+		},
+	})
+}
--- a/internal/exec/resolvable/resolvable.go
+++ b/internal/exec/resolvable/resolvable.go
@@ -13,9 +13,10 @@ import (
 
 type Schema struct {
 	schema.Schema
-	Query    Resolvable
-	Mutation Resolvable
-	Resolver reflect.Value
+	Query        Resolvable
+	Mutation     Resolvable
+	Subscription Resolvable
+	Resolver     reflect.Value
 }
 
 type Resolvable interface {
@@ -57,7 +58,7 @@ func (*Scalar) isResolvable() {}
 func ApplyResolver(s *schema.Schema, resolver interface{}) (*Schema, error) {
 	b := newBuilder(s)
 
-	var query, mutation Resolvable
+	var query, mutation, subscription Resolvable
 
 	if t, ok := s.EntryPoints["query"]; ok {
 		if err := b.assignExec(&query, t, reflect.TypeOf(resolver)); err != nil {
@@ -71,15 +72,22 @@ func ApplyResolver(s *schema.Schema, resolver interface{}) (*Schema, error) {
 		}
 	}
 
+	if t, ok := s.EntryPoints["subscription"]; ok {
+		if err := b.assignExec(&subscription, t, reflect.TypeOf(resolver)); err != nil {
+			return nil, err
+		}
+	}
+
 	if err := b.finish(); err != nil {
 		return nil, err
 	}
 
 	return &Schema{
-		Schema:   *s,
-		Resolver: reflect.ValueOf(resolver),
-		Query:    query,
-		Mutation: mutation,
+		Schema:       *s,
+		Resolver:     reflect.ValueOf(resolver),
+		Query:        query,
+		Mutation:     mutation,
+		Subscription: subscription,
 	}, nil
 }
 
--- a/internal/exec/selected/selected.go
+++ b/internal/exec/selected/selected.go
@@ -35,6 +35,8 @@ func ApplyOperation(r *Request, s *resolvable.Schema, op *query.Operation) []Sel
 		obj = s.Query.(*resolvable.Object)
 	case query.Mutation:
 		obj = s.Mutation.(*resolvable.Object)
+	case query.Subscription:
+		obj = s.Subscription.(*resolvable.Object)
 	}
 	return applySelectionSet(r, obj, op.Selections)
 }


klud1:
Any updates on merging this?

AndrewSamokhvalov:
@caseymrm Hey! Glad to see that someone made a movement towards subscription implementation! 🍾 Is it possible to use them after merging the pull request and do you have an example of it?

Kivutar:
This should be merged. It is a step forward in supporting subscriptions.
And the next step would be to have a WebSocket implementation like this one https://github.com/functionalfoundry/graphqlws

robermorales:
I heard at #88 that @neelance held their dedication on the project. First of all, thanks all, and specially thanks to @neelance for the work on this pretty cool library.

Any idea on how the development can be directed now? Are the new PRs going to be revised and eventually accepted in this repo? Should somebody create a fork to lead the project from another repo?

neelance:
I am not opposed to a fork. Like I mentioned on #88, when at some point I have time available to spend on this project, then it is likely to undergo some major changes. That's maybe also not what everyone wants. If there is a good fork at some point, then I may link to it from the readme.

On this PR: Is this change already helpful on its own? Without WebSockets, subscriptions won't be "live", the only thing you could do is normal one-time queries on endpoints called "subscription". I am not yet sure that it is a good idea to add a half-baked state of the subscriptions feature.

Kivutar:
If the websocket part is handled by another module like done with graphql-go+graphqlws, this present change *should* be enough for this repo.
A second change may be needed later in graphiql to include a websocket client.

caseymrm:
Yeah, to be clear, this just promotes subscriptions to work the same as query or mutation. So it's purely semantics, until this is merged I currently do my subscriptions as queries and the rest of the code is unchanged. You could say the same thing about mutation though- technically you could just use a query for everything, but I like the separation.

Overall, your library already has some of the code to parse subscription, so it seems worth finishing it enough to make them usable. The library does not handle transport mechanisms in general, so I think this can go in without websocket support. In go it's so trivial to do a long poll (my code just does a blocking read on a channel inside the resolver) that I'm just doing that for now.

But up to you of course, feel free to close, or if there's additional work that you think would make this more baked I'm happy to take a look.

paulxuca:
There's been some work at a fork https://github.com/mikeifomin/graphql-go/tree/subscription

mikeifomin:
@paulxuca  yeah. it's still in progress
My idea is to use channels in resolvers 
https://github.com/mikeifomin/graphql-subscriptions-go-example/blob/master/schema/resolver.go#L42

Does anyone have any better ideas?

tonyghita:
Have you watched Lee Byron's introduction to GraphQL subscriptions on the server? https://www.youtube.com/watch?v=bn8qsi8jVew

I don't think a purely channel-based solution would be flexible enough for the purposes of this library.

mikeifomin:
@tonyghita yes I saw it.
I've implement apollo ws server protocol. (https://github.com/mikeifomin/graphql-go/commit/b407faba716b55d3c811a54f7bb5e21969da6ba1) 
With manual browser testing on simple subscriptions it works.

The high level idea is to add method 
```go
func (s *Schema) Subscribe (...) <-chan *Response
```
like 
```go 
func (s *Schema) Exec (...) *Response
```
Im really serious about subscription with go. I'm already using graphQL subscription on my project and have mocked server with node.js.
@tonyghita can you get me a little more details about your idea of flexibility?

tonyghita:
> can you give me a little more details about your idea of flexibility?

It sounds like you've put more practice into subscriptions + go than I have, but the biggest questions for me are:

- _Can the subscription implementation be agnostic of transport?_
   In my ideal library, I could plug in a WebSocket transport or WebSub transport, both, or something else (Server-Sent Events?). Just like this library doesn't dictate that you have to use HTTP, I don't think it should dictate the subscription transport.

- _Can the subscription implementation be agnostic of event system (input?_
  We should probably be able to specify go channels, or kafka, or whatever mechanism to provide the functionality. I'm not sure it's best that the mechanism should be exposed via the library's interface.

- _Can it scale reasonably?_
  Unlike queries and mutations, subscriptions are __stateful__ and as such more difficult to scale horizontally and make robust to failure. Does it make sense to introduce a stateful component to this library? Should that be delegated to a separate component which can be scaled independently of the stateless component? (I understood this as a takeaway from Lee's talk)

Kivutar:
In the apollo-server stack, the SubscriptionClient interface can be provided by different modules, one of them being websocket https://github.com/apollographql/subscriptions-transport-ws

And the PubSub mechanism is also abstracted, so you can use the default one, or redis, or something else https://github.com/davidyaha/graphql-redis-subscriptions

Their pubsub interface returns an AsyncIterator that you plug in your resolver.

jmichalicek:
I really appreciate the work you guys are putting into this.  I fell in love with golang just a few months ago and have been surprised at the lack of GraphQL support.  I've got a project using this package I'm working on just to see how it turns out, but the general lack of GraphQL support in the Go community keeps me questioning the wisdom of that when I had a working version in Elixir and could churn one out in Python in a few days.

I'd love to see this project turn into Go's Apollo/Absinthe/Graphene.

mytototo:
Hi there,

Is there any ETA regarding this PR?
Is it possible to update the example with this feature?

Thanks a lot for your contribution.

matiasanaya:
I've put together a graphQL subscriptions [proof of concept](https://github.com/matiasanaya/go-graphql-subscription-example) that:

* Relies on a [fork](https://github.com/matiasanaya/graphql-go/tree/feature/support-subscriptions) of `github.com/graph-gophers/graphql-go` (which incorporates many of @mikeifomin's ideas and code) as the subscription implementation.
* Relies on a [fork](https://github.com/matiasanaya/graphqlws/tree/feature/support-graph-gophers-graphql-go) of `github.com/functionalfoundry/graphqlws` as the transport implementation.
* _Separates subscription implementation from transport implementation_ as suggested by @tonyghita.
* Has a _subscription implementation that is agnostic of the event system_, also suggested by @tonyghita.

I'll be focusing my efforts on the `github.com/graph-gophers/graphql-go` [fork](https://github.com/matiasanaya/graphql-go/tree/feature/support-subscriptions) next. What I've got right now is no more than a buggy (working) prototype. I'm keen on input from anyone who's interested.

matiasanaya:
I've updated the [example subscriptions server](https://github.com/matiasanaya/go-graphql-subscription-example) (which uses this library) with a new [websockets transport library](https://github.com/matiasanaya/graphql-transport-ws). If anyone is keen on using subscriptions please have a look at those projects and maybe contribute :)

pavelnikolov:
There is a working version of subscriptions now.

lindroth:
So I saw! Great work everyone and a big thanks! <3

PR #142, opened 2017-12-13
Accept two root resolvers

francescjaume:
#128

### DIFF
--- a/example/starwars/server/server.go
+++ b/example/starwars/server/server.go
@@ -12,7 +12,7 @@ import (
 var schema *graphql.Schema
 
 func init() {
-	schema = graphql.MustParseSchema(starwars.Schema, &starwars.Resolver{})
+	schema = graphql.MustParseSchema(starwars.Schema, &starwars.QueryResolver{}, &starwars.MutationResolver{})
 }
 
 func main() {
--- a/example/starwars/starwars.go
+++ b/example/starwars/starwars.go
@@ -284,16 +284,17 @@ type review struct {
 
 var reviews = make(map[string][]*review)
 
-type Resolver struct{}
+type QueryResolver struct{}
+type MutationResolver struct{}
 
-func (r *Resolver) Hero(args struct{ Episode string }) *characterResolver {
+func (r *QueryResolver) Hero(args struct{ Episode string }) *characterResolver {
 	if args.Episode == "EMPIRE" {
 		return &characterResolver{&humanResolver{humanData["1000"]}}
 	}
 	return &characterResolver{&droidResolver{droidData["2001"]}}
 }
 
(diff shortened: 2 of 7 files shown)


nicksnyder:
I am not a fan of this proposed solution. Instead of accepting two resolvers, I would prefer to still accept a single resolver and that resolver is responsible for defining an implementation for a query resolver and a mutation resolver as proposed in https://github.com/neelance/graphql-go/issues/145

wenerme:
Why tow resolver, if you can do 
```golang
schema := graphql.MustParseSchema(schema, &Resolver{AResolver: A,BResolver:B})
```

francescjaume:
My problem is that query and mutation has same name. 
How should i declare root resolvers, if i want to do something like this?

`func (r *Resolver) Product() *QueryResolver {
	return &QueryResolver{}
}
func (r *Resolver) Product() *MutationResolver {
	return &MutationResolver{}
}`

tonyghita:
@francescjaume I think normally you would attach some kind of action to your mutation.
```go
func (r *Resolver) Product() *QueryResolver {
  return &QueryResolver{}
}

func (r *Resolver) UpdateProduct() *MutationResolver {
  return &MutationResolver{}
}
```

If you insist on naming then the same, then I think probably the right thing to do would be to follow @nicksnyder's proposal on #145.

arquio:
Hi!!!

First: Happy New Year!

The case that @francescjaume expose **is posible in Graphql spec** but now we need use this PR because we can't resolve our schema.

Example:

We have 10 Products ( Product1, Product2...., Product 10) and we need group the mutatios and queries by product. 

```
schema{
    query: Query
    mutation: Mutation
}

type Query {
  product1: Product1Query
  product2: Product2Query
  product3: Product3Query
  product4: Product4Query
  product5: Product5Query
  product6: Product6Query
}

type Mutation {
  product1: Product1Mutation
  product2: Product2Mutation
  product3: Product3Mutation
  product4: Product4Mutation
  product5: Product5Mutation
  product6: Product6Mutation
}
```

This case is imposible using the actual implementation and the @francescjaume solution do posible this.

@nicksnyder I think that #145 resolves this but I'm not sure, do you think that your proposal solves this schema?

Thanks!!!

nicksnyder:
@arquio Yes a solution to #145 would solve your issue.

arquio:
Great!!!

Thanks!!!

tonyghita:
Thanks for the contribution @francescjaume! I do agree with @nicksnyder that it would be better to accept a single resolver that defines the entry-points for each operation type, as documented in the specification. 

I'm going to close this pull-request in favor of the single-resolver with separate child resolvers for each operation type.

pavelnikolov:
Now I'm considering doing it like this https://github.com/graph-gophers/graphql-go/pull/561

PR #148, opened 2017-12-29
Support context injection in testing

OscarYuen:
Existing graphql testing may not be able to support to inject context.
Context injection is added for wider testing ability. Thanks


neelance:
Thanks for contributing.

PR #156, opened 2018-01-30
trace: Log graphql.variables rather than tag

keegancsmith:
According to the OT documentation tag values can be numeric types, strings, or
bools. The behavior of other tag value types is undefined at the OpenTracing
level. For example `github.com/lightstep/lightstep-tracer-go` generates error
events.


keegancsmith:
cc @neelance

PR #161, opened 2018-02-06
Support error checking in gqltesting

iheanyi:
The way that things exist now, it seems like the tests only allow for testing the happy path for GraphQL queries and not the sad path. I think in some cases, we want to be able to test the sad path as well.

### DIFF
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"reflect"
 	"strconv"
 	"testing"
 
 	graphql "github.com/graph-gophers/graphql-go"
+	"github.com/graph-gophers/graphql-go/errors"
 )
 
 // Test is a GraphQL test case to be used with RunTest(s).
@@ -18,7 +20,7 @@ type Test struct {
 	OperationName  string
 	Variables      map[string]interface{}
 	ExpectedResult string
-	ErrorExpected  bool
+	ExpectedErrors []*errors.QueryError
 }
 
 // RunTests runs the given GraphQL test cases as subtests.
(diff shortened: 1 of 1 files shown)


iheanyi:
Any word on this . . ?

gqltesting/testing.go
```diff
@@ -18,6 +18,7 @@ type Test struct {
 	OperationName  string
 	Variables      map[string]interface{}
 	ExpectedResult string
+	ErrorExpected  bool
```
tonyghita: What if we provided `Errors []*errors.QueryError` to allow for checking that the errors received were all expected?

This would let us do something like:
```go
func checkErrors(t *testing.T, expected, actual []*errors.QueryError) {
	expectedCount, actualCount := len(expected), len(actual)
	if expectedCount > 0 {
		t.Fatal("TODO: compare expected errors with actual errors")
	}

	for _, err := range actual {
		t.Errorf("unexpected error: '%s'", err)
	}

	if actualCount > 0 {
		t.Fatal("got unexpected errors")
	}
}
```

iheanyi: I dig that.

iheanyi: Also, thanks for the feedback, I really do appreciate it!

iheanyi:
@tonyghita I've implemented your suggested changes but the build is failing on Sempahore because it's complaining about using internal packages? Unsure of how to fix this, wondering if you had any ideas. Sorry for the delay in getting these changes out.

iheanyi:
Closed and reopened so Semaphore can run properly. Would appreciate some eyes and feedback on this. :)

gqltesting/testing.go
```diff
@@ -65,3 +67,28 @@ func formatJSON(t *testing.T, data []byte) []byte {
...
+
+	if expectedCount != actualCount {
+		t.Fatalf("unexpected number of errors: got %d, want %d", expectedCount, actualCount)
+	}
+
+	if expectedCount > 0 {
+		for i, want := range expected {
+			got := actual[i]
+
+			if !reflect.DeepEqual(got, want) {
```
tonyghita: What do you think about checking that the output of their `Error()` methods are the same instead of `reflect.DeepEqual`?

Is there something in particular we're looking for with DeepEqual?

iheanyi: Hmmm, I'm not quite sure. When it comes to error checking, I've always done `reflect.DeepEqual` rather than checking against strings. I guess for checking in `DeepEqual`, it'd be asserting things more granularly based on where they occurred for the errors rather than just the message of it. It was a simple error like `errors.New()` I could be inclined to compare `Error()` output. But looking at it, it could be a bit overkill to have to construct the `Location` struct for the query errors as well.

tonyghita: Ah `Location` may make sense for validation errors. I think this is probably okay for now; we can always simplify later.

iheanyi: Awesome. Is this ready to ship then?

PR #162, opened 2018-02-07
added pets example

deltaskelta:
I set out to make a small example to learn the lib and it ended up being pretty complete with a simple sqlite data layer added. Let me know what you think about adding it to the examples.

### DIFF
--- a/example/pets/.gitignore
+++ b/example/pets/.gitignore
@@ -1,0 +1,2 @@
+db.sqlite
+tmp/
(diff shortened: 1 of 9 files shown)


tonyghita:
Hey, nice example! I hadn't seen a `graphql-go` server that resolves against sqlite before.

I'm not sure this example belongs in this repository though. If you were to make your own `graphql-go` example repo, we could definitely link to it here. That way, you can evolve the example without needing to submit PRs here (since it'd live in your repo), we won't bring in extra dependencies unrelated to the core functionality, and we can start building a library of example implementations built by the community.

Here's one I wrote that resolves against a REST API: https://github.com/tonyghita/graphql-go-example

PR #166, opened 2018-02-12
Update validation tests

tonyghita:
This change updates 'testdata/tests.json' with the latest test cases from the github.com/graphql/graphql-js reference implementation.

This change also  moves the 'tests' package into the 'validation' package since these tests specifically relate to validation logic.

cc @neelance


internal/validation/testdata/export.js
```diff
@@ -47,19 +47,19 @@ const fakeModules = {
 	'mocha': {
 		describe(name, f) {
 			switch (name) {
-			case 'within schema language':
-				return;
+				case 'within schema language':
```
tonyghita: My IDE made these formatting changes... I'm okay reverting these if undesirable.

internal/validation/testdata/export.js
```diff
@@ -70,12 +70,11 @@ const fakeModules = {
...
 
 const originalLoader = Module._load;
-Module._load = function(request, parent, isMain) {
+Module._load = function (request, parent, isMain) {
 	return fakeModules[request] || originalLoader(request, parent, isMain);
 };
 
-require('./src/validation/__tests__/ArgumentsOfCorrectType-test');
-require('./src/validation/__tests__/DefaultValuesOfCorrectType-test');
+require('./src/validation/__tests__/ExecutableDefinitions-test');
```
tonyghita: This was added since the last time `tests.json` was generated.

internal/validation/testdata/export.js
```diff
@@ -91,13 +90,17 @@ require('./src/validation/__tests__/OverlappingFieldsCanBeMerged-test');
 require('./src/validation/__tests__/PossibleFragmentSpreads-test');
 require('./src/validation/__tests__/ProvidedNonNullArguments-test');
 require('./src/validation/__tests__/ScalarLeafs-test');
+// TODO: Support GraphQL subscriptions.
+// require('./src/validation/__tests__/SingleFieldSubscriptions-test.js');
```
tonyghita: Should we leave this commented-out until subscriptions are supported?

internal/validation/testdata/gen.go
```diff
@@ -0,0 +1,9 @@
+// Package testdata copies validation test cases from the Javascript reference implementation.
+//
+// Usage:
```
tonyghita: I had a lot of trouble figuring out how to use this package. I wasn't able to get all the way there because `babel-node` doesn't respect the `graphqljs/.babelrc` settings.

I'd like some help documenting the actual steps needed to update the validation test data.

internal/validation/validation_test.go
```diff
@@ -33,22 +33,22 @@ func TestAll(t *testing.T) {
...
-		t.Fatal(err)
+		t.Fatalf("test data JSON decode failed: %v", err)
 	}
 
 	schemas := make([]*schema.Schema, len(testData.Schemas))
 	for i, schemaStr := range testData.Schemas {
 		schemas[i] = schema.New()
 		if err := schemas[i].Parse(schemaStr); err != nil {
-			t.Fatal(err)
+			t.Fatalf("schema parse failed: %v", err)
```
tonyghita: With the additional test cases, schema parsing fails because the schema definition language (SDL) has added the `&` symbol in describing [interface implementations](http://facebook.github.io/graphql/draft/#sec-Interfaces).

```graphql
interface NamedEntity {
  name: String
}

interface ValuedEntity {
  value: Int
}

type Person implements NamedEntity {
  name: String
  age: Int
}

type Business implements NamedEntity & ValuedEntity {
  name: String
  value: Int
  employeeCount: Int
}
```

tonyghita:
Closing, will make the change in two steps:
1. move the contents of `internal/tests` to `internal/validation`, since that's what the tests are testing.
2. update the validation tests with the most recent from `github.com/graphql/graphql-js`.

PR #168, opened 2018-02-18
add function to calculate data retrieval process based on time per se…

s1gu:
…cond

### DIFF
--- a/example/starwars/starwars.go
+++ b/example/starwars/starwars.go
@@ -6,8 +6,10 @@ package starwars
 import (
 	"encoding/base64"
 	"fmt"
+	"log"
 	"strconv"
 	"strings"
+	"time"
 
 	graphql "github.com/neelance/graphql-go"
 )
@@ -201,12 +203,6 @@ var humans = []*human{
 
 var humanData = make(map[graphql.ID]*human)
 
-func init() {
-	for _, h := range humans {
-		humanData[h.ID] = h
-	}
-}
-
 type droid struct {
 	ID              graphql.ID
 	Name            string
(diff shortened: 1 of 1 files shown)


tonyghita:
Have you looked at the `trace` package? That package defines a `Tracer` interface that can be implemented to provide metrics like duration of query resolution and even field resolution.

```go
type Tracer interface {
	TraceQuery(ctx context.Context, queryString string, operationName string, variables map[string]interface{}, varTypes map[string]*introspection.Type) (context.Context, TraceQueryFinishFunc)
	TraceField(ctx context.Context, label, typeName, fieldName string, trivial bool, args map[string]interface{}) (context.Context, TraceFieldFinishFunc)
} 
```

```go
type exampleTracer struct{
  logger *log.Logger
}

func (t *exampleTracer) TraceField(ctx context.Context, label, typeName, fieldName string, trivial bool, args map[string]interface{}) (context.Context, TraceFieldFinishFunc) {
  // Once registered, this function executes before each field is resolved.
  start := time.Now()

  return ctx, func(err *errors.QueryError) {
    // This function executes after each field is resolved. 
    t.logger.Printf("%s.%s (%q): resolved in %s", typeName, fieldName, label, time.Since(start))
  }
}

```

It would be great if we could work such a tracer implementation into this example instead of only adding duration logging for a single field.

Closing this PR for now, feel free to ask for clarifications and bare with me as I get up to speed on maintaining this repository.

example/starwars/starwars.go
```diff
@@ -332,7 +328,13 @@ func (r *Resolver) Character(args struct{ ID graphql.ID }) *characterResolver {
 }
 
 func (r *Resolver) Human(args struct{ ID graphql.ID }) *humanResolver {
+	startTime := time.Now()
+	for _, h := range humans {
```
tonyghita: Just a note: I would leave the initialization logic to the `init` function, where it executes only when the application starts, rather than on every request.

example/starwars/starwars.go
```diff
@@ -332,7 +328,13 @@ func (r *Resolver) Character(args struct{ ID graphql.ID }) *characterResolver {
...
+		humanData[h.ID] = h
+	}
 	if h := humanData[args.ID]; h != nil {
+		endTime := time.Now()
+		log.Println("duration:", endTime.Sub(startTime).Seconds())
```
tonyghita: Last note, just a word of caution: the `log` loggers will serialize writes to whatever Writer they have been initialized with. Depending on what rate the resolvers are being executed, this may cause your application to begin to deadlock on this log Writer.

PR #171, opened 2018-02-27
Validate uniqueness of type, directive, and entry point names

tinnywang:
Introduces name validation for types, directives, and entry points. The rules are outlined in the GraphQL spec ([1](http://facebook.github.io/graphql/draft/#sec-Schema), [2](http://facebook.github.io/graphql/draft/#sec-Naming-conventions)) and quoted below:

> All types within a GraphQL schema must have unique names. No two provided types may have the same name. No provided type may have a name which conflicts with any built in types (including Scalar and Introspection types).

> All directives within a GraphQL schema must have unique names.

> All types and directives defined within a schema must not have a name which begins with "__" (two underscores), as this is used exclusively by GraphQL’s introspection system.

> Types and fields required by the GraphQL introspection system that are used in the same context as user‐defined types and fields are prefixed with "__" two underscores. This in order to avoid naming collisions with user‐defined GraphQL types. Conversely, GraphQL type system authors must not define any types, fields, arguments, or any other type system artifact with two leading underscores.

Entry points must be named "query", "mutation", or "subscription", which is a rule enforced by Apollo.


internal/schema/schema.go
```diff
@@ -395,57 +418,87 @@ func resolveInputObject(s *Schema, values common.InputValueList) error {
 	return nil
 }
 
-func parseSchema(s *Schema, l *common.Lexer) {
+func parseSchema(s *Schema, l *common.Lexer) error {
```
tonyghita: I think we should keep error returns to a single form here for now—that is, errors are returned via `panic` (which is strange, yes). I believe the intention was to prevent error checking on every `l.ConsumeX()` call.

Either way, I'm not sure that introducing another error path is something we want to do in the short-to-medium term.

pavelnikolov:
I'm closing this PR as it's quite old and the query, mutation and subscription have been changed to [NamedType](https://github.com/graph-gophers/graphql-go/blob/master/ast/types.go#L15) since.

PR #178, opened 2018-03-10
Add a handler for the graphcool graphql-playground ide

dvic:
Not sure if this is wanted/needed, but if so, here it is :)

credits: modified version of https://github.com/vektah/gqlgen/blob/master/handler/playground.go


tonyghita:
Thanks for your contribution @dvic.

I agree with @mytototo that specific handler code probably does not belong in this repository. We've included the standard GraphiQL as an example, but perhaps that one should be pulled out into an example repository instead.

If you do have such an example repository, feel free to submit a PR to link to the repository 👍

Issue #150, opened 2018-01-09
[DISCUSSION] A change about Enum type check.


sanae10001:
Source code:
https://github.com/neelance/graphql-go/blob/5d3b13f2e2215f7d53c88b20fd2c37f0a37b5ffd/internal/exec/packer/packer.go#L120-L124

Change:
``` go
case *schema.Enum:
	if reflectType.Kind() != reflect.String {
		return nil, fmt.Errorf("wrong type, expected %s", reflect.String)
	}
```

So we can write this code:
``` go
type Enum string

const (
	EnumA Enum = "A"
	EnumB Enum = "B"
)

type ExampleInput struct {
	ExampleEnum Enum
}
```

Above code:
```go
type ExampleInput struct {
	ExampleEnum string
}
```
I think it's useful. But I don't sure if there are other effects.
Any suggestions？

GhostRussia:
It would be very good, as now it is necessary to do so:

graphql
```graphql
enum ImageVote {
    down
    neutral
    up
}
```

go
```go
type ImageVote string
const (
	VoteDown    ImageVote = "down"
	VoteNeutral ImageVote = "neutral"
	VoteUp      ImageVote = "up"
)
```

```go
func (r *Resolver) UpdateImageVote(ctx context.Context, args struct {
	ID   graphql.ID
	Vote string
}) (*ImageResolver, error) {
	vote := ImageVote(vote)
	...
}
```

sanae10001:
Merged. Close this. 
See https://github.com/graph-gophers/graphql-go/pull/187

PR #181, opened 2018-03-14
Assert Enum type equal String

sanae10001:
See issue [#150](https://github.com/graph-gophers/graphql-go/issues/150)


tonyghita:
This looks like a simple change. I'm still trying to wrap my head around all of the existing code and I haven't quite got to the reflection parts yet. 

Can you explain why the old approach wasn't sufficient, and why the new approach works?

svanburen:
If you're trying to coerce an input to your app's model, the "enum" must be type `string`. Presumably, with this change, something like:

```
type EnumType string
```

could be used in those models instead. (A more full example / explanation can be found in that issue listed above (#150)).

PR #187, opened 2018-03-26
Assert Enum type equal String

sanae10001:
Change branch.
reopen PR https://github.com/graph-gophers/graphql-go/pull/181


sanae10001:
Updated code and test case, any suggestion？

PR #190, opened 2018-04-04
Validate Max Depth

KevinPike:
This change validates that queries don't exceed a configurable max depth. This is opt-in by setting a max depth using `graphql.MaxDepth`. A `ValidatorTracer` is added, similar to `Tracer`, to track validation errors.


KevinPike:
Type aliasing on that line is more convenience than necessity. We could remove and create a new interface for backwards compatibility.

PR #194, opened 2018-04-11
Use Struct Field Resolver instead of Method

### LINKED ISSUE #28: Use field as resolver instead of method.

### LINKED ISSUE #124: Allow returning string for an ID field?
Right now resolvers of fields with GraphQL type `ID` have to return either a Go `graphql.ID` type (imported from `neelance/graphql-go`), or declare their own ID type, implement `ImplementsGraphQLType` and `UnmarshalGraphQL` for it, and return that type. Aside from this issue there's no other need to import `neelance/graphql-go` in resolver definitions at all.

There is already some precedent, though I'm not sure if it's intentional: field _arguments_ of GraphQL type `ID` are passed correctly to Go resolver functions that expect strings. (See demo below.)

**TL;DR: Would it be possible to allow resolvers for fields of GraphQL type `ID` to simply return strings?**

---

See demo gist here: https://gist.github.com/vergenzt/c4dbc132e101d08e4e711796255742e9

With this schema:
```
schema {
	query: Query
}

type Query {
	stringIdAcceptedAsArgument(userId: ID!): String!
	stringIdNotAcceptedAsReturnValue: ID!
}
```

and these resolvers:
```
type resolver struct{}

func (*resolver) StringIdAcceptedAsArgument(args struct{ UserID string }) string {
	return "Hello, user " + args.UserID + "!"
}

func (*resolver) StringIdNotAcceptedAsReturnValue() string {
	return "123"
}
```

the only error thrown is on the resolver with the string _return_ type.

salman-ahmad:
This fixes #28. Essentially, it reduces lot of boilerplate code by making use of struct field as resolvers instead of creating methods for each struct field. 

Now, methods are required **only when**

1. a struct implements a GraphQL interface. In this case, you create methods for fields which overlap
1. a struct exposes a GraphQL field that accepts additional arguments i.e., `first: Int, last: Int` etc. 
1. there is GraphQL `union` type

By using struct fields as resolvers, one could also benefit from DRY codebase by not having a pair of a struct which look the same (one for GraphQL and one for DB)

### Does this break existing API? 

No. This change is completely transparent and you can continue using your existing codebase. 

### Can I still continue using methods as resolvers?

Yes, by default, it uses method resolvers. 

### How can I use struct fields as resolvers?

When invoking `graphql.ParseSchema()` or `graphql.MustParseSchema()` to parse the schema, you pass in `graphql.UseFieldResolvers()` as an argument. For an example, take a look at `./example/social/server/server.go`

### Additional Notes

- also fixes #124
- for an example, take a look at `./example/social/social.go` & `./example/social/server/server.go`
- Unit tests are missing


salmana1:
@tonyghita, please take a look at this when you get a chance, thanks.

mytototo:
Finally ! Thanks a lot @salmana1 this looks promising.
Any milestone to see support this officially out-of-the-box ?

mbinge:
please merge request!!!

salmana1:
@tonyghita any ETA on when you would have some time to review (and approve) this?

graphql.go
```diff
@@ -73,6 +72,13 @@ type Schema struct {
 // SchemaOpt is an option to pass to ParseSchema or MustParseSchema.
 type SchemaOpt func(*Schema)
 
+// Specifies whether to use struct field resolvers
+func UseFieldResolvers() SchemaOpt {
+	return func(s *Schema) {
+		s.schema.UseFieldResolvers = true
+	}
+}
```
pavelnikolov: I like that you use `SchemaOpt` here. This is awesome! 👍

internal/exec/exec.go
```diff
@@ -173,22 +173,33 @@ func execFieldSelection(ctx context.Context, r *Request, f *fieldToExec, path *p
...
+			if f.field.HasError && !callOut[1].IsNil() {
+				resolverErr := callOut[1].Interface().(error)
+				err := errors.Errorf("%s", resolverErr)
+				err.Path = path.toSlice()
+				err.ResolverError = resolverErr
+				return err
+			}
+		} else {
+			// TODO extract out unwrapping ptr logic to a common place
+			res := f.resolver
```
pavelnikolov: Probably it's a good idea to move `res` before the outer `if` in order to use it on line 185 as well.

internal/exec/exec.go
```diff
@@ -173,22 +173,33 @@ func execFieldSelection(ctx context.Context, r *Request, f *fieldToExec, path *p
...
-		}
-		callOut := f.resolver.Method(f.field.MethodIndex).Call(in)
-		result = callOut[0]
-		if f.field.HasError && !callOut[1].IsNil() {
-			resolverErr := callOut[1].Interface().(error)
-			err := errors.Errorf("%s", resolverErr)
-			err.Path = path.toSlice()
-			err.ResolverError = resolverErr
-			return err
+		if f.field.MethodIndex != -1 {
```
pavelnikolov: I would add a comment on top of this `if` statement. Something like:
```Go
// check if this is a method or field resolver
if f.field.MethodIndex != -1 {
```

internal/exec/resolvable/resolvable.go
```diff
@@ -197,7 +204,9 @@ func makeScalarExec(t *schema.Scalar, resolverType reflect.Type) (Resolvable, er
 	return &Scalar{}, nil
 }
 
-func (b *execBuilder) makeObjectExec(typeName string, fields schema.FieldList, possibleTypes []*schema.Object, nonNull bool, resolverType reflect.Type) (*Object, error) {
+func (b *execBuilder) makeObjectExec(typeName string, fields schema.FieldList, possibleTypes []*schema.Object,
+	nonNull bool, resolverType reflect.Type) (*Object, error) {
+
```
pavelnikolov: Please, remove empty line.

internal/exec/resolvable/resolvable.go
```diff
@@ -207,40 +216,67 @@ func (b *execBuilder) makeObjectExec(typeName string, fields schema.FieldList, p
...
-		if methodIndex == -1 {
+		methodIndex := -1
+		fieldIndex := -1
+
+		// 1) Use resolver type's method when
+		//	1.1) __Type and __Schema requests
+		// 	1.2) Or field has arguments
+		//	1.3) Or it is configured to use method
+		//	1.4) Or it is an interface
+		// 2) Otherwise use resolver type's field
```
pavelnikolov: Please, reformat the comment as:
```
		// 1. Use method resolver when:
		//	a. __Type and __Schema requests; or
		// 	b. The requested field has arguments; or
		//	c. The schema is not configured to use field resolvers; or
		//	d. It is an interface
		// 2. Otherwise, use a field resolver
```

internal/exec/resolvable/resolvable.go
```diff
@@ -207,40 +216,67 @@ func (b *execBuilder) makeObjectExec(typeName string, fields schema.FieldList, p
...
+		if isResolverSchemaOrType(rt) == true || len(f.Args) > 0 ||
+			b.schema.UseFieldResolvers == false || rt.Kind() == reflect.Interface {
```
pavelnikolov: `UseFieldResolver` is already a boolean, right? Can't you just use `!b.schema.UseFieldResolvers`? Or you want it to be more verbose?

internal/exec/resolvable/resolvable.go
```diff
@@ -207,40 +216,67 @@ func (b *execBuilder) makeObjectExec(typeName string, fields schema.FieldList, p
...
-		}
-		if resolverType.Method(methodIndex).Type.NumOut() != 2 {
-			return nil, fmt.Errorf("%s does not resolve %q: method %q should return a value and a bool indicating success", resolverType, typeName, "To"+impl.Name)
-		}
-		a := &TypeAssertion{
-			MethodIndex: methodIndex,
-		}
-		if err := b.assignExec(&a.TypeExec, impl, resolverType.Method(methodIndex).Type.Out(0)); err != nil {
-			return nil, err
+	if b.schema.UseFieldResolvers == false || resolverType.Kind() != reflect.Interface {
```
pavelnikolov: again `<bool> == false` vs `!<bool`? Do we want the more verbose version? IMHO, it's better to use `!`...

internal/exec/resolvable/resolvable.go
```diff
@@ -253,66 +289,95 @@ func (b *execBuilder) makeObjectExec(typeName string, fields schema.FieldList, p
...
+	if err := b.assignExec(&fe.ValueExec, f.Type, out); err != nil {
 		return nil, err
 	}
+
 	return fe, nil
 }
 
+// find method with the same name or one which has `Resolver` suffix
 func findMethod(t reflect.Type, name string) int {
+	resName := name + "Resolver"
```
pavelnikolov: This is completely unrelated change (i.e. find method with name `<field>Resolver`). Please, send this in a separate PR. I'd personally, vote 👎 for this specific change, since adds unnecessary complexity to the library. Less is more in this case.

internal/schema/schema.go
```diff
@@ -45,6 +45,8 @@ type Schema struct {
 	objects         []*Object
 	unions          []*Union
 	enums           []*Enum
+
+	UseFieldResolvers bool
```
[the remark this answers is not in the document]
pavelnikolov: How about grouping of public/private fields? Should this be private?

pavelnikolov:
@tonyghita, please review this change as it is opt-in and would save us from writing a lot of boilerplate code. 🎉 🆒 😎 👍 ❤️ 

I'm curious to see the benchmark comparison between the field and method resolvers. 📈 📉 📊

pavelnikolov:
@neelance FYI

internal/exec/resolvable/resolvable.go
salmana1: imho, `!` can be overlooked sometime so more verbose make it clear that it's a check for `false`. I can change this to `!`

internal/exec/resolvable/resolvable.go
salmana1: This is not an unrelated change. As mentioned in the PR notes, when a struct implements an interface, then struct's resolver methods are used for each interface field. Hence, when a struct has a field `Email` and interface has the same field, then there will be a conflict between struct field and resolver method. So a `Resolver` suffix can be used to resolve that conflict. If you can suggest something else to achieve the same goal then I will be happy to implement that

salmana1:
@pavelnikolov thanks for reviewing the code and I will fix most of them. 

I totally understand what you are saying about multiple PRs. Please note that, this PR is coming from downsteam `master` branch which has bunch of fixes and we needed them for our usecase. I do not have the permission to change dowstream branch so I will try to do some git magic to avoid multiple things in this PR.

> resolving IDs as strings (I'd vote 👎 )

I think this is useful change and it's working well for my company. We don't have to wrap or unwrap `graphql.ID` to `string` every time. But I can create a separate PR for it.

> <field_name>Resolver to be used if available (I'd vote 👎 )

This is relevant to this PR and I have replied to your review comment

internal/exec/resolvable/resolvable.go
pavelnikolov: I prefer using `!`, but it's up to the owners of the library to decide.

internal/exec/resolvable/resolvable.go
pavelnikolov: As a workaround you can just use different struct fields. If you are implementing an interface, it's just the methods that will be used, right? This would mean that you can use differnt field names.

pavelnikolov:
> This is relevant to this PR and I have replied to your review comment

I'm still not 100% convinced that this is necessary. You can definitely change the field names in the struct or use a different struct and still implement an interface. I think that it should live in a separate PR.

pavelnikolov:
I'll create a test schema and will run benchmarks comparing the `master` branch against your branch with both fields and methods

spiffyjr:
Can this get some attention? This is pretty handy for reducing the amount of boilerplate needed.

iamtakingiteasy:
+1, This feature would be very practical.

eveld:
In large applications this would take away a lot of boilerplate code. Is there anything that still needs to happen for this to be merged?

pavelnikolov:
@tonyghita is this in your roadmap? Any feedback about this PR?

marcinwyszynski:
Thanks for working on that @salmana1 ! 🙇🏼‍♂️

pavelnikolov:
@salmana1 
> I totally understand what you are saying about multiple PRs. Please note that, this PR is coming from downsteam master branch which has bunch of fixes and we needed them for our usecase. I do not have the permission to change dowstream branch so I will try to do some git magic to avoid multiple things in this PR.

Please, I'd be happy to consider the struct field resolvers once you split this PR.

bazaglia:
+1!!

This is a must-have for reducing drastically the codebase.

internal/exec/resolvable/resolvable.go
pavelnikolov: I prefer using `!` because it'll be consistent with the rest of the code in this library.

pavelnikolov:
@salmana1 are you willing to split this PR? I'd be happy to review/merge the field resolvers.

bazaglia:
@salmana1 Let me know if you need any help.

Also, when passing `graphql.UseFieldResolvers()`, it ALWAYS uses the value from struct, except in the cases you described (like a field has arguments). What I expected was to use the struct value only if `fieldResolver` doesn't exists. Is that possible at all with the current implementation?

pavelnikolov:
@bazaglia Thanks for the feedback. Any feedback is highly appreciated! Can you please add some samples / unit-tests?

salmana1:
Sorry for not replying sooner. 

I will try to update this PR soon. I don't have access to the original forked branch; so, I will probably have to close this PR and recreate it. 

On a separate note, I would recommend to take a look at [gqlgen(https://github.com/99designs/gqlgen) library. At my last job, we switched to that library after we realized this library was not actively developed and PRs weren't getting merged.

pavelnikolov:
Thanks for your contribution. If you don't plan to resubmit this PR, please let us know because there are a lot of people who are interested in it.
This project is now actively maintained again and PRs/issues are responded promptly.

tobstarr:
I would really like to see this PR getting merged!

pavelnikolov:
@tobstarr I am thinking of an implementation that would keep the current implementation as default but would allow the users of the library to override the `findMethod` implementation as they see fit. I'll post an update soon. Chances are that this PR will not be merged in its current form.

pavelnikolov:
@tobstarr I am thinking of an implementation that would keep the current method implementation as default but would allow the users of the library to override the resolving implementation as they see fit. I'll post an update soon. Chances are that this PR will not be merged in its current form.

bazaglia:
@pavelnikolov nice, that's my upvote also, just like I commented before.

salman-ahmad:
I will be pushing out a cleaned up PR tomorrow and will close this PR

salman-ahmad:
Closing this PR - please see the new PR #282

PR #202, opened 2018-05-03
print the actual errors in checkErrors when it's appropriate

timfeirg:
Current implementation doesn't print any errors when there's no expected errors, it just print two numbers which makes it difficult to debug.

image

Using this PR, errors will be printed, which looks like this:

image

### DIFF
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -75,12 +75,7 @@ func formatJSON(data []byte) ([]byte, error) {
 }
 
 func checkErrors(t *testing.T, expected, actual []*errors.QueryError) {
-	expectedCount, actualCount := len(expected), len(actual)
-
-	if expectedCount != actualCount {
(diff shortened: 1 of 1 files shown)


timfeirg:
RFC @tonyghita

gqltesting/testing.go
```diff
@@ -75,12 +75,7 @@ func formatJSON(data []byte) ([]byte, error) {
 }
 
 func checkErrors(t *testing.T, expected, actual []*errors.QueryError) {
-	expectedCount, actualCount := len(expected), len(actual)
-
-	if expectedCount != actualCount {
-		t.Fatalf("unexpected number of errors: got %d, want %d", actualCount, expectedCount)
-	}
-
```
pavelnikolov: This code should not be deleted.

pavelnikolov:
It looks like this code is already in the master branch so I'll close this PR

Issue #15, opened 2016-11-10
What about new features like a query batching and subscriptions


DenisNeustroev:
https://dev-blog.apollodata.com/new-features-in-graphql-batch-defer-stream-live-and-subscribe-7585d0c28b07

neelance:
Definitely interested! Is there any indication on when this will end up in the official specification or some other spec document?

DenisNeustroev:
No, but Facebook uses these internally. The Apollo (client and server) also supports both features, the official GraphQL server supports only subscriptions.

neelance:
Okay. I'm not against those features, but there are other things I need to take care of first. Effectively this means that those features will either be added when I need them myself or via PR.

paralin:
I'd like to help implement these, ~maybe~

In my project, I do a pipeline where I start out defining the schema in Protobuf, and it gets converted into a variety of different things (TypeScript definitions, GraphQL schema).

Right now I'm looking at an arms race, really:

 - [PlayLyfe](github.com/playlyfe/go-graphql) - got my relatively large project working with some intermediate code-generation and interface manipulations (I generate a function to get the GraphQL schema type name for all of my ProtoBuf schemas). The code base is shit, though. I don't see adding subscriptions to that being a very easy thing.
 - [Go-Graphql](https://github.com/graphql-go/graphql) - massive, seems well made, but pretty inactive and for some reason there is no documentation / any way to convert a schema AST into a Go schema. Supports subscriptions (!)
 - [This repo](https://github.com/neelance/graphql-go) - nicely written, requires some intermediate code generation from me, but otherwise I like it more than the others. The main difficulty here is I have to do [this stuff](https://github.com/neelance/graphql-go/blob/master/example/starwars/starwars.go#L394) which I really hate. I can code-gen most of this, but it just seems like a mess. Some reflection could work better.

Either way, if we could support subscriptions in this repo I think I would probably adopt it for my projects. I'm going to try to write the code-gen stuff in my codebase to make everything work as-is with this repo. If I can do that, I'd love to help implement subscriptions.

paralin:
I've ported all my stuff to run on this library now, nice! Requires a LOT of code generation to cover trivial member field struct resolvers, but it's worth it I think. 

As for subscriptions, where are we on that? I can have a look at what would be required to add them in. Thoughts on what the API could look like @neelance ?

I'm thinking `Exec(` signature is the same, but it just doesn't return until the subscription terminates or the context terminates.

[...]

PR #203, opened 2018-05-05
Support subscriptions

matiasanaya:
This addresses the subscription part of #15.

Please look through `subscription_test.go` which has an example subscription resolver implemented.

I've left `// TODO: ...` markers where _I feel_ the implementation could be improved, though not sure how.

Missing from this PR is tracing for the execution of each query.

I've put together a [graphQL server that supports subscriptions](https://github.com/matiasanaya/go-graphql-subscription-example) as a proof of concept based on these changes (though it might be out of sync since this PR is changing)

### DIFF
--- a/internal/exec/subscribe.go
+++ b/internal/exec/subscribe.go
@@ -1,0 +1,147 @@
+package exec
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/graph-gophers/graphql-go/errors"
+	"github.com/graph-gophers/graphql-go/internal/exec/resolvable"
+	"github.com/graph-gophers/graphql-go/internal/exec/selected"
+	"github.com/graph-gophers/graphql-go/internal/query"
+)
+
+type Response struct {
+	Data   json.RawMessage
+	Errors []*errors.QueryError
+}
+
+func (r *Request) Subscribe(ctx context.Context, s *resolvable.Schema, op *query.Operation) <-chan *Response {
+	var result reflect.Value
+	var f *fieldToExec
+	var err *errors.QueryError
+	func() {
+		defer r.handlePanic(ctx)
+
+		sels := selected.ApplyOperation(&r.Request, s, op)
+		var fields []*fieldToExec
+		collectFieldsToResolve(sels, s.Resolver, &fields, make(map[string]*fieldToExec))
+
+		// TODO: move this check into validation.Validate
+		if len(fields) != 1 {
+			err = errors.Errorf("%s", "can subscribe to at most one subscription at a time")
+			return
+		}
+		f = fields[0]
+
+		var in []reflect.Value
+		if f.field.HasContext {
+			in = append(in, reflect.ValueOf(ctx))
+		}
+		if f.field.ArgsPacker != nil {
+			in = append(in, f.field.PackedArgs)
+		}
+		callOut := f.resolver.Method(f.field.MethodIndex).Call(in)
+		result = callOut[0]
+
+		if f.field.HasError && !callOut[1].IsNil() {
+			resolverErr := callOut[1].Interface().(error)
+			err = errors.Errorf("%s", resolverErr)
+			err.ResolverError = resolverErr
+		}
+	}()
+
+	if err != nil {
+		return sendAndReturnClosed(&Response{Errors: []*errors.QueryError{err}})
+	}
+
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		return sendAndReturnClosed(&Response{Errors: []*errors.QueryError{errors.Errorf("%s", ctxErr)}})
+	}
+
+	c := make(chan *Response)
+	// TODO: handle resolver nil channel better?
+	if result == reflect.Zero(result.Type()) {
+		close(c)
+		return c
+	}
+
+	go func() {
+		for {
+			// Check subscription context
+			chosen, resp, ok := reflect.Select([]reflect.SelectCase{
+				{
+					Dir:  reflect.SelectRecv,
+					Chan: reflect.ValueOf(ctx.Done()),
+				},
+				{
+					Dir:  reflect.SelectRecv,
+					Chan: result,
+				},
+			})
+			switch chosen {
+			// subscription context done
+			case 0:
+				close(c)
+				return
+			// upstream received
+			case 1:
+				// upstream closed
+				if !ok {
+					close(c)
+					return
+				}
+
+				subR := &Request{
+					Request: selected.Request{
+						Doc:    r.Request.Doc,
+						Vars:   r.Request.Vars,
+						Schema: r.Request.Schema,
+					},
+					Limiter: r.Limiter,
+					Tracer:  r.Tracer,
+					Logger:  r.Logger,
+				}
+				var out bytes.Buffer
+				func() {
+					// TODO: configurable timeout
+					subCtx, cancel := context.WithTimeout(ctx, time.Second)
+					defer cancel()
+
+					// resolve response
+					func() {
+						defer subR.handlePanic(subCtx)
+
+						out.WriteString(fmt.Sprintf(`{"%s":`, f.field.Alias))
+						subR.execSelectionSet(subCtx, f.sels, f.field.Type, &pathSegment{nil, f.field.Alias}, resp, &out)
+						out.WriteString(`}`)
+					}()
+
+					if err := subCtx.Err(); err != nil {
+						c <- &Response{Errors: []*errors.QueryError{errors.Errorf("%s", err)}}
+						return
+					}
+
+					// Send response within timeout
+					// TODO: maybe block until sent?
+					select {
+					case <-subCtx.Done():
+					case c <- &Response{Data: out.Bytes(), Errors: subR.Errs}:
+					}
+				}()
+			}
+		}
+	}()
+
+	return c
+}
+
+func sendAndReturnClosed(resp *Response) chan *Response {
+	c := make(chan *Response, 1)
+	c <- resp
+	close(c)
+	return c
+}
--- a/subscription_test.go
+++ b/subscription_test.go
@@ -1,0 +1,196 @@
+package graphql_test
+
+import (
+	"context"
+	"encoding/json"
+	stdErrors "errors"
+	"testing"
+
+	graphql "github.com/graph-gophers/graphql-go"
+	"github.com/graph-gophers/graphql-go/errors"
+	"github.com/graph-gophers/graphql-go/gqltesting"
+)
+
+type rootResolver struct {
+	*helloResolver
+	*helloSaidResolver
+}
+
+type helloResolver struct{}
+
+func (r *helloResolver) Hello() string {
+	return "Hello world!"
+}
+
+var resolverErr = stdErrors.New("resolver error")
+
+type helloSaidResolver struct {
+	err      error
+	upstream <-chan *helloSaidEventResolver
+}
+
+type helloSaidEventResolver struct {
+	msg string
+	err error
+}
+
+func (r *helloSaidResolver) HelloSaid(ctx context.Context) (chan *helloSaidEventResolver, error) {
+	if r.err != nil {
+		return nil, r.err
+	}
+
+	c := make(chan *helloSaidEventResolver)
+	go func() {
+		for r := range r.upstream {
+			select {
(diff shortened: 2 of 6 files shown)


internal/exec/subscribe.go
```diff
@@ -0,0 +1,147 @@
...
+
+					if err := subCtx.Err(); err != nil {
+						c <- &Response{Errors: []*errors.QueryError{errors.Errorf("%s", err)}}
+						return
+					}
+
+					// Send response within timeout
+					// TODO: maybe block until sent?
+					select {
+					case <-subCtx.Done():
```
pavelnikolov: Should you send a timeout error here?

subscription_test.go
```diff
@@ -0,0 +1,196 @@
+package graphql_test
+
+import (
+	"context"
+	"encoding/json"
+	stdErrors "errors"
```
pavelnikolov: I'd personally add alias to the other `"errors"` package and leave the standard lib one as is.

subscription_test.go
```diff
@@ -0,0 +1,196 @@
...
+						&helloSaidEventResolver{msg: "Hello again!"},
+					),
+				},
+			}),
+			Query: `
+				subscription onHelloSaid {
+					helloSaid {
+            msg
+          }
+				}
```
pavelnikolov: Please, fix indentation above.

subscription_test.go
```diff
@@ -0,0 +1,196 @@
...
+	}()
+
+	return c, nil
+}
+
+func (r *helloSaidEventResolver) Msg() (string, error) {
+	return r.msg, r.err
+}
+
+func closedUpstream(rr ...*helloSaidEventResolver) <-chan *helloSaidEventResolver {
```
pavelnikolov: Should this be named `closedChannel` instead? For a moment, I got confused about the meaning of `upstream` here.

subscription_test.go
```diff
@@ -0,0 +1,196 @@
...
+		{
+			Name: "subscription_resolver_can_error",
+			Schema: graphql.MustParseSchema(schema, &rootResolver{
+				helloSaidResolver: &helloSaidResolver{err: resolverErr},
+			}),
+			Query: `
+				subscription onHelloSaid {
+					helloSaid {
+		        msg
+		      }
```
pavelnikolov: Fix indentation.

subscription_test.go
```diff
@@ -0,0 +1,196 @@
...
+			},
+		},
+		{
+			Name:   "schema_without_resolver_errors",
+			Schema: &graphql.Schema{},
+			Query: `
+				subscription onHelloSaid {
+					helloSaid {
+		        msg
+		      }
```
pavelnikolov: Fix indentation.

subscription_test.go
```diff
@@ -0,0 +1,196 @@
...
+    helloSaid: HelloSaidEvent!
+  }
+
+  type HelloSaidEvent {
+    msg: String!
+  }
+
+	type Query {
+		hello: String!
+	}
```
pavelnikolov: Fix indentation.

subscription_test.go
```diff
@@ -0,0 +1,196 @@
...
+			`,
+			ExpectedErr: stdErrors.New("schema created without resolver, can not subscribe"),
+		},
+	})
+}
+
+const schema = `
+  schema {
+    subscription: Subscription,
+		query: Query
```
pavelnikolov: Fix indentation.

subscription_test.go
```diff
@@ -0,0 +1,196 @@
...
+			Schema: graphql.MustParseSchema(schema, &rootResolver{}),
+			Query:  `invalid graphQL query`,
+			ExpectedResults: []gqltesting.TestResponse{
+				{
+					Errors: []*errors.QueryError{errors.Errorf("%s", `syntax error: unexpected "invalid", expecting "fragment" (line 1, column 9)`)},
+				},
+			},
+		},
+		{
+			Name:   "subscribe_to_query_errors",
```
pavelnikolov: Is it ever possible to get into this case in an actual application? The GraphQL lib will execute a Query if the operation is `query`, no? I don't see the reason for this test. Am I missing something!

pavelnikolov:
This is a good start! Thank you @matiasanaya!
I know it is outside of the scope of this PR, but it would be really nice to add a subscription to the StarWars example!

internal/exec/subscribe.go
matiasanaya: Don't believe it would make sense to try and send a timeout error since we've time'd out waiting to send an actual response. But I do think maybe this should be checking against `ctx.Done()` instead. Thoughts?

subscription_test.go
matiasanaya: done

subscription_test.go
matiasanaya: done

subscription_test.go
matiasanaya: done

subscription_test.go
matiasanaya: done

subscription_test.go
matiasanaya: done

subscription_test.go
matiasanaya: done

subscription_test.go
matiasanaya: This patch exposes `schema.Subscribe(ctx context.Context, queryString string, operationName string, variables map[string]interface{}) (<-chan *Response, error)` which will be leveraged by the transport. The queryString could contain anything. I'm not sure if it makes sense to leave this job to the transport layer.

Xeizzen:
Hello everyone who are pursuing the endeavor of implementing subscriptions. How it is going recently? Do you need assistance?

tsauvajon:
We strongly need subscriptions aswell, we would be glad to help if any work remains before merging this PR!

matiasanaya:
@Xeizzen @tsauvajon - I've written a new [transport library](https://github.com/matiasanaya/graphql-transport-ws) that aligns with this library and updated the [subscription example server](https://github.com/matiasanaya/go-graphql-subscription-example) with a more robust implementation. If you are keen, please have a look at those projects, try them out and contribute :)

tsauvajon:
LGTM, I tried it in our project and it works as planned.
I've used you "say hello" exemple only at the moment (https://github.com/poudre-aux-yeux/rapiquette/commit/271f9c7c88e8d449c17b07ac2c57007311c444c4), but I will share when I go further.

tsauvajon:
The thing I don't really understand is

```go
select {
  case r.pointScoredEvents <- e:
  case <-time.After(1 * time.Second):
}
```

Why use a `select`, and why this `time.After` ?

pavelnikolov:
The above select statement will block until one of these happen:
1) either a message is received from the `r.pointScoredEvents` channel (if there is no buffer or buffer capacity); or
2) the operation times out after 1 second

ande8133:
We also really need subscriptions. I was wondering how close this is to being checked in. =)

tsauvajon:
@ande8133 check @matiasanaya 's [link](https://github.com/matiasanaya/go-graphql-subscription-example). Subscriptions work for me following his example.

The only remaining things I have to do is changing the handler to enable CORS and return a 200 on an OPTION request (preflight). I've never really written webservers in Go so I need to learn that but that's not a problem with matiasanaya's work, just with my current knowledge.

stephenafamo:
Is there a reason why this is not yet merged?
Is any help needed? 
Everything seems to work fine.

pavelnikolov:
I was wondering the same...

bsr203:
@tonyghita @neelance  please consider spending as it been the most requested feature and many claim works well. thank you.

smithaitufe:
IMHO, It seems PRs are not appreciated.

bsr203:
@smithaitufe @neelance  commented [here](https://github.com/graph-gophers/graphql-go/issues/88#issuecomment-348370328) as he don't have the bandwidth for some time. I wish more capable maintainers are identified :-( .

stephenafamo:
Is there no one else maintaining this project? Is it just @neelance ?

neelance:
I am not maintaining this project any more, since I'm currently (unfortunately) not working on any project that uses GraphQL. Therefore I'm a bit detached from the GraphQL ecosystem atm and someone else would probably do a better job at maintaining this project. @tonyghita already volunteered, but we can add another helping hand if someone wants to take the job. ;-)

pavelnikolov:
I would like to volunteer. I have been following the project very carefully and my team has been using it in production for more than 1.5 years. We are interested in subscriptions but also extensions and other GraphQL features. @matiasanaya (the author of this PR) is also from my team and I peer-tested his work long time ago. I'm a little disappointed that there hasn't been much activity lately. 
If you accept contributors/maintainers I'd be happy to participate/help.

stephenafamo:
I'll be happy to, although since I've not contributed to this project before, I don't know if I am the best person to do it.

However, I am working heavily with something involving GraphQL, and will be for a while.

pavelnikolov:
Adding this project to the #gfrs list in the gophers Slack might be a good idea. This way it can be trully community managed! 
IMHO, This project is awesome! The only problem is that it is not maintained regularly. And there isn't much interaction with potential contributors/volunteers.

stephenafamo:
I agree @pavelnikolov

neelance:
I'd like to have a single primary maintainer for the project. The primary maintainer can add secondary maintainers if desired.

@pavelnikolov Using it for 1.5 years sounds like a good qualification. :)

@tonyghita Would it be okay with you to make @pavelnikolov the primary maintainer?

stephenafamo:
I think this would be great.

pavelnikolov:
Thank you for trusting in me! This is an awesome privilege and I willing take this responsibility seriously. 
I’m going on a quick holiday and when I’m back next Tuesday subscriptions will be my first priority. 
@matiasanaya let’s catch up early next week re subscriptions, ws transport and example(s).

romshark:
As I read about websocket transport I'd like to ask this question: **will the library remain transport-layer agnostic?** Will we be able to use graphql-go with subscriptions support on top of any custom transport layer?

I'm worried because we're using a home-grown websocket abstraction library called [webwire](https://github.com/qbeon/webwire-go) together with [graph-gophers/graphql-go](https://github.com/graph-gophers/graphql-go) and we currently implemented subscriptions without GraphQL support just by using [webwire server-side signals](https://github.com/qbeon/webwire-go#server-side-signals) and regular mutations

pavelnikolov:
@romshark I think that the library has to remain transport agnostic and as minimalistic as possible! The WS transport will be added as a separate repository under `graph-gophers`.

lindroth:
@matiasanaya does this library use one websocket for all the subscriptions that the client has, or is it multiple connections?

I know that with apollo it's possbile to send queries and mutations via websocket as well.

pavelnikolov:
@lindroth the library itself is transport agnostic. The subscriptions updates are sent over Go channel. It's up to the user of the library to decide how to handle it. There is a reference websocket implementation that @matiasanaya worked on and send a PR https://github.com/graph-gophers/graphql-transport-ws/pull/1
I'm reviewing both PRs now.

matiasanaya:
@lindroth if you end up using [graph-gophers/graphql-transport-ws](https://github.com/graph-gophers/graphql-transport-ws/pull/1) (once it's merged), one websocket can be used to handle all subscriptions, queries and mutations. It's based on the apollo implementation. In the end though it's going to be up to the client you use.

hampusohlsson:
Awesome work on this!

> one websocket can be used to handle all subscriptions, queries and mutations

@matiasanaya Trying to do exactly what you're suggesting here - use websockets for not just subscriptions but querying as well (to lower the number of http requests from the browser), but the implementation seems to explicitly prevent that.

https://github.com/graph-gophers/graphql-go/blob/940d2b01f2549ee5d9e87141ea909134bf56e3a9/subscriptions.go#L48-L50

I'm using [apollo-link-ws](https://github.com/apollographql/apollo-link/tree/master/packages/apollo-link-ws) on the browser side. Am I missing something?

pavelnikolov:
@hampusohlsson Thanks for your comment. I haven't used `apollo-link-ws` but maybe it's worth raising a separate issue.

matiasanaya:
@hampusohlsson I believe you are right and that I've, at some point, sneaked in this limitation. Care to raise a separate issue so we can track the work to support queries and mutations through subscriptions?

hampusohlsson:
No problem! Opened an issue here #286

Issue #60, opened 2017-02-18
More info on errors


EmperorEarth:
I've been using [Dave Cheney's errors pkg](https://github.com/pkg/errors) to add stack traces to all my errors. In dev, when I'm testing queries in graphiql/my app, I like to return the error directly in the response. I was wondering if there was any way to return the `%+v` version of errors back instead of the `%s` version, because right now, I'm not getting the line numbers of various errors and the rest of the stack trace.

neelance:
See #49. I think the same solution should be possible in your case. If yes, then I should add some documentation to the readme since this question came up several times now.

neelance:
Oops, I referenced the wrong issue in the commit. Reopening.

neelance:
@EmperorEarth Any response to my comment above?

tonyghita:
Not to hijack the issue, but the addition of query paths on errors to the specification seems near. See: https://github.com/facebook/graphql/pull/230

At first I was thinking of adding the `path` value after the query has executed and before we respond to the user, but I think this might be more appropriate in the library.

EmperorEarth:
@neelance So sorry! I'll spend some time on this today and tomorrow.

I've been working on my startup prototype solo so long, I've gotten out of the habit of checking on Github/Slack.

EmperorEarth:
@neelance, if I understand it right, it doesn't solve what I'm trying to do because the param is still `%s` in Printf, not `%+v`. But I don't really understand it at all, so?

As a basic example, as I'm rapidly writing code, sometimes I forget to check `err == sql.ErrNoRows()` before I check `err != nil` to return a `nil` resolver. With Cheney's `errors` package, if I wrap all my errors with the `func` `WithStack`, I'll get a line number.

I return all errors, and in my non-resolver code, only log them at the root. With graphql, however, we aggregate errors and proceed on the rest of the resolution as best we can. This is great, except when we return back to the client, in the client's `fetch().catch()`, `console.error(response.error)` gets an array of strings with very little contextual information. It gets very tedious to go through the response to see what did/didn't get returned. It would be helpful if the response to the client had the errors printed using `%+v` instead of `%s`, so the strings would have a line numbers.

Does that provide more context?

EmperorEarth:
@neelance, if I understand it right, it doesn't solve what I'm trying to do because the param is still `%s` in Printf, not `%+v`. But I don't really understand it at all, so maybe it's better if I try to explain the yak I'm shaving?

[...]

PR #207, opened 2018-05-11
Add support for custom error handlers

gracenoah:
This PR aims at being a more concise way of handling errors while resolving a query.

It relates to https://github.com/graph-gophers/graphql-go/issues/60 and https://github.com/graph-gophers/graphql-go/pull/49. The latter has no documentation at all so it is unclear how to interact with it.

Also, this PR adds tests to this feature which may also be used as a reference. Motivation behind this:

## Augment/Hide Implementation Details

If you use a stacktrace library you may not want to share that information with API consumers, or maybe you want to add more information to the error

## Server-side Logging

graphql-go swallows the errors and nothing gets logged on the server side. This gives developers the opportunity to do that too.

--

An error handler must be of type `type ErrorHandler func(error) *QueryError`. A default handler is provided making this feature opt-in and backwards compatible.

The basic principle is to use `SchemaOpts` to augment the schema. Basic usage:

```go
graphql.MustParseSchema(sampleSchemaStr, someResolvers, graphql.ErrorHandler(func(err error) *gqlerrors.QueryError {
    return gqlerrors.Errorf("Prefix - %s", err)
}))
```

Hope it helps.

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -17,6 +17,16 @@ type Location struct {
 	Column int `json:"column"`
 }
 
+// ErrorHandler describes a function that converts an error to a QueryError
+type ErrorHandler func(error) *QueryError
+
+// DefaultErrorHandler returns the default error handler
+func DefaultErrorHandler() ErrorHandler {
+	return func(err error) *QueryError {
+		return Errorf("%s", err)
+	}
+}
+
 func (a Location) Before(b Location) bool {
 	return a.Line < b.Line || (a.Line == b.Line && a.Column < b.Column)
 }
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"reflect"
 	"strconv"
 	"testing"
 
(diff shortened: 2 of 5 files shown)


gracenoah:
After looking a little bit more into https://github.com/graph-gophers/graphql-go/pull/49 I managed to achieve the same outcome with it (i.e. using the framework as it exists now).

The simple concept that I was missing was that `relay.Handler` may be replaced with whatever the user likes, so you may actually completely customize it.


```go
type Handler struct {
	Schema *graphql.Schema
}

func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	var params struct {
		Query         string                 `json:"query"`
		OperationName string                 `json:"operationName"`
		Variables     map[string]interface{} `json:"variables"`
	}
	if err := json.NewDecoder(r.Body).Decode(&params); err != nil {
		http.Error(w, err.Error(), http.StatusBadRequest)
		return
	}

	response := h.Schema.Exec(r.Context(), params.Query, params.OperationName, params.Variables)

        // magic below
	if response.Errors != nil {
		for _, rErr := range response.Errors {
			rErr.Message = fmt.Sprintf("Prefix - %s", rErr.ResolverError)
		}
	}

	responseJSON, err := json.Marshal(response)
	if err != nil {
		http.Error(w, err.Error(), http.StatusInternalServerError)
		return
	}

	w.Header().Set("Content-Type", "application/json")
	w.Write(responseJSON)
}
```
Having said that, it may render this PR moot. Feedback appreciated.

tonyghita:
You're exactly right that the `relay.Handler` is not required to be used--it's provided as a convenience or example to work off.

Modifying the `response.Errors` after execution is the correct way to get change anything related to those errors necessary before returning them to the API consumer.

PR #219, opened 2018-05-30
help wanted: stop returning null data if error is returned

kevdonk:
The spec clearly allows for the case where both data and errors are returned:
http://facebook.github.io/graphql/October2016/#sec-Response-Format

See this comment for an example from the sample graphql-js implementation:
https://github.com/graphql-go/graphql/issues/103#issuecomment-172276590

The code here:
https://github.com/graph-gophers/graphql-go/blob/1a11622ce658bdf2f7a63c80112d76e4794706ee/internal/exec/exec.go#L199-L203

automatically writes "null" for data if an error is returned.

To replicate, simply take any working schema and return an error in your root resolver, along with the data. If the error is returned, data will be null.

This worked for my use case, but I haven't tested it extensively (or written tests). I am concerned that a bunch of error cases that I'm not encountering will make it impossible to return proper data, and curious how that is handled.
Any advice on how/what to test or change, @tonyghita ?
I haven't looked too extensively at the code, but maybe there is sufficient error handling in `execSelectionSet` to prevent it from blowing up?

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -198,8 +198,6 @@ func execFieldSelection(ctx context.Context, r *Request, f *fieldToExec, path *p
 
 	if err != nil {
 		r.AddError(err)
-		f.out.WriteString("null") // TODO handle non-nil
-		return
 	}
 
 	r.execSelectionSet(traceCtx, f.sels, f.field.Type, path, result, f.out)


tonyghita:
@kevdonk thanks for opening this PR. 

According to the specification linked,

> If an error was encountered during the execution that prevented a valid response, the data entry in the response should be null.

This seems like the correct behavior according to the specifcation. Can you link the section the part that says both the `data` entry _and_ `errors` entry can be present?

There is a specification functionality that remains unimplemented in this library... currently we don't propagate non-null field errors to the nearest nullable parent. This is probably because we write the response bytes as we execute the query.

I _think_ that in order to implement this functionality, we'll have to decouple response writing from request execution, only writing the response once execution has finished.

kevdonk:
Under Errors: http://facebook.github.io/graphql/October2016/#sec-Errors
>If the data entry in the response is not null, the errors entry in the response may contain any errors that occurred during execution. If errors occurred during execution, it should contain those errors.

Between Data (7.2.1) and Errors (7.2.2), it has the following requirements:

>If an error was encountered before execution begins, the data entry should not be present in the result.

>If an error was encountered during the execution that prevented a valid response, the data entry in the response should be null.

>If the data entry in the response is null or not present, the errors entry in the response must not be empty. It must contain at least one error. The errors it contains should indicate why no data was able to be returned.

**The main quote you requested**

>If the data entry in the response is not null, the errors entry in the response may contain any errors that occurred during execution. If errors occurred during execution, it should contain those errors.

I believe that the combination of these means that you should return what valid data you can, along with errors. I think you might be right that in order to satisfy the spec, we need to decouple the request execution from the response writing.

pavelnikolov:
This PR has been outdated and the error response has changed significantly since this PR. I'm closing it. If you still encounter the issue, please, submit a failing unit test first.

PR #220, opened 2018-06-01
add GetFieldName Support

tsingsun:
support Get + fieldname to support other model define protocol like google protobuf or some orm which generate code use Getter and Setter

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -268,7 +268,16 @@ func (r *Request) execSelectionSet(ctx context.Context, sels []selected.Selectio
 
 	case *schema.Scalar:
 		v := resolver.Interface()
-		data, err := json.Marshal(v)
+		var (
+			data []byte
+			err  error
+		)
+		if m, ok := v.(json.Marshaler); ok {
+			data, err = m.MarshalJSON()
+		} else {
+			data, err = json.Marshal(v)
+		}
+
 		if err != nil {
 			panic(errors.Errorf("could not marshal %v", v))
 		}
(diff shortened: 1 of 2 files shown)


pavelnikolov:
This feels like magic to me. It'd definitely save some lines of code, but then others might request adding support for more protocols/ORMs. IMHO, the library should remain minimalistic and agnostic of the underlying data/protocol. I definitely, like the idea about using protobuf structs as resolvers, but do we really need this in the library? I'm more leaning towards _No_...
It just means one more thing that maintainers of this project have to think about...

tsingsun:
I also think that the library should be kept clean, but the GETER method is very common, isn't it?

pavelnikolov:
Actually the `Get` prefix is considered an anti-pattern in Golang and is not Go-idiomatic. Please, read this paragraph in Effective Go: https://golang.org/doc/effective_go.html?#Getters

tsingsun:
if user user [prototobuf](https://github.com/golang/protobuf),generate code user Getter and Setter.

tsingsun:
if user user golang [prototobuf](https://github.com/golang/protobuf) ,office code-gen use Getter and Setter.I will use in graphql resovler like 
```
type greet struct{
   *protobuf.Greet 
}
```
the protobuf Greet code use protoc-gen-go will be like
```
type Greet struck{
     Hello string
}
func (m * Greet) GetHello() string {
	if m != nil {
		return m.Msg
	}
	return ""
}
```
so i suggest adding the support for getter....
by the way my branch has added a bit of functionality after submission, it is my fault, should not use the master branch to initiate PR.  please refuse to drop this PR

pavelnikolov:
Look, I understand that protobuf/gRPC is a common choice when designing/developing microservices. My team also uses protobuf for the vast majority of our services. However, I believe that this library should not tolerate any prefixes (including `Get`). I am not one of the maintainers of this project, but if I was, I would vote against this "feature". Adding exceptions like this is not good for any library. Instead of adding exceptions for special methods, I would allow fields to be resolvers. Which is exactly what PR #194 does.

PR #221, opened 2018-06-02
provide build-in graphqli

andrzejsliwa:
Goal: Provide build-in Graphqli

Example:
```golang
func main() {
	s, err := loadSchema("./schema.graphql")
	if err != nil {
		log.Fatal(err)
	}

	schema := graphql.MustParseSchema(s, &Resolver{})

	r := mux.NewRouter()
	r.Handle("/graphql", &relay.Handler{Schema: schema, Graphqli: true})
	log.Fatal(http.ListenAndServe(":8080", r))
}
```

### DIFF
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -45,9 +45,13 @@ func UnmarshalSpec(id graphql.ID, v interface{}) error {
 
 type Handler struct {
 	Schema *graphql.Schema
+	Graphqli bool
 }
 
 func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if h.wasGraphqliHandled(w, r) {
+		return
+	}
 	var params struct {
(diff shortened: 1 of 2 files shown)


andrzejsliwa:
@maxbeatty: can you give more feedback than thumbs down? Any specific cons?

pavelnikolov:
I believe that this should not be part of the library. The library should be minimalistic and have very few features out of the box (GraphQL specification coverage + OpenTracing).
Each team might use different version of Graph_i_QL. Even if two teams use the same version, they might use different plugins for it (like custom implementations for storing queries etc). It's nice to have this as an example, but it should not be included out-of-the-box.

andrzejsliwa:
@pavelnikolov thx for feedback and good arguments against of merging it. I’m closing it.

PR #225, opened 2018-06-11
Change META schema file to a .graphql file

lamebear:
Moving the schema into its own file allows for syntax highlighting and a cleaner reading experience.

A `Makefile` is also included to simplify the execution of `go generate` to transpile the `.graphql` file into go code.


pavelnikolov:
Thank you but I'm not going to do this for now as it doesn't solve any problem. In addition to that Go allows for file embedding and that would be option that I would consider in future.

PR #228, opened 2018-06-18
Changing error comparision in testing

nikkon226:
When running tests, the errors are not equal.
When using []*errors.QueryError{errors.Errorf("%s",msg)}, all things being equal
the errors are returned as unequal. Comparing just the messages will be adequate
for application development environments.

The errors that are returned by the internal package cannot be duplicated (easily) when testing. This will allow the messages to be compared as it is the only required component that will be returned in the QueryError struct.

### DIFF
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -4,8 +4,8 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


pavelnikolov:
You have to use proper mock for predictable results.

PR #229, opened 2018-06-20
Implement error propagation for null vs non-null fields

### LINKED ISSUE #173: Spec compliance: Errors do not propagate to the closest nullable field
### How to reproduce:
- Define a schema with a nullable field, and a non-nullable sub-field
- Have the schema for the non-nullable field return an error

### Expected:

The nullable field should be `null` and the error should point to that field. Specifically, the non-nullable field should not appear in the response because it must never be `null` - the nullable parent field must be `null` instead.

From the spec:

> Since Non-Null type fields cannot be null, field errors are propagated to be handled by the parent field. If the parent field may be null then it resolves to null, otherwise if it is a Non-Null type, the field error is further propagated to it’s parent field.

http://facebook.github.io/graphql/draft/#sec-Errors-and-Non-Nullability

### Actual:

Both the nullable and the non-nullable field will be present, with the non-nullable field being `null` and the error in the `errors` array pointing to that error. This is not spec-compliant.

-------------

Version: b46637030579abd312c5eea21d36845b6e9e7ca4 (latest)

gracenoah:
fix #173

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -102,6 +102,15 @@ func (r *findDroidResolver) FindDroid(ctx context.Context) (string, error) {
 	}
 }
 
+type findDroidsResolver struct{}
+
+func (r *findDroidsResolver) FindDroids(ctx context.Context) []*droidResolver {
+	return []*droidResolver{&droidResolver{}, &droidResolver{resolverNotFoundError{
+		Code:    "NotFound",
+		Message: "This is not the droid you are looking for",
+	}}}
+}
+
 type findDroidOrHumanResolver struct{}
 
 func (r *findDroidOrHumanResolver) FindHuman(ctx context.Context) (*string, error) {
@@ -116,10 +125,13 @@ func (r *findDroidOrHumanResolver) FindDroid(ctx context.Context) (*droidResolve
 	}
 }
 
-type droidResolver struct{}
+type droidResolver struct{ err error }
 
-func (d *droidResolver) Name() string {
-	return "R2D2"
+func (d *droidResolver) Name() (string, error) {
+	if d.err != nil {
+		return "", d.err
+	}
+	return "R2D2", nil
 }
 
 type discussPlanResolver struct{}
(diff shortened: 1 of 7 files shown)


gracenoah:
Can someone please review and/or merge this PR?

gracenoah:
bump?

gracenoah:
Rebased this PR, but it's failing tests now because it doesn't handle subscriptions correctly.

gracenoah:
Updated to handle errors in subscriptions. I'm not sure what the expected behaviour there was, so please take a look at the examples in the tests.

gracenoah:
After doing some local testing, I think this PR needs a bit more work to handle nulls in subscriptions correctly. I'm working through the various test cases right now and trying to guess what the expected behaviour us, but any tips on that front would be appreciated.

gracenoah:
Updated. I think the tests cover the various edge cases that needed to be handled.

graphql_test.go
```diff
@@ -370,9 +370,7 @@ func TestErrorWithExtensions(t *testing.T) {
 				}
 			`,
 			ExpectedResult: `
-				{
-					"FindDroid": null
-				}
+				null
```
tonyghita: It would be good to have a test which ensures the behavior is as described in the specification: https://facebook.github.io/graphql/draft/#sec-Errors-and-Non-Nullability

Specifically, we should test that children of non-null nodes are not returned, and that the siblings of the non-null node are also not returned (because the error propagates all the way up to the nearest nullable parent.

gracenoah:
Updated with a more thorough test.

graphql_test.go
bsr203: on unrelated note. Great to see you back Tony :-)

tinnywang:
Hi @gracenoah, I have an [error propagation branch](https://github.com/tinnywang/graphql-go/compare/master...tinnywang:error-propagation) based off yours, with some additions/changes for handling nullability in lists and non-null fields resolving to null. Should I open a new PR here, or would you like me to open one in your fork so my changes are merged into your branch?

gracenoah:
Please open a separate pull request against this repo. Feel free to based it off my branch. I'm unlikely to push hard for this to be merged because we are switching to https://github.com/99designs/gqlgen. It supports the spec a lot better.

pavelnikolov:
Closing this PR as #296 is based on this PR + other changes.

pavelnikolov:
@gracenoah I'd be very curious to know if you have faced any other problems with this library. Especially, cases where this library doesn't support the GraphQL spec... 
Any feedback would be highly appreciated. Here or in the Gophers Slack...

PR #234, opened 2018-06-27
Pass back JSON marshaling error

matthew-noken:
It can be difficult to debug these errors without this information.


matthew-noken:
Hi, can this be merged?

PR #237, opened 2018-07-13
Properly handle nil interface resolvers

### LINKED ISSUE #233: Interface can't be used as nullable resolver
Hello,

I'm currently trying to use a go-interface as nullable resolver.
The structure of my code is like this:
```graphql
type Foo {
   getBar(): Bar
}

type Bar {
   doBarThings(): Bool!
}
```

```go
type Foo struct {}

func (f *Foo) GetBar() Bar {
	if someCondition {
		return &SpecificBar{}
	} else {
		return nil
	}
}

type Bar interface {
   DoBarThings() bool
}

type SpecificBar struct {}

func (sb *SpecificBar) DoBarThings() bool {
   // ...
   return true
}
```
This whole thing works without any problems, as long as `Bar` is not `nil`,
in which case I get `graphql: panic occurred: reflect: Method on nil interface value`.

Replacing 
```go
GetBar() Bar
```
with
```go
GetBar() *SpecificBar
```
Would fix the problem, however, I do want to have the option to have specific implementations for the `Bar` type.

My guess would be that this is a (somehow) desired behaviour, since nil interface values are a thing.
However, I never casted my `nil` to `Bar`, which (to my knowledge) would be required for nil to be recognized as interface value.

I have also tried to pass the interface as a pointer, in the hopes that it would trigger the library to expect a nillable value (like it is required for nullable arrays):
```go
GetBar() *Bar
```
Which just leads to `*Bar does not resolve "Bar": missing method for field "doBarThings"`.

tsholmes:
fixes #233 

Calling `reflect.ValueOf(nil)` returns the zero `Value`, which has type `Invalid`.
Calling `reflect.ValueOf` an interface typed nil returns a `Value` with type `Interface`.

Both of these cases skipped the null checks and fail later on when trying to resolve fields.

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -210,7 +210,7 @@ func (r *Request) execSelectionSet(ctx context.Context, sels []selected.Selectio
 	switch t := t.(type) {
 	case *schema.Object, *schema.Interface, *schema.Union:
 		// a reflect.Value of a nil interface will show up as an Invalid value
-		if resolver.Kind() == reflect.Invalid || (resolver.Kind() == reflect.Ptr && resolver.IsNil()) {
+		if resolver.Kind() == reflect.Invalid || ((resolver.Kind() == reflect.Ptr || resolver.Kind() == reflect.Interface) && resolver.IsNil()) {
 			if nonNull {
 				panic(errors.Errorf("got nil for non-null %q", t))
 			}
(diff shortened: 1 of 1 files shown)


tsholmes:
Couldn't change the base branch of this, so re-opened this as #239

PR #251, opened 2018-08-23
Time as an input with tests

### LINKED ISSUE #248: Unable to use scalar Time for Inputs
It looks like when you try to use scalar time as an input, you get an error message "cant unmarshal <string> (string) as time.Time 

screen shot 2018-08-16 at 9 54 03 pm

Melaninneal:
Please consider adding the ability to have type "time" for inputs fixes #248


internal/exec/packer/packer.go
```diff
@@ -325,7 +325,10 @@ func unmarshalInput(typ reflect.Type, input interface{}) (interface{}, error) {
 	if reflect.TypeOf(input) == typ {
 		return input, nil
 	}
-
+	if typ.String() == "time.Time" {
```
vetcher: may be we can make it in more flexible way?
what if I want to import package `time` like? 
```go
import . "time"
```
or 
```go
import stdtime "time"
```

pavelnikolov: No, I wouldn't add alias to standard library packages and I would definitely avoid `.` imports. https://github.com/golang/go/wiki/CodeReviewComments#import-dot

internal/exec/packer/packer.go
```diff
@@ -359,6 +362,10 @@ func unmarshalInput(typ reflect.Type, input interface{}) (interface{}, error) {
 	return nil, fmt.Errorf("incompatible type")
 }
 
+func convertStringToTime(timeString string) (interface{}, error){
```
pavelnikolov: Do you need to return `interface{}` here? Can't your return time?

internal/exec/packer/packer.go
pavelnikolov: Also:
> Avoid renaming imports except to avoid a name collision;...

https://github.com/golang/go/wiki/CodeReviewComments#imports

internal/exec/packer/packer.go
```diff
@@ -325,7 +325,10 @@ func unmarshalInput(typ reflect.Type, input interface{}) (interface{}, error) {
 	if reflect.TypeOf(input) == typ {
 		return input, nil
 	}
-
+	if typ.String() == "time.Time" {
+		return convertStringToTime(input.(string))
```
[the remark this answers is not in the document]
Melaninneal: i realized this doesn't quite work with Nullable time objects will add second condition and remove convertStringToTime func

BarthesSimpson:
@pavelnikolov - any chance this can be merged now that the convertStringToTime function has been removed?

internal/exec/packer/packer.go
```diff
@@ -5,10 +5,10 @@ import (
 	"math"
 	"reflect"
 	"strings"
-
-	"github.com/graph-gophers/graphql-go/errors"
+		"github.com/graph-gophers/graphql-go/errors"
 	"github.com/graph-gophers/graphql-go/internal/common"
 	"github.com/graph-gophers/graphql-go/internal/schema"
+	"time"
```
pavelnikolov: Please, fix the imports order

internal/exec/packer/packer.go
```diff
@@ -359,6 +368,8 @@ func unmarshalInput(typ reflect.Type, input interface{}) (interface{}, error) {
 	return nil, fmt.Errorf("incompatible type")
 }
 
+
+
```
pavelnikolov: Unrelated change. Please, remove these empty lines.

graphql_test.go
```diff
@@ -1691,6 +1706,8 @@ func TestInput(t *testing.T) {
 					recursive(value: {next: {next: {}}})
 					intID: id(value: 1234)
 					strID: id(value: "1234")
+					timeInput: inputTime(value: {t : "2018-08-22T11:34:55Z"})
+					nullableTimeInput: nullableInputTime(value: {t : "2018-08-22T11:34:55Z"})
```
pavelnikolov: Please, add unit tests with a `null` value.

internal/exec/packer/packer.go
```diff
@@ -325,7 +325,16 @@ func unmarshalInput(typ reflect.Type, input interface{}) (interface{}, error) {
...
+	switch typ.String(){
+	case  "time.Time": {
+		return time.Parse(time.RFC3339, input.(string))
+	}
+	case "*time.Time":
+		t, err := time.Parse(time.RFC3339, input.(string))
+		return &t, err
+	}
+	
+
```
pavelnikolov: you don't really need double empty lines, no?

pavelnikolov:
Please, add missing unit tests unit tests.

pavelnikolov:
Can you just use `graphql.Time` as the type of your argument? You'll need to define `scalar Time` in your schema in order to be able to use it though.

pavelnikolov:
The underlying issue have been fixed already with a different PR.

Issue #231, opened 2018-06-26
Cannot use library with go1.7


ss-limafriyadi:
Hello,

When I try to use this with go1.7 it returns the following error

```
# vendor/github.com/graph-gophers/graphql-go/internal/validation
vendor/github.com/graph-gophers/graphql-go/internal/validation/suggestion.go:25: undefined: sort.Slice
# vendor/github.com/graph-gophers/graphql-go/trace
vendor/github.com/graph-gophers/graphql-go/trace/validation_trace.go:7: syntax error: unexpected = in type declaration
```

I found that `sort.Slice` first released in go1.8 and `type =` syntax is introduced in go1.9.

Can we make this compatible with go1.7?

pavelnikolov:
I guess that we definitely need to document the minimal required Go version! Thanks for reporting this.

ss-limafriyadi:
Any plan to support go1.7? Or must we use minimum working version of go1.9?

pavelnikolov:
I don't think we need to support old versions. However from a single glance this type aliasing doesn't seem to be necessary at all. We could just use the existing func, no? I have to dig deeper into this code.

PR #261, opened 2018-09-16
Update validation_trace.go

continuum-ajay-sahani:
Compilation error fixed

type TraceValidationFinishFunc = TraceQueryFinishFunc

change to:
type TraceValidationFinishFunc TraceQueryFinishFunc

### DIFF
--- a/trace/validation_trace.go
+++ b/trace/validation_trace.go
@@ -4,7 +4,7 @@ import (
 	"github.com/graph-gophers/graphql-go/errors"
 )
 
-type TraceValidationFinishFunc = TraceQueryFinishFunc
+type TraceValidationFinishFunc TraceQueryFinishFunc
 
 type ValidationTracer interface {
 	TraceValidation() TraceValidationFinishFunc


vetcher:
what's golang version you use?

pavelnikolov:
What problem does this solve?

pavelnikolov:
Is this related to #231 ?

vetcher:
@pavelnikolov definitely

pavelnikolov:
What version of Go are you using?

continuum-ajay-sahani:
Go version 1.9, I was getting the compilation error.

trace/validation_trace.go
```diff
@@ -4,7 +4,7 @@ import (
 	"github.com/graph-gophers/graphql-go/errors"
 )
 
-type TraceValidationFinishFunc = TraceQueryFinishFunc
+type TraceValidationFinishFunc TraceQueryFinishFunc
```
pavelnikolov: TBH, `TraceValidationFinishFunc` is only ever used in this file. IMHO, we don't even need type aliasing here. I'd be happy to merge your PR but I think that the approach is wrong. Instead, we need to remove this line altogether and replace `TraceValidationFinishFunc` on lines 10 and 15 with `TraceQueryFinishFunc`.

pavelnikolov: WDYT?

pavelnikolov:
I'm closing this PR due to inactivity and I believe that it is no longer relevant after the tracing refactoring.

PR #264, opened 2018-09-21
reverse order of errors vs data in response

gracenoah:
In this PR I moved the errors field before the data field in the response object based on the recommendation of the graphql standard merged in this pr: https://github.com/facebook/graphql/pull/384
>Note: When `errors` is present in the response, it may be helpful for it to 
> appear first when serialized to make it more clear when errors are present
> in a response during debugging.

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -110,6 +110,7 @@ func Logger(logger log.Logger) SchemaOpt {
 
 // Response represents a typical response of a GraphQL server. It may be encoded to JSON directly or
 // it may be further processed to a custom response type, for example to include custom error data.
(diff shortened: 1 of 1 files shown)


pavelnikolov:
Thanks for the contribution!

graphql.go
```diff
@@ -111,8 +111,8 @@ func Logger(logger log.Logger) SchemaOpt {
 // Response represents a typical response of a GraphQL server. It may be encoded to JSON directly or
 // it may be further processed to a custom response type, for example to include custom error data.
```
pavelnikolov: Can you, please, add this link to the comment: https://github.com/facebook/graphql/pull/384/commits/7b40390d48680b15cb93e02d46ac5eb249689876

gracenoah: Done.

PR #265, opened 2018-09-27
panic message should me meaningful: invalid type should say which typ…

kirillDanshin:
…e has an error

Signed-off-by: Kirill Danshin <k@guava.by>

### DIFF
--- a/internal/exec/resolvable/resolvable.go
+++ b/internal/exec/resolvable/resolvable.go
@@ -173,7 +173,7 @@ func (b *execBuilder) makeExec(t common.Type, resolverType reflect.Type) (Resolv
 		return e, nil
 
 	default:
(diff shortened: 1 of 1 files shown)


internal/exec/resolvable/resolvable.go
```diff
@@ -173,7 +173,7 @@ func (b *execBuilder) makeExec(t common.Type, resolverType reflect.Type) (Resolv
 		return e, nil
 
 	default:
-		panic("invalid type")
+		panic("invalid type: " + t.String())
```
pavelnikolov: I'm interested to see cases which cause this panic.

kirillDanshin: just make a typo: define input type as `type`, not `input` :)

PR #266, opened 2018-09-27
Introduce resolver prefix

kirillDanshin:


pavelnikolov:
Why should we add this prefix to the library? What problem does it solve? Why can't you just rename the method of your struct? If the resolver struct comes from a library that you have no control of, you can wrap it and expose the proper method which invokes the other one. I do see the need for adding this to the library. Although this was requested one more time earlier this year, I don't see any value in this.
Can you show an example where this is really needed and solves an actual problem? I'm especially interested a problem that cannot be resolved in any better way?

kirillDanshin:
@pavelnikolov we have had an issue and the best way to solve it was this changes. For more, take a look below.

---

let's assume that we have
```
type A struct {
    FieldA string
    FieldB int
    // etc.
}
```

to resolve it, previously we had options like `FieldA_()`, `FIELDA()`, `F_I_E_L_D_A()` and other pretty bad ideas.

now we have ability to just add prefix. let's assume, we choose `GQL` as prefix: `GQLFieldA()` etc., or if we choose `Resolve`: `ResolveFieldA()`. that's more convinient way to work with GQL.

pavelnikolov:
Introducing this feature would potentially add unnecessary complexity and on the other hand it doesn't bring much value to the table.
I am closing this PR as there are other ways to address your problems.

kirillDanshin:
@pavelnikolov memory overhead is too much for us. although my example includes only scalar fields, but we have **much** more examples with more complicated types like `struct { big.Int, big.Int }` etc., I just don't want to make over-complicated examples. unfortunately, without this change we will be forced to keep our fork alive, because resolving scalar types directly doesn't solve our case at all, and bringing in `_` workarounds or using ALLCAPS is hard to maintain in production code.

pavelnikolov:
Does wrapping the struct in another struct hurt your memory performance significantly? I'd be very interested to see benchmarks.

pavelnikolov:
@kirillDanshin I am sorry for not merging this straight away. I truly appreciate contributions and didn't mean to be rude. However, I would like to keep this library as minimalistic as possible and only add things if they bring real value and are absolutely necessary - this just doesn't seem to be one of them.

pavelnikolov:
@kirillDanshin There is another option without adding this to the library! @neelance suggested that the `findMethod` might be configurable and run with the current implementation by default but allow users of the library to override it. This way you'll be able to implement prefixes/suffixes or whatever other features you might be interested in...
WDYT?

PR #267, opened 2018-09-27
improve gqltesting

alicebob:
Hi,

gqltesting is very helpful for testing graphql endpoints, but there are two small problems which this PR fixes:

- If you send an invalid graphql query (for example you ask for a field which doesn't exists) you get:
```
    testing.go:49: got: invalid JSON: unexpected end of JSON input ("")
```
That's because the `Result` field is empty. But there is an error object, so the first commit first checks the errors, and if they are as expected then it'll do the body of the result.

- the error check gave a rather uninformative error:
```
    testing.go:82: unexpected number of errors: got 1, want 0
```

The second commit makes that:
```
    testing.go:80: unexpected error: got [graphql: Cannot query field "timestamp" on type "SMSList". (line 1, column 40)], want []
```
and as a bonus the code is now trivial.

### DIFF
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -43,6 +43,9 @@ func RunTest(t *testing.T, test *Test) {
 		test.Context = context.Background()
 	}
 	result := test.Schema.Exec(test.Context, test.Query, test.OperationName, test.Variables)
+
+	checkErrors(t, test.ExpectedErrors, result.Errors)
+
 	// Verify JSON to avoid red herring errors.
 	got, err := formatJSON(result.Data)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pavelnikolov:
Hi @alicebob 
thank you very much for your PR!
I have a small request for you, can you please split it into two and send two PRs, please? I have comments for both changes but ideally I want to keep unrelated changes separate. Also, you might consider adding some unit tests and asses that you get correct message when you move the error checking first?

alicebob:
Thank you for your reply @pavelnikolov . I've (finally!) split them up. The new two PRs are linked above.

Thanks!

PR #269, opened 2018-10-03
Parse string based descriptions

### LINKED ISSUE #147: Support latest syntax for descriptions
Descriptions in schemas are now represented as strings instead of comments as of [graphql-js 0.12.0](https://github.com/graphql/graphql-js/releases/tag/v0.12.0)
https://github.com/graphql/graphql-js/pull/927

### LINKED ISSUE #196: Support for string literals as description
As of version 0.12.3 the default style of describing a type or field is to add a string literal instead of a comment.
https://github.com/graphql/graphql-js/blob/master/src/utilities/extendSchema.js#L44

Using the schema
```graphql
"This is a description."
schema {
    query: Query
}
```
results in a crash.

gracenoah:
This implementation is probably lax about where string descriptions are allowed. Additional test cases would be needed to verify the accuracy of the implementation. Also, this is backwards compatible with comment based descriptions and there's an option to disable the backwards compatibility.

fix #196 
fix #147

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -34,7 +34,7 @@ func ParseSchema(schemaString string, resolver interface{}, opts ...SchemaOpt) (
 		opt(s)
 	}
 
-	if err := s.schema.Parse(schemaString); err != nil {
+	if err := s.schema.Parse(schemaString, s.noCommentsAsDescriptions); err != nil {
 		return nil, err
 	}
 
@@ -63,15 +63,23 @@ type Schema struct {
 	schema *schema.Schema
 	res    *resolvable.Schema
 
-	maxDepth         int
-	maxParallelism   int
-	tracer           trace.Tracer
-	validationTracer trace.ValidationTracer
-	logger           log.Logger
+	maxDepth                 int
+	maxParallelism           int
+	tracer                   trace.Tracer
+	validationTracer         trace.ValidationTracer
+	logger                   log.Logger
+	noCommentsAsDescriptions bool
 }
 
 // SchemaOpt is an option to pass to ParseSchema or MustParseSchema.
 type SchemaOpt func(*Schema)
+
+// NoCommentsAsDescriptions disables the parsing of comments as descriptions
+func NoCommentsAsDescriptions() SchemaOpt {
+	return func(s *Schema) {
+		s.noCommentsAsDescriptions = true
+	}
+}
 
 // MaxDepth specifies the maximum field nesting depth in a query. The default is 0 which disables max depth checking.
 func MaxDepth(n int) SchemaOpt {
--- a/graphql.go
+++ b/graphql.go
@@ -34,7 +34,7 @@ func ParseSchema(schemaString string, resolver interface{}, opts ...SchemaOpt) (
 		opt(s)
 	}
 
-	if err := s.schema.Parse(schemaString, s.noCommentsAsDescriptions); err != nil {
+	if err := s.schema.Parse(schemaString, s.useStringDescriptions); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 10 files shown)


pavelnikolov:
Thank you for your PR! I'll review it shortly!

gracenoah:
Thinking about this PR again this morning, I realize it's a real quick and dirty fix that works for my current use case, but in general is way too permissive. I'm thinking of improving it by splitting up Consume() from ConsumeDescription() instead of having a flag passed into the same function that consumes either one. That way I can significantly reduce the possible invalid schemas accepted. I think I was trying too hard to keep my changes only within the Consume() function and that led to a bad design. Feel free to review other PRs first. I'll let you know when I think this is in a mergable state.

gracenoah:
Updated. I think this is a bit more reasonable. I'm still not happy with the level of testing, but it did work with the new flag on in our codebase.

graphql.go
```diff
@@ -34,7 +34,7 @@ func ParseSchema(schemaString string, resolver interface{}, opts ...SchemaOpt) (
 		opt(s)
 	}
 
-	if err := s.schema.Parse(schemaString); err != nil {
+	if err := s.schema.Parse(schemaString, s.noCommentsAsDescriptions); err != nil {
```
pavelnikolov: Should we name the variable `useStringComments` instead? I think that shorter is better. WDYT?

Arachnid:
Is there anything preventing this from being merged? It'd be great to be able to load schemas with documentation in them.

pavelnikolov:
@Arachnid I'm sorry for the delay. I'm currently looking into subscriptions and then I'll get back to the remaining PRs - I think that tis name (i.e. `noCommentsAsDescriptions`) is too long. Otherwise, LGTM.

graphql.go
gracenoah: I'll rename it to useStringsAsDescriptions. The spec uses the term "description" so I prefer to use that here too.

gracenoah: Actually, `useStringDescriptions` is as short as I can make it.

gracenoah:
Ready to merge if you are happy with it.

internal/common/lexer.go
```diff
@@ -132,19 +166,57 @@ func (l *Lexer) Location() errors.Location {
...
+func (l *Lexer) consumeTripleQuoteComment() {
+	if l.next != '"' {
+		panic("consumeTripleQuoteComment used in wrong context: no third quote?")
+	}
+
+	if l.descComment != "" {
+		l.descComment += "\n"
+	}
+
+	comment := ""
```
pavelnikolov: I would prefer using the default value here instead of using an assignment:
```go
var comment string
```

internal/common/lexer.go
```diff
@@ -132,19 +166,57 @@ func (l *Lexer) Location() errors.Location {
...
+	numQuotes := 0
```
pavelnikolov: default val again:
```go
var numQuotes int
```

internal/common/lexer.go
```diff
@@ -84,6 +88,31 @@ func (l *Lexer) Consume() {
...
+		if l.sc.Peek() == '"' {
+			// Consume the third quote
+			l.next = l.sc.Next()
+			l.consumeTripleQuoteComment()
+		} else {
+			l.consumeStringComment(tokenText)
+		}
+		return true
+	}
+	return false
```
pavelnikolov: This line is not covered by unit tests...

internal/common/lexer.go
```diff
@@ -132,19 +166,57 @@ func (l *Lexer) Location() errors.Location {
 	}
 }
 
+func (l *Lexer) consumeTripleQuoteComment() {
+	if l.next != '"' {
+		panic("consumeTripleQuoteComment used in wrong context: no third quote?")
+	}
+
+	if l.descComment != "" {
+		l.descComment += "\n"
```
pavelnikolov: Not covered by unit tests. Can you add a case for this, please?

internal/common/lexer.go
```diff
@@ -132,19 +166,57 @@ func (l *Lexer) Location() errors.Location {
...
+		if numQuotes == 3 || l.next == scanner.EOF {
+			break
+		}
+	}
+	l.descComment += strings.TrimSpace(comment[:len(comment)-numQuotes])
+}
+
+func (l *Lexer) consumeStringComment(str string) {
+	if l.descComment != "" {
+		l.descComment += "\n"
```
pavelnikolov: Not covered by unit tests. Can you add a test case for this, please?

graphql.go
```diff
@@ -63,16 +63,24 @@ type Schema struct {
...
+	tracer                trace.Tracer
+	validationTracer      trace.ValidationTracer
+	logger                log.Logger
+	useStringDescriptions bool
 }
 
 // SchemaOpt is an option to pass to ParseSchema or MustParseSchema.
 type SchemaOpt func(*Schema)
 
+// UseStringDescriptions disables the parsing of comments as descriptions
```
pavelnikolov: Can you make this comment a little bit more descriptive, please? This comment will be used by `godoc` to generate documentation for the library and some users might not even know about strings vs comments as descriptions.

gracenoah:
Sorry I haven't been responsive. I think I'll be able to get back to this next week.

internal/common/lexer.go
gracenoah: Actually this case doesn't make sense. A one line string comment will always be the only description.

gracenoah:
updated

PR #270, opened 2018-10-04
Adding new Engine API that allows users to implement custom resolvers.

chirino:
Yeah i know it's kinda of a big change.  But it does pack lots of new features.  If you guys think it's merge worthy I'll take the time rebase on the latest master to pick up recent changes.


chirino:
This PR should be keeping API compatibility with the previous API, but adds additional apis which are documented in the [readme](https://github.com/graph-gophers/graphql-go/blob/376c493e3320924bf2b122e543848aabf4ed9eb2/README.md#resolvers).

bsr203:
wish there were some kind of benchmark test to see any performance implications for significant changes.

chirino:
@bsr203 agree, do you guys have any handy?

bsr203:
@chirino sorry. I don't. thanks for your effort on improving the library for all of us :-)

chirino:
Ok I just added a little benchmark.  These are the results on my mac book pro (go version go1.11 darwin/amd64):

This PR run at: 434279 ns/op
```
$ go test -bench=.

BenchmarkStarwarsQuery-8   	    3000	    434279 ns/op
PASS
ok  	github.com/graph-gophers/graphql-go	6.399s
```

This same benchmark against master ( https://github.com/chirino/graphql-go/tree/benchmark ) runs at: 361503 ns/op

```
$ go test -bench=.
goos: darwin
goarch: amd64
pkg: github.com/graph-gophers/graphql-go
BenchmarkStarwarsQuery-8   	    5000	    361503 ns/op
PASS
ok  	github.com/graph-gophers/graphql-go	1.867s
```

So basically this PR introduces about a 16% perf dop :(
Let me see if I can figure out how to improve that a bit.

pavelnikolov:
Thank you very much for your contribution and ideas! I appreciate the effort that you have put into this 👏 
However, this PR contains way too many changes and IMHO needs to be split into multiple PRs (only one logical change per PR would be perfect).
Ideally, if we add some of this functionality, it needs to be 100% opt-in and maybe configurable/customizable. To expand a little bit more on this - we might leave the existing behavior as the default but allow users of the library to override the resolve function in their projects. This way there will be unlimited possibilities... We can add your code as reference examples but leave the implementation to the users' imagination. WDYT?

chirino:
I'm not sure how to break this out into to smaller PRs.  Feel free to provide specific suggestions.
The existing behaviour is still the default.  It just has additional fallback resolution strategies.
Notice that the existing unit tests were not really changed.

pavelnikolov:
A quote from you:
> Yeah i know it's kinda of a big change. But it does pack lots of new features...

How about sending a separate PR for each new feature?

chirino:
This PR is getting stale, and I don't have the energy to break it up, I'll just maintain my own fork for my needs.

PR #271, opened 2018-10-08
Add community example

hsukvn:
I wrote a simple example of grapqlQL API server by this library integrated with gin http framework. Let me know what you guys think about sharing this example. Thanks.


hsukvn:
Hi @pavelnikolov 

This repo demonstrate how to build an graphQL API server to get Linux system status and manipulate systemd service

currently is able to query
- CPU
- memory
- network interface
- systemd service

and mutate
- enable service
- disable service
- start service
- stop service

adriancuadros:
This has some sweet docs on using actually using the api. Perhaps it would be a good idea to open up a samples page on the wiki and start adding additional community samples to it.

pavelnikolov:
It is a nice idea. I've been thinking about this for a while. I'll probably create a separate Wiki page as you suggest. This way I'll add a single link in the `README.md` that will point to the wiki page. And the wiki page will only contain links to examples/resources.

pavelnikolov:
I added it to the wiki https://github.com/graph-gophers/graphql-go/wiki/Examples

PR #272, opened 2018-10-10
Add extensions in QueryError

ivanp:
Added "extensions" field in QueryError to conform with [Apollo Server errors](https://www.apollographql.com/docs/apollo-server/features/errors.html).

### DIFF
--- a/errors/errors.go
+++ b/errors/errors.go
@@ -5,11 +5,12 @@ import (
 )
 
 type QueryError struct {
-	Message       string        `json:"message"`
-	Locations     []Location    `json:"locations,omitempty"`
-	Path          []interface{} `json:"path,omitempty"`
-	Rule          string        `json:"-"`
-	ResolverError error         `json:"-"`
+	Message       string                 `json:"message"`
+	Locations     []Location             `json:"locations,omitempty"`
+	Path          []interface{}          `json:"path,omitempty"`
+	Rule          string                 `json:"-"`
+	ResolverError error                  `json:"-"`
+	Extensions    map[string]interface{} `json:"extensions,omitempty"`
 }
 
 type Location struct {
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -66,22 +66,40 @@ func (r *timeResolver) AddHour(args struct{ Time graphql.Time }) graphql.Time {
 
 var starwarsSchema = graphql.MustParseSchema(starwars.Schema, &starwars.Resolver{})
 
-type findDroidResolver struct{}
+type ResolverError interface {
+	error
+	Extensions() map[string]interface{}
+}
 
-type withExtensionError struct {
+type resolverNotFoundError struct {
 	Code    string `json:"code"`
 	Message string `json:"message"`
 }
 
-func (e withExtensionError) Error() string {
-	return fmt.Sprintf("Error [%s] %s", e.Code, e.Message)
+func (e resolverNotFoundError) Error() string {
+	return fmt.Sprintf("Error [%s]: %s", e.Code, e.Message)
 }
 
+func (e resolverNotFoundError) Extensions() map[string]interface{} {
+	return map[string]interface{}{
+		"code":    e.Code,
+		"message": e.Message,
+	}
+}
+
+type findDroidResolver struct{}
+
 func (r *findDroidResolver) FindDroid(ctx context.Context) (string, error) {
-	return "", withExtensionError{
+	return "", resolverNotFoundError{
 		Code:    "NotFound",
 		Message: "This is not the droid you are looking for",
 	}
+}
+
+type discussPlanResolver struct{}
+
+func (r *discussPlanResolver) DismissVader(ctx context.Context) (string, error) {
+	return "", errors.New("I find your lack of faith disturbing")
 }
 
 func TestHelloWorld(t *testing.T) {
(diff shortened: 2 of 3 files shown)


pavelnikolov:
What problem does this solve? I'm very interested in adding extensions to this library but I don't see the value of this commit 🤔

ivanp:
I am trying to write a custom error response according to the [GraphQL error spec](https://facebook.github.io/graphql/June2018/#sec-Errors), so i can write something like this:

```{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ],
      "extensions": {
        "code": "CAN_NOT_FETCH_BY_ID",
        "timestamp": "Fri Feb 9 14:33:09 UTC 2018"
      }
    }
  ]
}```

**Comment by smithaitufe** [comment]
According to the spec, the extensions property must be a map or a hash. It says that when it is serialized to json, it should be an object.

**Comment by pavelnikolov** [comment]
Can you, please, add unit tests?

**Comment by ivanp** [comment]
I was thinking, what if we make the ResolverError field outputs to json?

```
type QueryError struct {
	Message       string                 `json:"message"`
	Locations     []Location             `json:"locations,omitempty"`
	Path          []interface{}          `json:"path,omitempty"`
	Rule          string                 `json:"-"`
	ResolverError error                  `json:"extensions,omitempty"`
}
```

That way if the resolver returns a custom error with tags it will be displayed in extensions field.

File: errors/errors.go [hunk 1]
```diff
@@ -9,7 +9,7 @@ type QueryError struct {
 	Locations     []Location    `json:"locations,omitempty"`
 	Path          []interface{} `json:"path,omitempty"`
 	Rule          string        `json:"-"`
-	ResolverError error         `json:"-"`
+	ResolverError error         `json:"extensions,omitempty"`
```
pavelnikolov: Shouldn't we name this field `Extensions` with type `map[string]interface{}`. This way it can still contain an error as well as other fields. For example, it might be a validation error and you might include the error(s) as well as the invalid input field(s) or you might want to include request ID, trace ID or other fields that are not directly related to the error 🤔

smithaitufe: I agree with @pavelnikolov  that the name should be `Extensions` and the type should be a `map[string]interface{}`

ivanp:
OK I have reverted to the previous commit to add `Extensions` as `map[string]interface{}`.

errors/errors.go
```diff
@@ -5,11 +5,12 @@ import (
...
-	Locations     []Location    `json:"locations,omitempty"`
-	Path          []interface{} `json:"path,omitempty"`
-	Rule          string        `json:"-"`
-	ResolverError error         `json:"-"`
+	Message       string                 `json:"message"`
+	Locations     []Location             `json:"locations,omitempty"`
+	Path          []interface{}          `json:"path,omitempty"`
+	Rule          string                 `json:"-"`
+	ResolverError error                  `json:"-"`
+	Extensions    map[string]interface{} `json:"extensions,omitempty"`
```
pavelnikolov: I'm wondering if the `Extensions` field should be used in the `Error()` method as well... 🤔

graphql_test.go
```diff
@@ -304,6 +323,45 @@ func TestNilInterface(t *testing.T) {
 	})
 }
 
+func TestErrorWithExtension(t *testing.T) {
+	err := withExtensionError{
+		Code:    "NotFound",
+		Message: "This is not the droid you are looking for",
```
pavelnikolov: The test is named `TestErrorWithExtension`, however the error doesn't have an extension field set. Was this indetntional? Can you, please, add some more test cases? Then I'll be happy to merge this PR.

pavelnikolov: Maybe you need to add cases both with and without the extension field 🤔

pavelnikolov: Well you can just return:
```go
err := errors.QueryError{
    Message: "error message goes here",
    Extensions: map[string]interface{}{"key1": "val1", "key2": "val2"},
}

return nil, err
```

pavelnikolov: I guess we have to return some kind of a custom error that implements this interface (this interface will not be part of the library):
```go
type ResolverError interface {
    error
    Extensions() map[string]interface{}
}

```

Then at the top of `internal/exec/exec.go` define this interface:
```go
type extensionser interface {
	Extensions() map[string]interface{}
}
```

and later in `internal/exec/exec.go:190` the code will look like:
```go
if f.field.HasError && !callOut[1].IsNil() {
    resolverErr := callOut[1].Interface().(error)
    err := errors.Errorf("%s", resolverErr)
    err.Path = path.toSlice()
    if ex, ok := callOut[1].Interface().(extensionser) {
        err.Extensions = ex.Extensions()
    }
    return err
}
```

This way it will be up to the user to define any custom error type (e.g. `AuthenticationError`, `ForbiddenError`, `ValidationError`, `ServerError`, etc.) 
Ideally, this will also need to be documented with another example.

pavelnikolov:
You might as well consider sending an example with custom errors 🤔

ivanp:
Thanks! Got it, I'll send you some examples.

pavelnikolov:
The example needs to be added as a testable example in the `errors` package so that it appears here: https://godoc.org/github.com/graph-gophers/graphql-go/errors

PR #273, opened 2018-10-10
Give more details about errors while parsing want/got values

fadi-alkatut:
Add more info in error message

### DIFF
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -43,16 +43,17 @@ func RunTest(t *testing.T, test *Test) {
 		test.Context = context.Background()
 	}
 	result := test.Schema.Exec(test.Context, test.Query, test.OperationName, test.Variables)
-	checkErrors(t, test.ExpectedErrors, result.Errors)
-	if test.ExpectedResult != "" {
+	if result.Errors != nil {
+		checkErrors(t, test.ExpectedErrors, result.Errors)
+	} else {
 		// Verify JSON to avoid red herring errors.
 		got, err := formatJSON(result.Data)
 		if err != nil {
-			t.Fatalf("got: invalid JSON: %s\n response value is: [%s]", err, got)
+			t.Fatalf("got: invalid JSON: %s", err)
 		}
 		want, err := formatJSON([]byte(test.ExpectedResult))
 		if err != nil {
-			t.Fatalf("want: invalid JSON: %s\n expected value is: [%s]", err, want)
+			t.Fatalf("want: invalid JSON: %s", err)
 		}
 		if !bytes.Equal(got, want) {
 			t.Logf("got:  %s", got)
@@ -76,19 +77,24 @@ func formatJSON(data []byte) ([]byte, error) {
 
 func checkErrors(t *testing.T, expected, actual []*errors.QueryError) {
 	expectedCount, actualCount := len(expected), len(actual)
+
 	if expectedCount != actualCount {
-		t.Fatalf("unexpected number of errors: got %d, want %d\nerror(s):\n%+v", actualCount, expectedCount, actual)
+		t.Fatalf("unexpected number of errors: got %d, want %d", actualCount, expectedCount)
 	}
+
 	if expectedCount > 0 {
 		for i, want := range expected {
 			got := actual[i]
+
 			if !reflect.DeepEqual(got, want) {
 				t.Fatalf("unexpected error: got %+v, want %+v", got, want)
 			}
 		}
+
 		// Return because we're done checking.
 		return
 	}
+
 	for _, err := range actual {
 		t.Errorf("unexpected error: '%s'", err)
 	}
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -49,11 +49,11 @@ func RunTest(t *testing.T, test *Test) {
 		// Verify JSON to avoid red herring errors.
 		got, err := formatJSON(result.Data)
 		if err != nil {
-			t.Fatalf("got: invalid JSON: %s", err)
+			t.Fatalf("got: invalid JSON: %s\n response value is: [%s]", err, got)
 		}
 		want, err := formatJSON([]byte(test.ExpectedResult))
 		if err != nil {
-			t.Fatalf("want: invalid JSON: %s", err)
+			t.Fatalf("want: invalid JSON: %s\n expected value is: [%s]", err, want)
 		}
 		if !bytes.Equal(got, want) {
 			t.Logf("got:  %s", got)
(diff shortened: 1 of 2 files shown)


pavelnikolov:
Just add a unit test and I'd be happy to merge this.

gqltesting/testing.go
```diff
@@ -46,13 +46,14 @@ func RunTest(t *testing.T, test *Test) {
 	// Verify JSON to avoid red herring errors.
 	got, err := formatJSON(result.Data)
 	if err != nil {
-		t.Fatalf("got: invalid JSON: %s", err)
+		response := string(got[:])
```
pavelnikolov: This line is not needed, please delete it.

gqltesting/testing.go
```diff
@@ -46,13 +46,14 @@ func RunTest(t *testing.T, test *Test) {
...
+		t.Fatalf("got: invalid JSON: %s\n response value is: %s", err, response)
 	}
 	want, err := formatJSON([]byte(test.ExpectedResult))
 	if err != nil {
-		t.Fatalf("want: invalid JSON: %s", err)
+		expected := string(want[:])
```
pavelnikolov: This line is not needed, please delete it.

gqltesting/testing.go
```diff
@@ -46,13 +46,14 @@ func RunTest(t *testing.T, test *Test) {
 	// Verify JSON to avoid red herring errors.
 	got, err := formatJSON(result.Data)
 	if err != nil {
-		t.Fatalf("got: invalid JSON: %s", err)
+		response := string(got[:])
+		t.Fatalf("got: invalid JSON: %s\n response value is: %s", err, response)
```
pavelnikolov: Please, rewrite this as:
```Go
t.Fatalf("got: invalid JSON: %s\n response value is: %s", err, got)
```

gqltesting/testing.go
```diff
@@ -46,13 +46,14 @@ func RunTest(t *testing.T, test *Test) {
...
 	}
 	want, err := formatJSON([]byte(test.ExpectedResult))
 	if err != nil {
-		t.Fatalf("want: invalid JSON: %s", err)
+		expected := string(want[:])
+		t.Fatalf("want: invalid JSON: %s\n expected value is: %s", err, expected)
```
pavelnikolov: Please, rewrite this as:
```Go
t.Fatalf("want: invalid JSON: %s\n expected value is: %s", err, want)
```

fadi-alkatut:
Having more experimenting done on this issue I found the best solution is
- check for errors before expected result;
- print more details about unexpected errors;
- and only check for result if it's expected

gqltesting/testing.go
```diff
@@ -46,13 +46,11 @@ func RunTest(t *testing.T, test *Test) {
...
 	got, err := formatJSON(result.Data)
 	if err != nil {
-		response := string(got[:])
-		t.Fatalf("got: invalid JSON: %s\n response value is: %s", err, response)
+		t.Fatalf("got: invalid JSON: %s\n response value is: [%s]", err, got)
 	}
 	want, err := formatJSON([]byte(test.ExpectedResult))
 	if err != nil {
-		expected := string(want[:])
-		t.Fatalf("want: invalid JSON: %s\n expected value is: %s", err, expected)
```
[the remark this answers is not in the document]
fadi-alkatut: give more details on errors, but it wasn't very efficient

gqltesting/testing.go
```diff
@@ -43,22 +43,22 @@ func RunTest(t *testing.T, test *Test) {
...
-		t.Fatalf("want: invalid JSON: %s", err)
-	}
-
 	checkErrors(t, test.ExpectedErrors, result.Errors)
-
-	if !bytes.Equal(got, want) {
-		t.Logf("got:  %s", got)
-		t.Logf("want: %s", want)
-		t.Fail()
+	if test.ExpectedResult != "" {
```
pavelnikolov: Can we ever expect the result to be empty string? If yes, can you, please, reverse this condition:
```go
if test.ExpectedResult == "" {
    // handle this case
}
// unindent the rest of the code
```

graphql_test.go
```diff
@@ -247,57 +246,20 @@ func TestBasic(t *testing.T) {
...
-
-func (r *testNilInterfaceResolver) B() (interface{ Z() int32 }, error) {
-	return nil, errors.New("x")
-}
-
-func (r *testNilInterfaceResolver) C() (interface{ Z() int32 }, error) {
-	return nil, nil
-}
-
-func TestNilInterface(t *testing.T) {
```
pavelnikolov: Why is this test deleted? Do you need to rebase your PR?

gqltesting/testing.go
```diff
@@ -43,22 +43,23 @@ func RunTest(t *testing.T, test *Test) {
...
-
-	checkErrors(t, test.ExpectedErrors, result.Errors)
-
-	if !bytes.Equal(got, want) {
-		t.Logf("got:  %s", got)
-		t.Logf("want: %s", want)
-		t.Fail()
+	if result.Errors != nil {
+		checkErrors(t, test.ExpectedErrors, result.Errors)
+	} else {
```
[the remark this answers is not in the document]
pavelnikolov: For example:
```go
if result.Errors != nil && checkErrors(t, test.ExpectedErrors, result.Errors) {
    return
}
```
or you can even move the `results.Errors != nil` check inside of the `checkErrors` func. And if it returns `true` you can exit early, no? WDYT?

gqltesting/testing.go
```diff
@@ -43,17 +43,14 @@ func RunTest(t *testing.T, test *Test) {
...
+	checkErrors(t, test.ExpectedErrors, result.Errors)
 	// Verify JSON to avoid red herring errors.
 	got, err := formatJSON(result.Data)
-	if err != nil {
+	if err != nil && test.ExpectedResult != "" {
 		t.Fatalf("got: invalid JSON: %s\n response value is: [%s]", err, got)
 	}
 	want, err := formatJSON([]byte(test.ExpectedResult))
-	if err != nil {
+	if err != nil && test.ExpectedResult != ""  {
```
pavelnikolov: Why is this required? And why do you skip it if the expected result is `""`?

pavelnikolov: After a second look, I understood what's going on. In case of an error, fail only if there is expected result.

fadi-alkatut: Yes, because if an error happens and returns empty response it will give error when doing  `formatJSON` and this will fail the test even if you are expecting the error and an the empty result.

pavelnikolov:
@fadi-alkatut this PR now has conflicts. Do you think it's worth resolving these? Or should the PR be closed?

pavelnikolov:
@fadi-alkatut ping

fadi-alkatut:
I think we went with another PR that had a cleaner solution for the problem. So this PR isn’t relevant anymore.

PR #276, opened 2018-10-20
First test for errors, then for content

alicebob:
If there are serious errors the returned `Data` field might be empty, which is a JSON parse error. So we need to check errors first, then the rest.

This is half of #267, which @pavelnikolov requested me to split up.

> Also, you might consider adding some unit tests and asses that you get correct message when you move the error checking first?

I wouldn't mind, but frankly I have no idea how to make unittests for unittests :(
I do want to refer to the comments in #267 which show a before/after example i got from using this in my code.

### DIFF
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -43,6 +43,9 @@ func RunTest(t *testing.T, test *Test) {
 		test.Context = context.Background()
 	}
 	result := test.Schema.Exec(test.Context, test.Query, test.OperationName, test.Variables)
+
+	checkErrors(t, test.ExpectedErrors, result.Errors)
+
 	// Verify JSON to avoid red herring errors.
 	got, err := formatJSON(result.Data)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


gqltesting/testing.go
```diff
@@ -43,6 +43,9 @@ func RunTest(t *testing.T, test *Test) {
 		test.Context = context.Background()
 	}
 	result := test.Schema.Exec(test.Context, test.Query, test.OperationName, test.Variables)
+
+	checkErrors(t, test.ExpectedErrors, result.Errors)
+
 	// Verify JSON to avoid red herring errors.
```
pavelnikolov: If we check the errors first, this comment doesn't make sense anymore. Please, delete it.

pavelnikolov (changes requested):
Please, remove the comment and rebase the PR against master.

pavelnikolov:
If we check for errors first, we might have invalid JSON to compare and then that would also lead to confusing errors.

PR #277, opened 2018-10-20
simpler error comparisons

alicebob:
Before you would get the unhelpful message:

    testing.go:82: unexpected number of errors: got 1, want 0

but now you get:

    testing.go:80: unexpected error: got [graphql: Cannot query field "timestamp" on type "SMSList". (line 1, column 40)], want []

This is half of #267, which @pavelnikolov asked me to split up.

### DIFF
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -74,27 +74,8 @@ func formatJSON(data []byte) ([]byte, error) {
 	return formatted, nil
 }
 
-func checkErrors(t *testing.T, expected, actual []*errors.QueryError) {
-	expectedCount, actualCount := len(expected), len(actual)
-
-	if expectedCount != actualCount {
-		t.Fatalf("unexpected number of errors: got %d, want %d", actualCount, expectedCount)
-	}
-
-	if expectedCount > 0 {
-		for i, want := range expected {
-			got := actual[i]
-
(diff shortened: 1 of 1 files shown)


gqltesting/testing.go
```diff
@@ -74,27 +74,8 @@ func formatJSON(data []byte) ([]byte, error) {
...
-
-		// Return because we're done checking.
-		return
-	}
-
-	for _, err := range actual {
-		t.Errorf("unexpected error: '%s'", err)
+func checkErrors(t *testing.T, want, got []*errors.QueryError) {
+	if !reflect.DeepEqual(got, want) {
+		t.Fatalf("unexpected error: got %+v, want %+v", got, want)
```
pavelnikolov: This message is not correct anymore because you compare collections of errors and not single errors.

fadi-alkatut:
but do you think there is a need to move `checkErrors` call above this line for it to work? `got, err := formatJSON(result.Data)` because in my case the test was terminated on the error check after that line due to empty response without reaching the `checkErrors`

pavelnikolov:
I merged this as it is a step in the right direction. @fadi-alkatut is correct and indeed it can be improved further.

alicebob:
Thanks!

PR #282, opened 2018-10-30
Use Struct Field Resolver instead of Method

### LINKED ISSUE #28: Use field as resolver instead of method.

salman-ahmad:
This fixes #28. Essentially, it reduces lot of boilerplate code by making use of struct field as resolvers instead of creating methods for each struct field. 

Now, methods are required **only when**

1. a struct implements a GraphQL interface. In this case, you create methods for fields which overlap
1. a struct exposes a GraphQL field that accepts additional arguments i.e., `first: Int, last: Int` etc. 
1. there is GraphQL `union` type

By using struct fields as resolvers, one could also benefit from DRY codebase by not having a pair of a struct which look the same (one for GraphQL and one for DB)

### Does this break existing API? 

No. This change is completely transparent and you can continue using your existing codebase. 

### Can I still continue using methods as resolvers?

Yes, by default, it uses method resolvers. Even when `UseFieldResolvers()` is turned on, the method is given higher priority than the field. When there is no method, the field will be used if there is any. 

### How can I use struct fields as resolvers?

When invoking `graphql.ParseSchema()` or `graphql.MustParseSchema()` to parse the schema, you pass in `graphql.UseFieldResolvers()` as an argument. For an example, take a look at `./example/social/server/server.go`

### Additional Notes

- for an example, take a look at `./example/social/social.go` & `./example/social/server/server.go`
- Unit tests are missing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
 /internal/validation/testdata/graphql-js
 /internal/validation/testdata/node_modules
 /vendor
-.DS_Store
-.idea/
-.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 /internal/validation/testdata/graphql-js
 /internal/validation/testdata/node_modules
 /vendor
+.DS_Store
+.idea/
+.vscode/
--- a/README.md
+++ b/README.md
@@ -72,10 +72,10 @@ opts := []graphql.SchemaOpt{graphql.UseFieldResolvers()}
 schema := graphql.MustParseSchema(s, &query{}, opts...)
 ```   
 
-When using `UseFieldResolvers`, a field will be used *only* when:
-- there is no method
-- it does not implement an interface
-- it does not have arguments 
+When using `UseFieldResolvers` schema option, a struct field will be used *only* when:
+- there is no method for a struct field
+- a struct field does not implement an interface method
+- a struct field does not have arguments
 
 The method has up to two arguments:
 
--- a/README.md
+++ b/README.md
@@ -65,7 +65,17 @@ $ curl -XPOST -d '{"query": "{ hello }"}' localhost:8080/query
 
 ### Resolvers
 
-A resolver must have one method for each field of the GraphQL type it resolves. The method name has to be [exported](https://golang.org/ref/spec#Exported_identifiers) and match the field's name in a non-case-sensitive way.
+A resolver must have one method or field for each field of the GraphQL type it resolves. The method or field name has to be [exported](https://golang.org/ref/spec#Exported_identifiers) and match the schema's field's name in a non-case-sensitive way.
+You can use struct fields as resolvers by using `SchemaOpt: UseFieldResolvers()`. For example,
+```
+opts := []graphql.SchemaOpt{graphql.UseFieldResolvers()}
+schema := graphql.MustParseSchema(s, &query{}, opts...)
+```   
+
+When using `UseFieldResolvers`, a field will be used *only* when:
+- there is no method
+- it does not implement an interface
+- it does not have arguments 
 
 The method has up to two arguments:
 
--- a/graphql.go
+++ b/graphql.go
@@ -2,9 +2,8 @@ package graphql
 
 import (
 	"context"
-	"fmt"
-
 	"encoding/json"
+	"fmt"
 
 	"github.com/graph-gophers/graphql-go/errors"
 	"github.com/graph-gophers/graphql-go/internal/common"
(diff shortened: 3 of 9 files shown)


bazaglia:
@pavelnikolov are you OK with merging this?

nanozuki:
I think this is a really useful feature and can't wait to use it. What't the plan for merging?

pavelnikolov:
I'll review the PR shortly. I'm sorry for the delay on this.

eveld:
Can't wait to throw away half of my resolver code after this is merged! :)

jenting:
Any update for this PR be merged?

.gitignore
```diff
@@ -1,3 +1,6 @@
 /internal/validation/testdata/graphql-js
 /internal/validation/testdata/node_modules
 /vendor
+.DS_Store
+.idea/
+.vscode/
```
pavelnikolov: Unrelated change! Please, restore this file - these need to be in your global ignore file.

salman-ahmad: I have restored this file

salman-ahmad:
@pavelnikolov I have addressed your review comment. Please merge this PR, thank you1

ItsLeeOwen:
Thank the gods, and @salman-ahmad  I was just looking for this functionality.  ❤️

pavelnikolov:
> I would suggest that graphql.UseFieldResolvers() possibly become the default.

@ItsLeeOwen , your suggestion would introduce a breaking change which I would really like to avoid. Instead, this feature will be opt-in.

ItsLeeOwen:
@pavelnikolov   Cool, I'm curious what would that breaking change be?

pavelnikolov:
@ItsLeeOwen oh, my bad 🤦‍♂️ 
There is no breaking change. I got confused..

README.md
```diff
@@ -65,7 +65,17 @@ $ curl -XPOST -d '{"query": "{ hello }"}' localhost:8080/query
...
 
-A resolver must have one method for each field of the GraphQL type it resolves. The method name has to be [exported](https://golang.org/ref/spec#Exported_identifiers) and match the field's name in a non-case-sensitive way.
+A resolver must have one method or field for each field of the GraphQL type it resolves. The method or field name has to be [exported](https://golang.org/ref/spec#Exported_identifiers) and match the schema's field's name in a non-case-sensitive way.
+You can use struct fields as resolvers by using `SchemaOpt: UseFieldResolvers()`. For example,
+```
+opts := []graphql.SchemaOpt{graphql.UseFieldResolvers()}
+schema := graphql.MustParseSchema(s, &query{}, opts...)
+```   
+
+When using `UseFieldResolvers`, a field will be used *only* when:
```
pavelnikolov: The above needs to be clarified a little, for example:
> When using `UseFieldResolvers` schema option, a struct field will be used *only* when:

README.md
```diff
@@ -65,7 +65,17 @@ $ curl -XPOST -d '{"query": "{ hello }"}' localhost:8080/query
...
+- there is no method
+- it does not implement an interface
```
pavelnikolov: > it does not implement an interface

What does "it" stand for in this sentence? I understand it but a first-time user of this library might not understand this. Can you, please, put a more descriptive sentence?

README.md
```diff
@@ -65,7 +65,17 @@ $ curl -XPOST -d '{"query": "{ hello }"}' localhost:8080/query
...
+- it does not have arguments
```
pavelnikolov: Same here - "it" is ambiguous for new starters.

graphql.go
```diff
@@ -84,6 +83,13 @@ func UseStringDescriptions() SchemaOpt {
 	}
 }
 
+// Specifies whether to use struct field resolvers
```
pavelnikolov: Comments need to start with the func/method name
https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

Please, use:
```
UseFieldResolvers...
```

pavelnikolov:

tobstarr:
:tada: just at the right time! Thank you so much!

marcelvanwaaijen:
Question about the social example: What is the use of the FriendsResolver method? When is it called? I put some debug logging in it, but that never shows, also the pagination doesn't seem to be doing anything currently?

pavelnikolov:
@marcelvanwaaijen I haven't looked into it in details but my wild guess is that it was an existing example and when this feature was implemented we ended up adding `graphql.UseFieldResolvers()` which uses fields instead of methods. Then we started using the fields and stopped using the method. I would suggest opening a separate issue if this is the case as this one was related to adding the field resolvers feature, which is already done.

PR #289, opened 2018-11-13
Switch to interface{} for Subscribe() return

matiasanaya:
**Breaking change (subscriptions)**

It's a lot easier to build `graph-gophers/graphql-transport-ws` around the `interface{}` type.

This PR is to make the changes in [this other PR](https://github.com/graph-gophers/graphql-transport-ws/pull/2) possible.

### DIFF
--- a/subscriptions.go
+++ b/subscriptions.go
@@ -19,14 +19,14 @@ import (
 // If the context gets cancelled, the response channel will be closed and no
 // further resolvers will be called. The context error will be returned as soon
 // as possible (not immediately).
-func (s *Schema) Subscribe(ctx context.Context, queryString string, operationName string, variables map[string]interface{}) (<-chan *Response, error) {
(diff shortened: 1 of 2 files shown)


matiasanaya:
I really would prefer to return `json.RawMessage` but this library favours `interface{}` as far as I can tell. Thoughts?

subscriptions.go
```diff
@@ -19,14 +19,14 @@ import (
 // If the context gets cancelled, the response channel will be closed and no
 // further resolvers will be called. The context error will be returned as soon
 // as possible (not immediately).
-func (s *Schema) Subscribe(ctx context.Context, queryString string, operationName string, variables map[string]interface{}) (<-chan *Response, error) {
+func (s *Schema) Subscribe(ctx context.Context, queryString string, operationName string, variables map[string]interface{}) (<-chan interface{}, error) {
```
pavelnikolov: What's the benefit of this change?

Issue #153, opened 2018-01-22
Support extend type syntax


wenerme:
I want to build several schema for different tenant, I want to some type is extendable, for user, if some module is enabled, the new query function will added to user type. 

From this article [Extend your Graphcool API with Resolvers](https://blog.graph.cool/extend-your-graphcool-api-with-resolvers-ca0f0270bca7), found the syntax like

```graphql
extend type Query {
  getWeatherByCity(city: String!): GetWeatherByCityPayload
}
```

Is it possible to implement this syntax in graphql-go ? Just merge the field to the main type, or is there any way to do this in graphql-go except dynamic generate schema ?

wenerme:
Seems, it's not too hard to do this, the worst implementation is 

```go
		case "extend":
			switch t:=l.ConsumeIdent();t {
			case "type":
				obj := parseObjectDecl(l)
				obj.Desc = desc
				old := s.Types[obj.Name]
				if old == nil {
					l.SyntaxError(fmt.Sprintf(`extend type %q not found`, obj.Name))
				}
				origin, ok := old.(*Object)
				if !ok {
					l.SyntaxError(fmt.Sprintf(`extend type %q must be object, but found %q`, obj.Name, old.Kind()))
				}
				origin.Fields = append(origin.Fields, obj.Fields...)
			default:
				l.SyntaxError(fmt.Sprintf(`unexpected %q, expecting "type"`, t))
			}
```

add this to `schema.go`, and it works, maybe the extend type should be tracked in Object, can do the same for interface or input, very interesting.

wenerme:
I also suggest support `extend <type> <ObjectName> by <ObjectName>`.
Because, I we always define types for different module

```graphql
type AQuery{
doA():A
}
type BQuery{
doB():B
}

# If we want to enable the module
extend type Query by AQuery
```

Very useful for real word usage. I will try to do this in my own fork.

wenerme:
This syntax is implemented in [goaphql](https://github.com/wenerme/goaphql), and several other very useful syntax.

For bigger scale use, I think how to construct the schema is become more important, some syntax like this become very useful.

```graphql
# file core.graphqls - basic difinitions

scalar Date
scalar Html

directive @role(name:String!)

# file user.graphqls - api for user
type UserQuery{}
type UserMutation{}

schema User{
    query: UserQuery
    mutation: UserMutation
}

# file admin.graphqls - api for amdin
type AdminQuery{}
type AdminMutation{}

[...]

PR #293, opened 2018-12-01
Add Support For extensions

srikrsna:
This PR adds support for `extend`. See #153.

I've added initial support for parsing schema with extend. All current tests pass. Please let me know how should I go about this.

### DIFF
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -45,6 +45,7 @@ type Schema struct {
 	objects         []*Object
 	unions          []*Union
 	enums           []*Enum
+	extensions      []*Extension
 }
 
 // Resolve a named type in the schema by its name.
@@ -157,6 +158,15 @@ type InputObject struct {
 	// TODO: Add a list of directives?
 }
 
+// Extension type defines a graphql type extension.
+// Schema, Object, Input, and Scalars can be extended.
+//
+// https://facebook.github.io/graphql/draft/#sec-Type-System-Extensions
+type Extension struct {
+	Type NamedType
+	// TODO: Add a list og directives
+}
+
 // FieldsList is a list of an Object's Fields.
 //
 // http://facebook.github.io/graphql/draft/#FieldsDefinition
@@ -254,6 +264,10 @@ func (s *Schema) Parse(schemaString string, useStringDescriptions bool) error {
 		return err
 	}
 
+	if err := mergeExtensions(s); err != nil {
+		return err
+	}
+
 	for _, t := range s.Types {
 		if err := resolveNamedType(s, t); err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


hampusohlsson:
+1 what's the status of this PR? Happy to contribute if help is needed

internal/schema/schema.go
```diff
@@ -322,6 +336,88 @@ func (s *Schema) Parse(schemaString string, useStringDescriptions bool) error {
...
 
+func mergeExtensions(s *Schema) error {
+	for _, ext := range s.extensions {
+		typ := s.Types[ext.Type.TypeName()]
+		if typ == nil {
+			return fmt.Errorf("trying to extend of unknown type %q", ext.Type.TypeName())
+		}
+
+		if typ.Kind() != ext.Type.Kind() {
+			return fmt.Errorf("trying to extend of type %q with type %q", typ.Kind(), ext.Type.Kind())
```
pavelnikolov: There is additional `of` in the error message. Maybe, it should be: 
```
"trying to extend type %q with type %q"
```

internal/schema/schema.go
```diff
@@ -322,6 +336,88 @@ func (s *Schema) Parse(schemaString string, useStringDescriptions bool) error {
...
+		}
+
+		switch og := typ.(type) {
+		case *Object:
+			e := ext.Type.(*Object)
+
+			for _, field := range e.Fields {
+				if og.Fields.Get(field.Name) != nil {
+					return fmt.Errorf("extended field already available %q", field.Name)
```
pavelnikolov: Is it not possible to override fields of extended types?

internal/schema/schema.go
```diff
@@ -322,6 +336,88 @@ func (s *Schema) Parse(schemaString string, useStringDescriptions bool) error {
...
+			for _, en := range e.Values {
+				for _, on := range og.Values {
+					if on.Name == en.Name {
+						return fmt.Errorf("union type provided again in extension %q for %q", on, og.Name)
+					}
+				}
+			}
+			og.Values = append(og.Values, e.Values...)
+		default:
+			return fmt.Errorf(`unexpected %q, expecting "schema", "type", "enum", "interface", "union" or "input"`, og.TypeName())
```
pavelnikolov: How about `scalar` types?

internal/schema/schema.go
```diff
@@ -442,13 +538,55 @@ func parseSchema(s *Schema, l *common.Lexer) {
...
+
 		default:
 			// TODO: Add support for type extensions.
 			l.SyntaxError(fmt.Sprintf(`unexpected %q, expecting "schema", "type", "enum", "interface", "union", "input", "scalar" or "directive"`, x))
 		}
 	}
 }
 
+func parseExtension(s *Schema, l *common.Lexer) {
+	//l.ConsumeWhitespace()
```
pavelnikolov: This comment is not needed. Please, remove it.

pavelnikolov:
@srikrsna thank you very much for this PR. I'd be happy to see extensions support in this library.
@hampusohlsson I would like to proceed with this PR and in my opinion we need (in no specific order):

- unit tests
- some documentation or at least update of the README.md that this feature has been added
- extensions example (not necessarily part of this PR) either in the Star Wars example or a separate one dedicated to this feature

PR #296, opened 2018-12-18
Implement error propagation

tinnywang:
This is based on https://github.com/graph-gophers/graphql-go/pull/229, with some additions/changes to handle the following:
- non-nullable fields that resolve to `null`
- non-nullable lists of nullable elements (`[]!`)
- nullable lists of nullable elements (`[]`)

Additionally, to determine if a parent field resolves to `null`, check if any immediate child resolved to `null` instead of checking if any descendant returned an error.

### DIFF
--- a/gqltesting/testing.go
+++ b/gqltesting/testing.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"reflect"
+	"sort"
 	"strconv"
 	"testing"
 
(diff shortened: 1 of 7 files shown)


gqltesting/testing.go
```diff
@@ -76,7 +78,19 @@ func formatJSON(data []byte) ([]byte, error) {
...
 		t.Fatalf("unexpected error: got %+v, want %+v", got, want)
 	}
 }
+
+func sortErrors(errors []*errors.QueryError) {
+	if len(errors) <= 1 {
+		return
+	}
+	sort.Slice(errors, func(i, j int) bool {
+		return fmt.Sprintf("%s", errors[i].Path) < fmt.Sprintf("%s", errors[j].Path)
```
pavelnikolov: Why do you need the `fmt.Sprintf` here if it's just printing the same value?

pavelnikolov: Oh, ignore that... I just noticed that `Path` is of type `[]interface{}`

pavelnikolov:
@tinnywang, thank you! I'll review your this PR tonight and will get back to you.

subscription_test.go
```diff
@@ -100,11 +145,7 @@ func TestSchemaSubscribe(t *testing.T) {
 				},
 				{
 					Data: json.RawMessage(`
-						{
-							"helloSaid": {
-								"msg":null
-							}
-						}
+						null
```
pavelnikolov: This unit test case perfectly demonstrates the change

pavelnikolov:
Thank you very much for your contribution and effort to better implement the GraphQL spec.

PR #300, opened 2018-12-31
convert to go module

joefitzgerald:
This PR converts `graphql-go` to a [go module](https://blog.golang.org/modules2019). Ideally, this would be paired with a release (perhaps a `v0.x.y` version) so that this package can be released periodically.


pavelnikolov:
Thank you for your contribution

PR #301, opened 2019-01-07
Add a logging hook for telemetry

tdb-alcorn:
I'm putting this up here to see if there is any interest in merging it. If so, I'll go through and fix up the imports to point to this repo again.

### DIFF
--- a/internal/exec/packer/packer.go
+++ b/internal/exec/packer/packer.go
@@ -354,8 +354,21 @@ func unmarshalInput(typ reflect.Type, input interface{}) (interface{}, error) {
 		if reflect.TypeOf(input).ConvertibleTo(typ) {
 			return reflect.ValueOf(input).Convert(typ).Interface(), nil
 		}
+
+	case reflect.Ptr:
+		value := reflect.ValueOf(input)
+		if value.CanAddr() {
+			return value.Addr().Interface(), nil
+		}
+		newValuePtr := reflect.New(typ.Elem())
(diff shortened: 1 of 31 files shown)


pavelnikolov:
What problem are you trying to solve? Can you give a more specific example, please?

tdb-alcorn:
This PR does three things:

1. A bug fix around hanging validation
2. Enable unmarshaling into pointer values as long as they point to a value of the correct type (I consider this a bug fix, it's clearly something a graphql library should do)
3. Add logging to the validator to enable telemetry e.g. logging the query name (we use this to profile our API by query type)

internal/exec/packer/packer.go
```diff
@@ -354,8 +354,21 @@ func unmarshalInput(typ reflect.Type, input interface{}) (interface{}, error) {
...
+		value := reflect.ValueOf(input)
+		if value.CanAddr() {
+			return value.Addr().Interface(), nil
+		}
+		newValuePtr := reflect.New(typ.Elem())
+		newValue := reflect.Indirect(newValuePtr)
+		if newValue.CanSet() {
+			newValue.Set(value)
+			return newValuePtr.Interface(), nil
+		}
```
pavelnikolov: This is unrelated change.

pavelnikolov:
Thanks you. Could you, please, create issues for these problems first and after we discuss the issues you can send one PR for each of the issues. IMHO, you should not try to solve multiple problems in a single PR. I'm closing this for now.

PR #309, opened 2019-02-14
Add DisableIntrospection SchemaOpt

svanburen:
Allows the server to disable schema introspection.

This basically just gates the `__type`, `__typename`, and `__schema` introspection schema behind a `SchemaOpt`. gqlgen recently implemented similar functionality (https://github.com/99designs/gqlgen/pull/447), albeit also on a per-request basis. I think this functionality would be quite desired for running GraphQL servers in production.

Thoughts?


pavelnikolov:

PR #313, opened 2019-03-27
add more community example

wuriyanto48:
Hello, i create an example of `graphql go`. I also integrate this example with echo framework, GORM, Postgres database and unit testing.
thank you


pavelnikolov:
I'm closing this PR because the example has a lot of dependencies. They will get outdated very quickly and will not be relevant to most users with time.

PR #318, opened 2019-04-12
Handle large Float value that come without decimal point

dliusc:
This change is to make the library more tolerant to the large Float value number that is greater than `2^32-1` but doesn't come with decimal point `.`
Currently, when a Float input takes a number that is greater than 32 bit, the library crashes.  E.g.
```
"graphql: panic occurred: strconv.ParseInt: parsing \"1554745767989\": value out of range"`. 
```

### DIFF
--- a/internal/common/literals.go
+++ b/internal/common/literals.go
@@ -24,10 +24,19 @@ func (lit *BasicLit) Value(vars map[string]interface{}) interface{} {
 	switch lit.Type {
 	case scanner.Int:
 		value, err := strconv.ParseInt(lit.Text, 10, 32)
-		if err != nil {
-			panic(err)
+		if err == nil {
+			return int32(value)
 		}
-		return int32(value)
+		if numError, ok := err.(*strconv.NumError); ok {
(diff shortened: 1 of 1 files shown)


internal/common/literals.go
```diff
@@ -24,10 +24,19 @@ func (lit *BasicLit) Value(vars map[string]interface{}) interface{} {
...
+			if numError.Err == strconv.ErrRange {
+				val, e := strconv.ParseFloat(lit.Text, 64)
+				if e != nil {
+					panic(e)
+				}
+				return val
+			}
 		}
-		return int32(value)
+		panic(err)
```
pavelnikolov: Is it possible to use `fallthrough` instead? Also, I'd prefer the happy paths to be last, for example:
```go
if err != nil {
	if nerr, ok := err.(*strconv.NumError); ok && nerr.Err == strconv.ErrRange {
		fallthrough // go to the parse float case
	}
	panic(err)
}
return int32(value)
```

PR #321, opened 2019-04-24
Fix subscribe panic

billettc:
Handle errs from request after ApplyOperation is executed
Prevent panic when "var f *fieldToExec" is nil
Added field "name" on pack error

### DIFF
--- a/internal/exec/subscribe.go
+++ b/internal/exec/subscribe.go
@@ -23,7 +23,7 @@ type Response struct {
 func (r *Request) Subscribe(ctx context.Context, s *resolvable.Schema, op *query.Operation) <-chan *Response {
 	var result reflect.Value
 	var f *fieldToExec
-	var err *errors.QueryError
+	var errs []*errors.QueryError
 	func() {
 		defer r.handlePanic(ctx)
 
@@ -31,9 +31,14 @@ func (r *Request) Subscribe(ctx context.Context, s *resolvable.Schema, op *query
 		var fields []*fieldToExec
 		collectFieldsToResolve(sels, s.Resolver, &fields, make(map[string]*fieldToExec))
 
+		if len(r.Errs) > 0 {
+			errs = r.Errs
+			return
+		}
+
 		// TODO: move this check into validation.Validate
 		if len(fields) != 1 {
-			err = errors.Errorf("%s", "can subscribe to at most one subscription at a time")
+			errs = []*errors.QueryError{errors.Errorf("%s", "can subscribe to at most one subscription at a time")}
 			return
 		}
 		f = fields[0]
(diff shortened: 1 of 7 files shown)


pavelnikolov:
What is the panic that you faced before? Can you, please, provide a simple example to reproduce the issue that this PR fixes?

internal/exec/subscribe.go
```diff
@@ -23,17 +23,22 @@ type Response struct {
...
 
+		if len(r.Errs) > 0 {
+			errs = r.Errs
+			return
+		}
+
 		// TODO: move this check into validation.Validate
 		if len(fields) != 1 {
-			err = errors.Errorf("%s", "can subscribe to at most one subscription at a time")
+			errs = []*errors.QueryError{errors.Errorf("%s", "can subscribe to at most one subscription at a time")}
```
pavelnikolov: this doesn't make any sense to be `errors.Errorf`. It can be `errors.New` instead.

internal/exec/subscribe.go
```diff
@@ -50,16 +55,22 @@ func (r *Request) Subscribe(ctx context.Context, s *resolvable.Schema, op *query
 
 		if f.field.HasError && !callOut[1].IsNil() {
 			resolverErr := callOut[1].Interface().(error)
-			err = errors.Errorf("%s", resolverErr)
+			err := errors.Errorf("%s", resolverErr)
```
pavelnikolov: same here - this needs to be `errors.New`

internal/exec/subscribe.go
```diff
@@ -50,16 +55,22 @@ func (r *Request) Subscribe(ctx context.Context, s *resolvable.Schema, op *query
...
 			err.ResolverError = resolverErr
+			errs = []*errors.QueryError{err}
 		}
 	}()
 
-	if err != nil {
-		if _, nonNullChild := f.field.Type.(*common.NonNull); nonNullChild {
-			return sendAndReturnClosed(&Response{Errors: []*errors.QueryError{err}})
+	if len(errs) > 0 {
+
```
pavelnikolov: No need for this empty line here.

internal/exec/subscribe.go
```diff
@@ -50,16 +55,22 @@ func (r *Request) Subscribe(ctx context.Context, s *resolvable.Schema, op *query
...
+		var nonNullChild bool
```
pavelnikolov: And here

subscription_test.go
```diff
@@ -121,8 +121,8 @@ func TestSchemaSubscribe(t *testing.T) {
 				helloSaidResolver: &helloSaidResolver{
 					upstream: closedUpstream(
 						&helloSaidEventResolver{msg: "Hello world!"},
-						&helloSaidEventResolver{err: resolverErr},
 						&helloSaidEventResolver{msg: "Hello again!"},
+						&helloSaidEventResolver{err: resolverErr},
```
pavelnikolov: Why is this test changed? Shoudln't you leave it intact and add new tests that used to panic before but don't do it anymore after this fix?

pavelnikolov:
There are conflicts to be resolved.

Issue #195, opened 2018-04-12
Data race on schema parsing


vladimiroff:
I've stumbled upon this in a service where schema is not parsed globally in a `init` func, but passed as a dependency to given resolver. Placing `t.Parallel()` before schema parsing causes data races found by the race detector. Placing it after the parsing solves the issue. Decided to try and reproduce it with the starwars example.

Example tests over `github.com/graph-gophers/graphql-go/example/starwars` with `t.Parallel()`:

```go
package starwars_test

import (
        "testing"

        graphql "github.com/graph-gophers/graphql-go"
        "github.com/graph-gophers/graphql-go/example/starwars"
)

func TestParse1(t *testing.T) {
        t.Parallel()
        graphql.ParseSchema(starwars.Schema, &starwars.Resolver{})
}

func TestParse2(t *testing.T) {
        t.Parallel()
        graphql.ParseSchema(starwars.Schema, &starwars.Resolver{})
}
```

When executed with the race detector:

```
==================
WARNING: DATA RACE
Read at 0x00c4200b8928 by goroutine 6:
  github.com/graph-gophers/graphql-go/internal/schema.resolveField()
      /home/kiril/go/src/github.com/graph-gophers/graphql-go/internal/schema/schema.go:348 +0xc5
  github.com/graph-gophers/graphql-go/internal/schema.resolveNamedType()
      /home/kiril/go/src/github.com/graph-gophers/graphql-go/internal/schema/schema.go:329 +0x2f5
  github.com/graph-gophers/graphql-go/internal/schema.(*Schema).Parse()
      /home/kiril/go/src/github.com/graph-gophers/graphql-go/internal/schema/schema.go:258 +0x211
  github.com/graph-gophers/graphql-go.ParseSchema()
      /home/kiril/go/src/github.com/graph-gophers/graphql-go/graphql.go:37 +0x23a
  github.com/graph-gophers/graphql-go/example/starwars_test.TestParse1()
      /home/kiril/go/src/github.com/graph-gophers/graphql-go/example/starwars/starwars_test.go:12 +0xa9
  testing.tRunner()
      /usr/lib/go/src/testing/testing.go:777 +0x16d

Previous write at 0x00c4200b8928 by goroutine 7:
  github.com/graph-gophers/graphql-go/internal/schema.resolveField()
      /home/kiril/go/src/github.com/graph-gophers/graphql-go/internal/schema/schema.go:352 +0x131
[...]

PR #322, opened 2019-04-26
Resolve schema parsing data races

dackroyd:
Moving package globals to per-schema objects, resolving data races from parsing multiple schemas in parallel. While this has limited likelihood in production code (since typically just 1 schema would be used), tests are a different story

This change is _not_ well tested. I've verified that the race detection is no longer triggered within this project, or within one of my own that uses the library, while the tests still pass.

The CI build for the project will also need to be updated to execute tests with the race detector enabled going forward.

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -164,6 +164,8 @@ func (r *discussPlanResolver) DismissVader(ctx context.Context) (string, error) 
 }
 
 func TestHelloWorld(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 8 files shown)


dackroyd:
Fixes #195

bmhatfield:
Cross-referencing this issue back to this PR: https://github.com/graph-gophers/graphql-go/issues/327

This PR introduces a panic on schema introspection for me. Checking out an earlier commit resolves the issue.

Issue #325, opened 2019-05-03
Invalid enum values do not result in GraphQL errors


dackroyd:
Enum values supplied via input variables or returned in responses are not correctly applying the coercion rules defined by the GraphQL spec. This was previously raised by issue #262, however that issue was incorrectly closed.

https://graphql.github.io/graphql-spec/June2018/#sec-Enums

> **Result Coercion**
> GraphQL servers must return one of the defined set of possible values. If a reasonable coercion is not possible they must raise a field error.

> **Input Coercion**
> GraphQL has a constant literal to represent enum input values. GraphQL string literals must not be accepted as an enum input and instead raise a query error.

https://graphql.github.io/graphql-spec/June2018/#sec-Coercing-Variable-Values

> If value cannot be coerced according to the input coercion rules of variableType, throw a query error.

https://graphql.github.io/graphql-spec/June2018/#sec-Value-Completion

> If fieldType is a Scalar or Enum type:
Return the result of “coercing” result, ensuring it is a legal value of fieldType, otherwise null.

The behaviour as currently implemented:

**Invalid enum value declared within query string:** ✅ PASS

Query:
```
query HeroForEpisode {
	hero(episode: WRATH_OF_KHAN) {
		name
	}
}
```

Result:
```
{
  "errors": [
    {
      "message": "Argument \"episode\" has invalid value WRATH_OF_KHAN.\nExpected type \"Episode\", found WRATH_OF_KHAN.",
      "locations": [{"line": 3, "column": 20}]
    }
  ]
}
```

**Invalid enum value declared within input variables:** ❌ FAIL

Query:
```
query HeroForEpisode($episode: Episode!) {
	hero(episode: $episode) {
		name
	}
}
```

Variables: `{"episode": "FINAL_FRONTIER"}`

Result:
```
{
  "data": {
    "hero": {
      "name": "R2-D2"
     }
  }
}
```

Expected:
```
{
  "errors": [
    {
      "message": "Argument \"episode\" has invalid value FINAL_FRONTIER.\nExpected type \"Episode\", found FINAL_FRONTIER.",
      "locations": [{"line": 3, "column": 20}]
    }
  ]
}
```

**Invalid enum value returned by a resolver:** ❌ FAIL

Query:
```
query Hero {
	hero {
		name
		appearsIn
	}
}
```

Result:
```
{
  "data": {
    "hero": {
      "name": "Spock",
      "appearsIn": ["STAR_TREK"]
    }
  }
}
```

[...]

PR #326, opened 2019-05-03
Fix Enum validation

dackroyd:
Validating enum values supplied via variable input, and those returned from resolvers, ensuring that only valid enum values defined by the GraphQL schema are accepted


dackroyd:
Fixed #325

PR #329, opened 2019-05-08
Add option for custom panic handler

hampusohlsson:
This PR adds an option to provide a custom error handler for panics.

By giving more control over the panic error creation, it's possible for the consumer of this library to capture the panics and format the message, add extenstions and send to 3rd party error reporting services.

I opted for keeping existing `log.LogPanic` functionality to not make breaking changes. But you can write your own panic handler that implements both interfaces. Here's an example of how I'm using a custom panic handler. 

```go
// main.go
...
panicHandler := &myerrorlib.PanicHandler{}
rootSchema := graphql.MustParseSchema(
	schema.GetRootSchema(),
	resolvers.NewRootResolver(),
	graphql.Logger(panicHandler),
	graphql.PanicHandler(panicHandler),
)
...
```
and then in my case I'll provide a custom PanicHandler like so:
```go
package myerrorlib

import (
	"context"
)

type PanicHandler struct{}

// LogPanic is a no-op to silence the default panic logging.
// Implements the Logger interface from `graph-gophers/graphql-go/log` 
func (p *PanicHandler) LogPanic(ctx context.Context, value interface{}) {
	return
}

// MakePanicError logs and reports the error using the common error library.
// Implements the PanicHandler interface from `graph-gophers/graphql-go/errors` 
func (p *PanicHandler) MakePanicError(
	ctx context.Context,
	value interface{},
) *errors.QueryError {
	// create and report custom error for 3rd party logging
	err := createAndReportErrorf(ctx, "panic: %s", value)
	return &errors.QueryError{
		Message:    err.Error(),
		// Decorate error with status code, trace id etc...
		Extensions: err.Extensions(),
	}
}
```

### DIFF
--- a/errors/panic_handler.go
+++ b/errors/panic_handler.go
@@ -1,0 +1,18 @@
+package errors
+
+import (
+	"context"
+)
+
+// PanicHandler is the interface used to create custom panic errors that occur during query execution
+type PanicHandler interface {
+	MakePanicError(ctx context.Context, value interface{}) *QueryError
+}
+
+// DefaultPanicHandler is the default PanicHandler
+type DefaultPanicHandler struct{}
+
+// MakePanicError creates a new QueryError from a panic that occurred during execution
+func (h *DefaultPanicHandler) MakePanicError(ctx context.Context, value interface{}) *QueryError {
+	return Errorf("graphql: panic occurred: %v", value)
+}
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -20,24 +20,21 @@ import (
 
 type Request struct {
 	selected.Request
-	Limiter chan struct{}
-	Tracer  trace.Tracer
-	Logger  log.Logger
+	Limiter      chan struct{}
+	Tracer       trace.Tracer
+	Logger       log.Logger
+	PanicHandler errors.PanicHandler
 }
 
 func (r *Request) handlePanic(ctx context.Context) {
 	if value := recover(); value != nil {
 		r.Logger.LogPanic(ctx, value)
-		r.AddError(makePanicError(value))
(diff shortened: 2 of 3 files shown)


errors/panic_handler.go
```diff
@@ -0,0 +1,18 @@
+package errors
+
+import (
+	"context"
+)
+
+// PanicHandler is the interface used to create custom panic errors that occur during query execution
+type PanicHandler interface {
+	MakePanicError(ctx context.Context, value interface{}) *QueryError
+}
```
pavelnikolov: The interface needs to be declared where it is used, not where you declare implementations of that interface. Please, move the interface, where it is used.

internal/exec/exec.go
```diff
@@ -20,26 +20,23 @@ import (
 
 type Request struct {
 	selected.Request
-	Limiter chan struct{}
-	Tracer  trace.Tracer
-	Logger  log.Logger
+	Limiter      chan struct{}
+	Tracer       trace.Tracer
+	Logger       log.Logger
+	PanicHandler errors.PanicHandler
```
pavelnikolov: The interface needs to be declared in this file.

pavelnikolov:
Please, add unit tests.

errors/panic_handler.go
hampusohlsson: OK, I followed the existing pattern for `log.Logger` which has the interface defined in the same file as the default logger. But thanks for the pointer, will move the `PanicHandler` interface to the `exec.go` file.

pavelnikolov:
@hampusohlsson are you still interested in this change? The code seems to have conflicts. I'm closing this PR. Please, feel free to reopen it later.

PR #330, opened 2019-05-09
Fix Input Value Validation

dackroyd:
Validating Input values supplied via variable input are supplied with the correct input structure and values. This includes ensuring that enum embedded within input values are validated as well


dackroyd:
Related to #325

dackroyd:
Please keep this on hold - I've realised that nesting Input types don't validate correctly, which I'll look at fixing

internal/validation/testdata/tests.json
```diff
@@ -1752,17 +1873,7 @@
       "vars": {
         "colors": "BROWN"
       },
-      "errors": [
```
dackroyd: This case was actually an example where input coercion should have _allowed_ the value, so expecting an error here was actually incorrect.

internal/validation/validation.go
```diff
@@ -193,7 +193,8 @@ func validateValue(c *opContext, v *common.InputValue, val interface{}, t common
 		}
 		vv, ok := val.([]interface{})
 		if !ok {
-			c.addErr(v.Loc, "VariablesOfCorrectType", "Variable \"%s\" has invalid type %T.\nExpected type \"%s\", found %v.", v.Name.Name, val, t, val)
+			// Input coercion rules allow single items without wrapping array
```
dackroyd: This resolves the issue with input coercion for lists where a single value is supplied in variables, without a wrapping list

internal/validation/validation.go
```diff
@@ -214,6 +215,19 @@ func validateValue(c *opContext, v *common.InputValue, val interface{}, t common
 			}
 		}
 		c.addErr(v.Loc, "VariablesOfCorrectType", "Variable \"%s\" has invalid value %s.\nExpected type \"%s\", found %s.", v.Name.Name, e, t, e)
+	case *schema.InputObject:
+		if val == nil {
+			return
+		}
+		in, ok := val.(map[string]interface{})
+		if !ok {
+			c.addErr(v.Loc, "VariablesOfCorrectType", "Variable \"%s\" has invalid type %T.\nExpected type \"%s\", found %T.", v.Name.Name, val, t, val)
```
pavelnikolov: Why do you print the type (i.e. `%T`) twice? Did you mean to print the value?

dackroyd: I've updated this to print the value (quoted). Not sure what I was going for here with reprinting the `%T` value

dackroyd: Actually, removed the quoting, since this is not a string value

pavelnikolov:
Thank you!

marcoazn89:
Sorry if this is me not understanding Graphql but it seems that this line will trigger an error when the input is not there https://github.com/graph-gophers/graphql-go/pull/330/files#diff-bd40f10a5bef7a0945987430ac7762bdR223 which makes sense when the inputs are required. However, it should not do that if the input is optional and not present.

Let me know if this makes sense!

pavelnikolov:
@marcoazn89 thanks for reporting this. Do you have an example to reproduce an error?

PR #336, opened 2019-06-10
update opentracing-go version

gosp:
The latest opentracing-go didn't depend on "golang.org/x/net" any more.
let's update it.


tonyghita:
Thank you!

PR #338, opened 2019-06-25
Support for embedded struct type in resolver

### LINKED ISSUE #331: Support for embedded struct type in resolver
I would like my resolver to expand on an existing struct type, like from another package, when using `UseFieldResolvers`. To do that, I compose the objects by embedded the base struct inside the resolver struct. However, this doesn't compile because it looks like graphql-go only traverses the first level of fields to find the one that satisfies the schema. This would be helpful so I don't have to implement tedious resolver methods for every field.

```golang
package resolvers

import "github.com/gxb5443/project1/users"

type AdminResolver struct {
  users.Admin
}
```

eloyekunle:
Fixes #331 

Updated example in `./example/social/social.go` to demonstrate how it works.

Follow up to: #282 
Ref: #28

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -382,7 +382,7 @@ type Timestamps struct {
 	UpdatedAt string
 }
 
-func TestEmbedded(t *testing.T) {
+func TestEmbeddedStruct(t *testing.T) {
 	gqltesting.RunTests(t, []*gqltesting.Test{
 		{
 			Schema: graphql.MustParseSchema(`
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -1179,13 +1179,13 @@ func TestDeprecatedDirective(t *testing.T) {
 	})
 }
 
-type testBadEnumResolver struct {}
+type testBadEnumResolver struct{}
 
 func (r *testBadEnumResolver) Hero() *testBadEnumCharacterResolver {
 	return &testBadEnumCharacterResolver{}
 }
 
-type testBadEnumCharacterResolver struct {}
+type testBadEnumCharacterResolver struct{}
 
 func (r *testBadEnumCharacterResolver) Name() string {
 	return "Spock"
@@ -1227,7 +1227,7 @@ func TestEnums(t *testing.T) {
 			`,
 			ExpectedErrors: []*gqlerrors.QueryError{
 				{
-					Message: "Argument \"episode\" has invalid value WRATH_OF_KHAN.\nExpected type \"Episode\", found WRATH_OF_KHAN.",
+					Message:   "Argument \"episode\" has invalid value WRATH_OF_KHAN.\nExpected type \"Episode\", found WRATH_OF_KHAN.",
 					Locations: []gqlerrors.Location{{Column: 20, Line: 3}},
 					Rule:      "ArgumentsOfCorrectType",
 				},
@@ -1265,7 +1265,7 @@ func TestEnums(t *testing.T) {
 			Variables: map[string]interface{}{"episode": "FINAL_FRONTIER"},
 			ExpectedErrors: []*gqlerrors.QueryError{
 				{
-					Message: "Variable \"episode\" has invalid value FINAL_FRONTIER.\nExpected type \"Episode\", found FINAL_FRONTIER.",
+					Message:   "Variable \"episode\" has invalid value FINAL_FRONTIER.\nExpected type \"Episode\", found FINAL_FRONTIER.",
 					Locations: []gqlerrors.Location{{Column: 26, Line: 2}},
 					Rule:      "VariablesOfCorrectType",
 				},
@@ -1327,7 +1327,7 @@ func TestEnums(t *testing.T) {
 			ExpectedErrors: []*gqlerrors.QueryError{
 				{
 					Message: "Invalid value STAR_TREK.\nExpected type Episode, found STAR_TREK.",
-					Path:      []interface{}{"hero", "appearsIn", 0},
+					Path:    []interface{}{"hero", "appearsIn", 0},
 				},
 			},
 		},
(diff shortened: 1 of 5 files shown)


pavelnikolov:
What happens if there is more than one embedded struct with the same field?
This feature has been requested in the past by many people. There is one problem with it, though. Imagine we have more than one embedded struct containing fields with the same name. Then you'd end up with ambiguous code. For example: https://play.golang.org/p/0gvABy9DZ5Y
IMHO, this feature should not be merged because it creates edge cases and possible ambiguities.

eloyekunle:
@pavelnikolov What if we check for ambiguity while parsing the fields?

eloyekunle:
@pavelnikolov I have added a method, 'getFieldCount' that helps with checking for ambiguities so embedded structs containing fields with the same name will throw an error.

eloyekunle:
@pavelnikolov Please take another look.

internal/exec/resolvable/resolvable.go
```diff
@@ -228,13 +228,17 @@ func (b *execBuilder) makeObjectExec(typeName string, fields schema.FieldList, p
 
 	Fields := make(map[string]*Field)
 	rt := unwrapPtr(resolverType)
+	fieldsCount := getFieldCount(rt, map[string]int{})
 	for _, f := range fields {
-		fieldIndex := -1
+		var fieldIndex []int
```
pavelnikolov: Why does the field index is an array? I don't get it.

internal/exec/resolvable/resolvable.go
```diff
@@ -380,13 +384,46 @@ func findMethod(t reflect.Type, name string) int {
...
+		if strings.EqualFold(stripUnderscore(name), stripUnderscore(field.Name)) {
+			return append(index, i)
 		}
 	}
-	return -1
+
+	return index
+}
+
+// getFieldCount helps resolve ambiguity when more than one embedded struct contains fields with the same name.
```
pavelnikolov: When this ambiguity happens do you throw some kind of an error? I'd appreciate a unit tests that catches this case.

internal/exec/resolvable/resolvable.go
```diff
@@ -380,13 +384,46 @@ func findMethod(t reflect.Type, name string) int {
 	return -1
 }
 
-func findField(t reflect.Type, name string) int {
+func findField(t reflect.Type, name string, index []int) []int {
```
pavelnikolov: Why does this method need an integer array? `, index []int`?

internal/exec/resolvable/resolvable.go
```diff
@@ -228,13 +228,17 @@ func (b *execBuilder) makeObjectExec(typeName string, fields schema.FieldList, p
...
+	fieldsCount := getFieldCount(rt, map[string]int{})
 	for _, f := range fields {
-		fieldIndex := -1
+		var fieldIndex []int
 		methodIndex := findMethod(resolverType, f.Name)
 		if b.schema.UseFieldResolvers && methodIndex == -1 {
-			fieldIndex = findField(rt, f.Name)
+			if fieldsCount[strings.ToLower(stripUnderscore(f.Name))] > 1 {
+				return nil, fmt.Errorf("%s does not resolve %q: ambiguous field %q", resolverType, typeName, f.Name)
+			}
```
pavelnikolov: Is there a unit test about this error?

internal/exec/resolvable/resolvable.go
```diff
@@ -380,13 +384,46 @@ func findMethod(t reflect.Type, name string) int {
...
+func findField(t reflect.Type, name string, index []int) []int {
 	for i := 0; i < t.NumField(); i++ {
-		if strings.EqualFold(stripUnderscore(name), stripUnderscore(t.Field(i).Name)) {
-			return i
+		field := t.Field(i)
+
+		if field.Type.Kind() == reflect.Struct && field.Anonymous {
+			newIndex := findField(field.Type, name, []int{i})
+			if len(newIndex) > 1 {
+				return append(index, newIndex...)
```
pavelnikolov: I do not understand why we need to return an array.

internal/exec/resolvable/resolvable.go
eloyekunle: It is an array because it is an index sequence for [`Type.FieldByIndex`](https://golang.org/pkg/reflect/#Value.FieldByIndex).

It allows for several layers of anonymously-nested fields.

internal/exec/resolvable/resolvable.go
eloyekunle: This function builds the index sequence for [`Type.FieldByIndex`](https://golang.org/pkg/reflect/#Value.FieldByIndex) recursively, returning the sequence if found and an empty array otherwise.

internal/exec/resolvable/resolvable.go
eloyekunle: See https://github.com/graph-gophers/graphql-go/pull/338#discussion_r324255233 and https://github.com/graph-gophers/graphql-go/pull/338#discussion_r324257504.

internal/exec/resolvable/resolvable.go
eloyekunle: Added unit test [here](https://github.com/graph-gophers/graphql-go/pull/338/files#diff-d7906d668830f592334a2a80cf6d298cR3030).
Also removed `get` prefix.

internal/exec/resolvable/resolvable.go
eloyekunle: Yes, added [here](https://github.com/graph-gophers/graphql-go/pull/338/files#diff-d7906d668830f592334a2a80cf6d298cR3030).

eloyekunle:
I've also added a unit test for the feature here: https://github.com/graph-gophers/graphql-go/pull/338/files#diff-d7906d668830f592334a2a80cf6d298cR385

graphql_test.go
```diff
@@ -3018,23 +3100,65 @@ func TestErrorPropagation(t *testing.T) {
...
+type ambiguousResolver struct {
+	Name string // ambiguous
+	University
+}
+
+type University struct {
+	Name string // ambiguous
+}
+
+func TestPanicAmbiguity(t *testing.T) {
```
pavelnikolov: Can you, please, move `ambiguousResolver` and `University` inside of the test (e.i. `TestPanicAmbiguity`)?

graphql_test.go
```diff
@@ -351,6 +351,88 @@ func TestBasic(t *testing.T) {
...
+}
+
+type Instructor struct {
+	Name string
+}
+
+type Timestamps struct {
+	CreatedAt string
+	UpdatedAt string
+}
```
pavelnikolov: Same here. Can you, please, move the definition of these structs inside of the test `TestEmbeddedStruct`?

pavelnikolov:
Actually, I'll leave it like this for consistency. Refactoring should be in a separate PR.

PR #342, opened 2019-07-17
Fix bug: make Schema.Validate accept arguments

romshark:
Make `Schema.Validate` accept arguments and pass them over to `validation.Validate`
to allow parameterized queries to be validated correctly.

----

I'm having a problem with [`func (s *Schema) Validate(queryString string) []*errors.QueryError`](https://github.com/graph-gophers/graphql-go/blob/8f92f34fc59823d34fc08bfdc9fd266b854e2b50/graphql.go#L145). It doesn't take the _arguments_ as a parameter and thus fails to validate a parameterized query like this one:
```graphql
query($uid: ID!) {
  customer(id: $uid) {
    id
    firstName
    lastName
  }
}
```
```
graphql: Variable "uid" has invalid value null.
Expected type "ID!", found null. (line 1, column 8);
```

Without parameterization it validates just fine:
```graphql
query {
  customer(id: "foo") {
    id
    firstName
    lastName
  }
}
```

Now I'm wondering why [`func (s *Schema) Validate`](https://github.com/graph-gophers/graphql-go/blob/8f92f34fc59823d34fc08bfdc9fd266b854e2b50/graphql.go#L145) won't actually take `(queryString string, args map[string]interface{})` instead since [`func Validate`](https://github.com/graph-gophers/graphql-go/blob/8f92f34fc59823d34fc08bfdc9fd266b854e2b50/internal/validation/validation.go#L66) takes `nil` arguments right there: [`validation.Validate(s.schema, doc, nil, s.maxDepth)`](https://github.com/graph-gophers/graphql-go/blob/8f92f34fc59823d34fc08bfdc9fd266b854e2b50/graphql.go#L151).

I just tried passing the argument through and parameterized queries are validated correctly now. Is _not taking the args as parameter_ to [`func (s *Schema) Validate`](https://github.com/graph-gophers/graphql-go/blob/8f92f34fc59823d34fc08bfdc9fd266b854e2b50/graphql.go#L145) a deliberate design decision or just a misunderstanding that's fixed by this PR?

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -142,13 +142,13 @@ type Response struct {
 }
 
 // Validate validates the given query with the schema.
-func (s *Schema) Validate(queryString string) []*errors.QueryError {
+func (s *Schema) Validate(queryString string, args map[string]interface{}) []*errors.QueryError {
 	doc, qErr := query.Parse(queryString)
 	if qErr != nil {
 		return []*errors.QueryError{qErr}
 	}
 
-	return validation.Validate(s.schema, doc, nil, s.maxDepth)
+	return validation.Validate(s.schema, doc, args, s.maxDepth)
 }
 
 // Exec executes the given query with the schema's resolver. It panics if the schema was created


pavelnikolov:
@romshark thank you for your contribution. I am happy with your PR and I'd be happy to merge it. However, there is one problem with it - it introduces a breaking change. Ideally, I want to avoid introducing breaking changes.

Before:
```go
 func (s *Schema) Validate(queryString string) []*errors.QueryError {
```

(Potentially) After:

```go
type ValidationConfig struct {
    args map[string]interface{}
}

type ValidationOpt func(cfg *ValidationConfig)

func WithArgs(args map[string]interface{}) {
    return func(cfg *ValidationConfig) {
        cfg.args = args
    }
}

func (s *Schema) Validate(queryString string, opts ...ValidationOpt) []*errors.QueryError {
    var cfg ValidationConfig
    for _, o := range opts {
        o(cfg)
    }
	doc, qErr := query.Parse(queryString)
	if qErr != nil {
		return []*errors.QueryError{qErr}
    }
    return validation.Validate(s.schema, doc, cfg.args, s.maxDepth)
}
```

This way none of the existing clients of the public API would be affect. 

```go
// existing users
errs := s.Validate(query)

// users who would like to opt-in for this feature
errs := s.Validate(query, graphql.WithArgs(args))
```
In addition to that, please, include unit test coverage. Thank you.

romshark:
@pavelnikolov breaking the API is never a good option. However, `ValidationOpt` feels unnecessarily complex. I can't imagine what kind of validation options there'd be other than parameters.

Why not just provide a new simple function instead?
```
// existing users
errs := s.Validate(query)

// users who would like to opt-in for this feature
errs := s.ValidateWithArgs(query, args)
```

pavelnikolov:
[ValidateWithVariables](https://pkg.go.dev/github.com/graph-gophers/graphql-go#Schema.ValidateWithVariables) is implemented now.

PR #343, opened 2019-07-24
Query Caching Example

dackroyd:
Adding example app that allows responses to be cached based on hints
 added to the request context by resolvers. HTTP Cache-Control header
 is added using a customised version of the `relay` package HTTP handler
 implementation

### DIFF
--- a/example/caching/server/server.go
+++ b/example/caching/server/server.go
@@ -1,0 +1,139 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"net/http"
+
+	"github.com/graph-gophers/graphql-go"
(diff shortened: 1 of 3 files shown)


example/caching/server/server.go
```diff
@@ -0,0 +1,140 @@
...
+					try {
+						return JSON.parse(responseBody);
+					} catch (error) {
+						return responseBody;
+					}
+				});
+			}
+
+			ReactDOM.render(
+				// React.createElement(GraphiQL, {fetcher: graphQLFetcher}),
```
pavelnikolov: Why do you need this comment here?

dackroyd: Left this in by mistake. Have removed it now

Issue #348, opened 2019-08-27
graphql: panic occurred: reflect: call of reflect.Value.Field on chan Value


SogoCZE:
Hello, 

I've noticed that when you try to use subscriptions with basic HTTP requests that do not include
a `graphql-ws` subprotocol it resolves that subscription with `schema.exec` which ends up using a subscriptions resolver that returns chan and causes reflection panic. 

So I thought it would be better to prevent this panic and return a reasonable error message in case of a subscription request without `graphql-ws` subprotocol.

I accomplished that by adding a simple if statement to the exec method.

```go
if op.Type == "SUBSCRIPTION" {
    return &Response{Errors: []*errors.QueryError{errors.Errorf("graphql-ws protocol header is missing")}}
}
```

I'm new to this library so I would like to ask if I missed something or it's a real problem and I should create pull-request?

pavelnikolov:
Well, additional validation never hurts. PR with that validation + a unit test would be welcome.

PR #349, opened 2019-08-27
Validation for subscriptions in schema.Exec

SogoCZE:
I've added validation for subscriptions in `scheme.Exec` which is invalid. I also added unit test for it.

PR for https://github.com/graph-gophers/graphql-go/issues/348.


pavelnikolov:

PR #356, opened 2019-10-04
Strip Common Indentation from BlockString Descriptions

### LINKED ISSUE #355: BlockStringValues do not strip common indentation
Multiline string (Block String) support leaves common leading indentation on each line, resulting in unexpected indentation in documentation. This has impacts such as incorrectly formatting when rendered in CommonMark (like in GraphiQL), as subsequent lines after the first end up being treated as code blocks.

The spec calls out the need to strip this leading indentation: https://graphql.github.io/graphql-spec/June2018/#sec-String-Value

> Since block strings represent freeform text often used in indented positions, the string value semantics of a block string excludes uniform indentation and blank initial and trailing lines via BlockStringValue().

https://github.com/graphql/graphql-js implements this, and refers to that section of the spec: https://github.com/graphql/graphql-js/blob/master/src/language/blockString.js#L3:L34

I have verified this behaviour, comparing `graphql-js` and `graphql-go` for this basic schema:

```
  schema {
     query: Query
  }
  type Query {
    """
    First line of docs.

    Second line of docs.

        ... code example ...
        ... code example ...
        ... code example ...
    """
    hello: String!
  }
```

For the `hello` field `graphql-js`, the introspection query returns the `description` value as:

```
First line of docs.\n\nSecond line of docs.\n\n    ... code example ...\n    ... code example ...\n    ... code example ...
```

While `graphql-go` returns:

```
First line of docs.\n\n    Second line of docs.\n\n        ... code example ...\n        ... code example ...\n        ... code example ...
```

dackroyd:
Fixes #355

### DIFF
--- a/internal/common/blockstring.go
+++ b/internal/common/blockstring.go
@@ -1,0 +1,103 @@
+// MIT License
+//
+// Copyright (c) 2019 GraphQL Contributors
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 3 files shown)


internal/common/blockstring.go
```diff
@@ -0,0 +1,77 @@
+package common
```
dackroyd: This code is adapted from the graphql-js implementation, which is under the MIT license. This will need attribution as per that license, which I assume is preferred via a comment at the top of the file here?

pavelnikolov: @dackroyd, please add a comment at the top of the file that this was translated from the reference implementation and add a link to the original code. I think that this should be enough

dackroyd: I've added in the MIT License as a header to the file, (copied from https://github.com/graphql/graphql-js/blob/5eb7c4ded7ceb83ac742149cbe0dae07a8af9a30/LICENSE), along with a comment explaining that the implementation here is derived from that reference implementation.

PR #359, opened 2019-10-07
Expose graphql input

roopakv:
This PR simply exposes the input graphQL on the relay handler so that users' graphql methods' can use this for things such as logging.

Currently we are forced to double parse the body, once before we use the relay handler just so that we can log the operation name etc in our logs.

### DIFF
--- a/relay/relay.go
+++ b/relay/relay.go
@@ -43,22 +43,24 @@ func UnmarshalSpec(id graphql.ID, v interface{}) error {
 	return json.Unmarshal([]byte(s[i+1:]), v)
 }
 
+type RelayInput struct {
+	Query         string                 `json:"query"`
+	OperationName string                 `json:"operationName"`
+	Variables     map[string]interface{} `json:"variables"`
+}
+
 type Handler struct {
 	Schema *graphql.Schema
+	Input  RelayInput
 }
 
 func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	var params struct {
-		Query         string                 `json:"query"`
-		OperationName string                 `json:"operationName"`
-		Variables     map[string]interface{} `json:"variables"`
-	}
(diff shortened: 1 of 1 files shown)


relay/relay.go
```diff
@@ -43,22 +43,24 @@ func UnmarshalSpec(id graphql.ID, v interface{}) error {
...
 
+type RelayInput struct {
+	Query         string                 `json:"query"`
+	OperationName string                 `json:"operationName"`
+	Variables     map[string]interface{} `json:"variables"`
+}
+
 type Handler struct {
 	Schema *graphql.Schema
+	Input  RelayInput
```
pavelnikolov: This is not a good idea as `Input` gets reused between requests. As a matter of fact, this will 100% lead to concurrency bugs.

pavelnikolov:
This is not ok because it leads to concurrency bugs.

roopakv:
ahhh i missed that. is there a better easier way to expose the parsed informationto my handler?

pavelnikolov:
There should be a better way. But the questions is why do you even need that? Maybe you can log trace ID and then view the tracelog for interesting requests (in case they are sampled)...

roopakv:
We want to bump metrics based on our graphQL operations for a start. Yes a trace ID will solve some things, but it doesnt really give us everything

roopakv:
@pavelnikolov what do you think about putting the parsed operation inside context?

pavelnikolov:
No. The library needs to be agnostic and as minimal as possible. However, you as the owner of the context can put anything you want inside. You can also use a custom tracer, which also increases some counter...

roopakv:
I understand the minimalism attitude but now we will have double parse the body just to find the operation name. That feels very restrictive don't you think? I can't imagine we are the only ones who want to track these metrics

PR #361, opened 2019-10-09
Add Example of Custom Errors

### LINKED ISSUE #274: Add example of custom errors
Now that #272 has been merged it would be nice to have an example in the `errors` package which appears here https://godoc.org/github.com/graph-gophers/graphql-go/errors.
I think that input validation error(s) sounds like a good idea.

dackroyd:
Resolves #274


pavelnikolov:
Thank you, @dackroyd

Issue #362, opened 2019-10-28
Unclear error when resolver arguments don't meet schema definition


dackroyd:
There are many ways of mismatching the input arguments for a field between the resolver implementations and the GraphQL schema definition. When this mismatch occurs, the error returned for parsing the schema is difficult to follow. An example:

Schema:

```
schema {
	query: Query
}
type Query {
	hello(input: HelloInput!): String!
}
type HelloInput {
	name: String
}
```

Implementation 1: Missing 'args' wrapper struct

```
type HelloResolver struct{}
type HelloInput struct {
	Name string
}

func (r *HelloResolver) Hello(in HelloInput) string {
	return "Hello " + in.Name + "!"
}
```

Fails with:

```
missing argument {"name" {'\x06' '\r'}}
    			returned by (*mypkg.HelloResolver).Hello
```

Implementation 2: Missing `name` field from input struct

```
type HelloResolver struct{}
type HelloInput struct {
	World string
}

func (r *HelloResolver) Hello(args struct { Input HelloInput }) string {
	return "Hello " + args.Input.World + "!"
}
```

Fails with:

```
field "Input": missing argument {"name" {'\t' '\a'}}
    			returned by (*mypkg.HelloResolver).Hello
```

While these are not defining the implementation correctly as is documented, the error is not as clear as it could be. Some issues:

- The printed argument uses the `%q` formatting verb, which doesn't appear to be documented outside of use with `int` and `string` types (https://golang.org/pkg/fmt/). This verb used with `common.Ident` pulls in the 'location' values of the argument, however these are represented poorly by the formatting
- Error doesn't identify the (input) resolver struct that is declared incorrectly. It does identify the function (and receiver) that the struct is used by, but it's not obvious that the argument struct is the issue
- Error mentions "returned by", which is not the case for input arguments

dackroyd:
Similar case came up in #339

pavelnikolov:
Fixed by #363

PR #363, opened 2019-10-28
Clarify errors for mismatching input implementation

dackroyd:
Producing clearer error messages when field input arguments are
 implemented by code:

 * Which does not match the schema e.g. missing field; or
 * Function missing struct wrapper for field arguments


dackroyd:
Fixes #362

PR #364, opened 2019-10-30
Allow `schema` to be omitted when using default root op names

### LINKED ISSUE #304: Populating entry points without explicity schema{}
From https://facebook.github.io/graphql/June2018/#sec-Root-Operation-Types : 

> While any type can be the root operation type for a GraphQL operation, the type system definition language can omit the schema definition when the query, mutation, and subscription root types are named Query, Mutation, and Subscription respectively.
> 
> Likewise, when representing a GraphQL schema using the type system definition language, a schema definition should be omitted if it only uses the default root operation type names.
> 
> This example describes a valid complete GraphQL schema, despite not explicitly including a schema definition. The Query type is presumed to be the query root operation type of the schema.

If I understand this correctly, having a type named `type Query` with no `schema {}` should be the same as doing `type Query ... schema { query: Query }` but currently this does not work.

dackroyd:
Fixes #304

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -37,6 +37,9 @@ func ParseSchema(schemaString string, resolver interface{}, opts ...SchemaOpt) (
 	if err := s.schema.Parse(schemaString, s.useStringDescriptions); err != nil {
 		return nil, err
 	}
+	if err := s.validateSchema(); err != nil {
+		return nil, err
+	}
 
 	r, err := resolvable.ApplyResolver(s.schema, resolver)
 	if err != nil {
@@ -183,6 +186,11 @@ func (s *Schema) exec(ctx context.Context, queryString string, operationName str
 	if op.Type == query.Subscription {
 		return &Response{Errors: []*errors.QueryError{&errors.QueryError{ Message: "graphql-ws protocol header is missing" }}}
 	}
+	if op.Type == query.Mutation {
+		if _, ok := s.schema.EntryPoints["mutation"]; !ok {
+			return &Response{Errors: []*errors.QueryError{{ Message: "no mutations are offered by the schema" }}}
+		}
+	}
 
 	// Fill in variables with the defaults from the operation
 	if variables == nil {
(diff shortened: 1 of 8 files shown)


graphql_test.go
```diff
@@ -2512,15 +2742,19 @@ func (r *inputResolver) Nullable(args struct{ Value *int32 }) *int32 {
 	return args.Value
 }
 
-func (r *inputResolver) List(args struct{ Value []*struct{ V int32 } }) []int32 {
+func (r *inputResolver) List(args struct {
```
dackroyd: Crept in, will revert these unrelated changes

internal/exec/resolvable/resolvable.go
```diff
@@ -362,7 +362,8 @@ func (b *execBuilder) makeFieldExec(typeName string, f *schema.Field, m reflect.
 	var out reflect.Type
 	if methodIndex != -1 {
 		out = m.Type.Out(0)
-		if typeName == "Subscription" && out.Kind() == reflect.Chan {
+		sub, ok := b.schema.EntryPoints["subscription"]
```
dackroyd: This is an existing bug which prevents the `subscription` root op having a type named anything other than `Subscription`. Had to fix it here to verify correct behaviour for custom type names, but can split that work if desired

dackroyd:
Looking through open PRs on the repo, I notice that this has been raised previously under #126 back in 2017, but never merged. The codebase has drifted quite a lot since then, so it has picked up a few conflicts

graphql.go
```diff
@@ -223,6 +231,39 @@ func (s *Schema) exec(ctx context.Context, queryString string, operationName str
...
+
+func validateRootOp(s *schema.Schema, name string, mandatory bool) error {
+	t, ok := s.EntryPoints[name]
+	if !ok {
+		if mandatory {
+			return fmt.Errorf("root operation %q must be defined", name)
+		}
+		return nil
+	}
+	if t.Kind() != "OBJECT" {
```
dackroyd:
```
if _, ok := t.(*schema.Object); !ok {
```

^ could be a better way of checking this?

pavelnikolov: They are equivalent, IMHO

PR #370, opened 2019-12-17
Add support for directives in schema parser

Ackar:
This adds support for parsing directives in the schema.

It's probably of limited use for now but it's mostly so it stops crashing on schemas with directives, and it's a first step towards a better directive support.

### DIFF
--- a/internal/common/values.go
+++ b/internal/common/values.go
@@ -6,12 +6,13 @@ import (
 
 // http://facebook.github.io/graphql/draft/#InputValueDefinition
 type InputValue struct {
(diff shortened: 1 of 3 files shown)


tsholmes:
This would enable us to fully autogenerate the go interfaces to match the schema using directives for some configuration options. @tonyghita what needs to be done to get this merged?

tsholmes:
@pavelnikolov @tonyghita ping

pavelnikolov:
Thank you very much for your contribution.

PR #371, opened 2019-12-24
add support for apollo client

esemeniuc:
Tested working for 
React 16.12 
Apollo-Client 2.6.8
go 1.13


bsr203:
There are security implication allowing CORS from any origin. You can easily add CORS with https://github.com/rs/cors. Which is the best maintained library and just need to add as a middle ware. Please don't add to this library.

pavelnikolov:
I agree. This should not be part of the library. You can add a middleware in the examples folder, though.

PR #373, opened 2020-01-07
Expose query's selected fields to resolvers via context

### LINKED ISSUE #17: No way to get requested fields inside resolver

keithmattix:
Since #169 seems to have stagnated a bit, I decided to throw my hat in the ring. I'm not sure how everyone feels about exposing this information via context vs. a third parameter, but hopefully this can serve as a starting point for more discussion on the issue. Fixes #17

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,17 @@ func (r *helloWorldResolver) Hello(ctx context.Context) (string, error) {
 }
 ```
 
+### Schema Options
+
+- `UseStringDescriptions()` enables the usage of double quoted and triple quoted. When this is not enabled, comments are parsed as descriptions instead.
+- `UseFieldResolvers()` specifies whether to use struct field resolvers.
+- `MaxDepth(n int)` specifies the maximum field nesting depth in a query. The default is 0 which disables max depth checking.
+- `MaxParallelism(n int)` specifies the maximum number of resolvers per request allowed to run in parallel. The default is 10.
+- `Tracer(tracer trace.Tracer)` is used to trace queries and fields. It defaults to `trace.OpenTracingTracer`.
+- `ValidationTracer(tracer trace.ValidationTracer)` is used to trace validation errors. It defaults to `trace.NoopValidationTracer`.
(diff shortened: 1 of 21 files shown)


willglynn:
I would rather have field selections exposed as an argument (#169) than via `Context` (this PR). If my resolver needs field selections, it's because there's logic in my resolver which depends on those field selections, and I'd prefer to state that explicitly in my resolver's function signature. My resolvers' use of `Context` typically doesn't affect control flow, while field selection always does.

Having said that, I would rather have field selections exposed via `Context` than not exposed at all. This kind of functionality is important to me, and regardless of how it's exposed, merging either PR would allow me to stop using a fork.

dusansimic:
This feature is really useful to me and it'd be very nice if it gets pulled. I hope this will be done soon.

keithmattix:
Closing in favor of #428

PR #379, opened 2020-03-07
WIP Limit goroutines spawned to resolve fields

amzuko:
This is a first pass at limiting the number of goroutines that we spawn in executing field sets. On my machine, it seems to keep memory and cpu usage under control (although there's definitely many other things that we could do to improve that) for the ballet project with deep group-by's.


amzuko:
(sorry, meant to open this against our internal fork)

internal/exec/exec.go
```diff
@@ -73,18 +73,28 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 
 	var fields []*fieldToExec
 	collectFieldsToResolve(sels, s, resolver, &fields, make(map[string]*fieldToExec))
-
 	if async {
 		var wg sync.WaitGroup
-		wg.Add(len(fields))
-		for _, f := range fields {
-			go func(f *fieldToExec) {
+		const numberOfExecutors = 4
```
pavelnikolov: How did you land at 4. Is there a reason why this number isn't configurable?

pavelnikolov:
I wonder how much this would change the behaviour of the library. I would be interested to see benchmarks after you finish working on this.

amzuko:
Hey Pavel,

Thank you for your work on graphql-go!

I pulled 4 out of... somewhere the sun doesn’t shine. This PR was just to bring the change to the attention of some teammates of mine, and I opened it against your repo entirely by mistake. I think that this is far the blunt of a change as-is, especially to try to merge into master.

Anecdotally, and I’m happy to share more detailed benchmarks when we have them, this dramatically improves performance for graphql queries with large numbers (X-XX,000) of list elements (>80% reduction in memory usage). Significant latency improvements as well, apparently attributable to not needing to allocate and move more stack for tons of goroutines.

We know that the “right” solution is pagination of some sort at the application level, but that’s a pretty involved change.

The library seems to “work” with the change. If I understand the execution code correctly (and I definitely don’t completely understand it), this doesn’t provide a strict limit on the number of goroutines spawned, but simply limits the growth associated with lists and levels of object depth.

I am curious: why spawn so many goroutines in the first place? I know that they are “cheap”, but... most of our resolver methods are just returning a strict field. Maybe others are doing something else that might involve a blocking operation, but it’s sparse (in our codebase — maybe we are odd?). Seems heavy at first glance. When making this decision, did you perform any benchmarking or testing that led you down this path? Is there a pattern that we’re missing that encourages it? Forgive me if this is documented somewhere — I spent a bit of time looking, but again, don’t want to claim to be completely up to speed on graphql-go...

PR #382, opened 2020-04-03
readme section added `Testing clients/IDEs`

Nishchit14:


pavelnikolov:
Thank you for contributing but these links are not directly related to this library. I would prefer if we do not add them to the README.md file.

PR #384, opened 2020-04-26
Added Verbose Logger for logging query info during panic

abhif22:
Changes:
1. Added type `VerboseLogger` implements log.Logger. LogPanic method on VerboseLogger and logs the info passed on.
2. Modified the LogPanic method definition by adding field `info` to be logged.
3. Added LoggerInfo field on type `Request`. This loggerinfo will be used for verbose logging. Adding loggerinfo to each request.
4. Exposed function `VerboseLogger` type `SchemaOpt` to be used from outside graphql-go to change logging to verbose.

User can therefore pass-in the `VerboseLogger` `SchemaOpt` during parsing phase to set the logging to verbose effectively logging all the query info which led to panic. This can be really helpful to debug which query is causing problem in production systems by avoiding manually going through the stack trace or employing some hack.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -8,16 +8,27 @@ import (
 
 // Logger is the interface used to log panics that occur during query execution. It is settable via graphql.ParseSchema
 type Logger interface {
-	LogPanic(ctx context.Context, value interface{})
+	LogPanic(ctx context.Context, value interface{}, info interface{})
 }
 
 // DefaultLogger is the default logger used to log panics that occur during query execution
 type DefaultLogger struct{}
 
 // LogPanic is used to log recovered panic values that occur during query execution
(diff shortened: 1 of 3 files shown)


log/log.go
```diff
@@ -8,16 +8,27 @@ import (
 
 // Logger is the interface used to log panics that occur during query execution. It is settable via graphql.ParseSchema
 type Logger interface {
-	LogPanic(ctx context.Context, value interface{})
+	LogPanic(ctx context.Context, value interface{}, info interface{})
```
pavelnikolov, line 11: I don't want the Logger interface to add one more argument to this method. This is a breaking change of the API and I want to avoid it as much as possible.

Issue #386, opened 2020-05-10
Issue when running with go-fuzz


jveiga:
Hi, I was learning how to use go-fuzz with a project and tested with this project in commit `dae41bde9ef91c12e78863f0299348612f2d6214`.
When testing `query.Parse` I found a `program hanged (timeout 10 seconds)` , similar to #4,
with input
```
query($"\344\334\234\344\334\344\234d44\201"
```

PS: I also found 2 other issues but they panic with `Invalid UTF-8 encoding`, which I assume is the correct behaviour. 

I can try to look into debugging it with delve

pavelnikolov:
This is something that I've wanted to do for quite some time. Unit tests, PRs are more than welcome. Fuzzy testing is extremely important from security point of view.

jveiga:
From what I'm seeing in the Go ecosystem there are two ways to fuzzy test and they have a small overlap. One is to run fuzz unit tests with [gofuzz](https://github.com/google/gofuzz) and the other is to run with [go-fuzz](https://github.com/dvyukov/go-fuzz) (yes naming...).

Gofuzz works by writing unit test like tests, similar to
```go
func TestFuzzQueryNonASCII(t *testing.T) {
  t.Parallel()
  f := fuzz.New()
  var query string
  for i := 0; i < 100000; i++{
    f.Fuzz(&query)
    require.NotPanics(t, func() {
      Parse(query)
    }, "panicked with input %s", string(query))
  }
}
```
for each `go test` call, it will try to call `query.Parse` with random inputs for X amount of times.
I would suggest keeping crash input and create a test harness to avoid regressions.

Now Go-fuzz requires some more work as it requires packaging a callback function into a function  with some ceremony like
```go
package query

func Fuzz(data []byte) int {
	Parse(string(data))
	return 0
}
```
This also requires some thought as you can "direct" the fuzzer to test certain inputs.

There are also some SaaS that fuzz and support open source projects, I assure you that I do not represent or work for any of these, like https://fuzzit.dev/ and https://fuzzbuzz.io/. 

Now, I can create a PR with some gofuzz unit tests and a regression with the mentioned query and will also try to get a fix.

PR #392, opened 2020-05-17
adding fuzz tests 

jveiga:
referencing #386 
I added github actions to test them in CI but can remove them since the project uses semaphore.

Will see about fixing the `query($~\344\334\234\344\334\344\234�d44\201"` input

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,9 @@
 module github.com/graph-gophers/graphql-go
 
-require github.com/opentracing/opentracing-go v1.1.0
+require (
+	github.com/google/gofuzz v1.1.0
+	github.com/opentracing/opentracing-go v1.1.0
+	github.com/stretchr/testify v1.3.0
+)
 
 go 1.13
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,10 @@
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-16.04
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '^1.13'
+    - run: go test ./...
--- a/go.sum
+++ b/go.sum
@@ -1,2 +1,12 @@
+github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/google/gofuzz v1.1.0 h1:Hsa8mG0dQ46ij8Sl2AYJDUv1oA9/d6Vk+3LG99Oe02g=
+github.com/google/gofuzz v1.1.0/go.mod h1:dBl0BpW6vV/+mYPU4Po3pmUjxk6FQPldtuIdl/M65Eg=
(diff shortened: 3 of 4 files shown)


agnivade:
This is a great idea @jveiga ! Now that we have official fuzz support in Go 1.18, would you be willing to update the PR to use that? Of course, we would need to wait for a few weeks for it to be released, but in the meanwhile we can start migrating this PR.

Any thoughts?

pavelnikolov:
Yes, indeed the fuzz testing is an amazing idea. Looking forward to Go 1.18.
It would be ideal if we could generate more sophisticated queries based on a given schema.

jveiga:
> This is a great idea @jveiga ! Now that we have official fuzz support in Go 1.18, would you be willing to update the PR to use that? Of course, we would need to wait for a few weeks for it to be released, but in the meanwhile we can start migrating this PR.

That's a great approach. I would ask then which functions to fuzz.

> Yes, indeed the fuzz testing is an amazing idea. Looking forward to Go 1.18.
It would be ideal if we could generate more sophisticated queries based on a given schema.

This kind of fuzzing is easier to do with [go-fuzz](https://github.com/dvyukov/go-fuzz) which I can also add an integration with. I will see what kind of features go 1.18 will support and update this branch with findings

agnivade:
> That's a great approach. I would ask then which functions to fuzz.

I think `ParseSchema` is a good start. We can keep the resolver constant, and keep changing the `schemaString`. Later, we can even add `(*Schema).Exec`

pavelnikolov:
The real value would come from fuzzing the `Exec` because this is where user input comes during request time.

jveiga:
updated with `graphql_test.FuzzSchemaExec` 

Feel free to add more methods to the implementation of `query`

Also, I realised that calling fuzz stores a local folder with regressions like
```bash
❯ tree testdata
testdata
└── fuzz
    └── FuzzSchemaExec
        ├── 671f05b72e69d643e6570ec3777c27147cb4ec551090f2d71ce1c4db1010c686
        └── c843983602e04670a35adb41fb79674de6dd362f431acf11c4f8f6fe7a78df84
```

Should I add these folders to `.gitignore`?

go.mod
```diff
@@ -1,5 +1,8 @@
 module github.com/graph-gophers/graphql-go
 
-require github.com/opentracing/opentracing-go v1.1.0
+require (
+	github.com/opentracing/opentracing-go v1.1.0
+	github.com/stretchr/testify v1.3.0 // indirect
```
pavelnikolov, line 5: Could you, please, remove `testify`. It seems that it's not used anywhere.

internal/schema/schema_test.go
```diff
@@ -1007,3 +1007,10 @@ func TestInterfaceImplementsInterface(t *testing.T) {
 		})
 	}
 }
+
+func FuzzParse(f *testing.F){
+	f.Fuzz(func(t *testing.T, schemaString string, useStringDescriptions bool){
+		s := schema.New()
+		schema.Parse(s, schemaString, useStringDescriptions)
```
pavelnikolov, line 1014: When using string descriptions the syntax changes. So maybe we'll end up with a bunch of syntax errors which are expected. I wonder how beneficial is such a test. As in what are the type of errors that we could catch and prevent.

jveiga:
@pavelnikolov running random inputs helps finding unexpected bugs, out of bounds errors, logic errors, etc.
This PR started when I found an input that sent the parser into an infinite loop. That's one of the benefits of fuzzing these inputs. It is possible to add an initial corpus with [F.Add](https://pkg.go.dev/testing@master#F.Add), we can feed it basic query, mutation and schemas to have better inputs.

pavelnikolov:
I guess we'll have to wait for Go 1.18 to be released early in 2022 to merge this PR

PR #396, opened 2020-05-26
include introspection doocs

linkthrow:
Leaving some comments in the readme about how to generate the schema.json file, as took me a while to realise


pavelnikolov:
I don't think that the README.md is the correct place to put this comment. The method is already in godoc https://godoc.org/github.com/graph-gophers/graphql-go#Schema.ToJSON

PR #398, opened 2020-06-13
Fix SIGSEGV when client subs to multiple fields

nmaquet:
This fixes a SIGSEGV panic which can happen when a client attempts to subscribe to more than one field.


nmaquet:
@pavelnikolov Would you consider merging this? It's a simple fix and it's all tested and ready to go :)

Issue #375, opened 2020-01-31
Limit Goroutines Spawned in execSelections


shannontan-bolt:
We recently ran into a problem with large amounts of go routines being spawned and overwhelming the memory of the machine leading to a shutdown. After receiving tens of thousands of records from the DB, they were converted into resolvers, and then execSelections was called. execSelections appears to spawn a go routine for each field within this list, which means that the amount of goroutines will be proportional to the amount of fields in the list.

However, the execution is limited within execSelectionSet by the size of the limiter channel - so there's only 10 goroutines allowed to execute at any point in time. Instead of creating the goroutines and limiting the execution of them, it would be preferable to just limit the amount of goroutines being created and letting them run.

I agree that we should be limiting the records we retrieve from the DB, but this should manifest as slowness rather than system instability.

There are a few proposed solutions:
- Add the ability to limit the amount of fields in `execList`. This would be a SchemaOpt + a check that would add a runtime error if the list of the fields is too large. This would prevent too many goroutines from being spawned.
- Refactor the code (likely a significant refactor) such that each execSelectionSet/execSelection is run in a separate goroutine thereby allowing the parent goroutine to complete. This would require introducing something else to manage state since children goroutines would no longer be able to just update a local variable within the parent goroutine.
- This is very similar to the above idea, but with another change to make it breadth first: https://github.com/graph-gophers/graphql-go/pull/372/files

amzuko:
We have encountered this problem as well, when returning a substantial set of structured data in a single query. We are also interested in finding ways to control the number of goroutines spawned.

keegancsmith:
At Sourcegraph this has been a problem for us as well. We implemented a small change which helps to bound the number of goroutines, and it is working well for us. https://github.com/graph-gophers/graphql-go/pull/400

pavelnikolov:
Closing this issue since #400 takes care of it.

PR #400, opened 2020-06-25
Limit the number of concurrent list nodes processed

ryanslade:
It uses the current capacity of the limiter as a hint as this is set
based on the maxParallelism field on the schema.

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -312,20 +312,20 @@ func (r *Request) execList(ctx context.Context, sels []selected.Selection, typ *
 	l := resolver.Len()
 	entryouts := make([]bytes.Buffer, l)
 
-	sem := make(chan struct{}, cap(r.Limiter))
 	if selected.HasAsyncSel(sels) {
-		var wg sync.WaitGroup
-		wg.Add(l)
+		concurrency := cap(r.Limiter)
+		sem := make(chan struct{}, concurrency)
 		for i := 0; i < l; i++ {
 			sem <- struct{}{}
 			go func(i int) {
 				defer func() { <-sem }()
-				defer wg.Done()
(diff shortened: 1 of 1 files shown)


ryanslade:
This is a small targeted change intended to help with https://github.com/graph-gophers/graphql-go/issues/375

keegancsmith:
Note that the semaphore is created per `execList`. IE this change will make it such that you won't spin up more than `maxParallelism` per slice we exec on. So the children of execList are not at all bound to the same semaphore (if there are any child lists, they will create their own semaphore).

To be a little more mathematical this bounds the number of goroutines to `maxParallelism^depth(tree)`. This isn't exactly true since we aren't bounding execing on fields, but we can use the same idea to limit that as well.

We noticed this make a significant improvement on sourcegraph.com where we send back large result lists (search results).

shannontan-bolt:
AH! My mistake, I was weirdly caught in between the approaches. There's still a possibility of large fanout, but that's mostly dependent on parallelism vs depth. I see what you're saying, still seems like a major improvement.

Sourcegraph scale is also probably significant larger - at least in the graphical sense. I hope this PR makes it in :)

shannontan-bolt:
@ryanslade - should we ping some of the repo owners to help get this in?

ryanslade:
@pavelnikolov Pinging you as you were the last person to merge a PR 😄

pavelnikolov:
Hey @ryanslade 
I am happy to merge this. However, please add a comment why this is necessary in the code. It will not be immediately obvious in future when someone tries to modify this code.

ryanslade:
@pavelnikolov No problem, done.

pavelnikolov:
Thank you for the contribution.

Issue #324, opened 2019-04-27
Security - Implement query complexity analysis


fproulx-eoscanada:
Add something like https://github.com/pa-bru/graphql-cost-analysis to detect and protect against malicious resource hogging queries.

pavelnikolov:
PRs are welcome! I'd be happy if this is added somehow as an opt-in feature as it might not be applicable to all users of the library. The library needs to remain as minimalistic as possible.

romshark:
Query cost analysis is a really tough thing to get right and I thus prefer query whitelisting now. It may be a bit off-topic but I'd like to mention the [GraphQL Shield](https://github.com/romshark/dgraph_graphql_go/tree/master/api/gqlshield) middleware that I've introduced in [v1.3.0](https://github.com/romshark/dgraph_graphql_go/releases/tag/v1.3.0) of my [Dgraph + GraphQL + Go tech demo](https://github.com/romshark/dgraph_graphql_go).

The shield implements a dynamic radix-tree based query whitelist. It _normalizes_ incoming queries, checks if they're whitelisted and then checks if all expected arguments are present and valid. It doesn't parse the query and thus provides better DoS protection. It's basically a performance-for-convenience trade-off.

I think I'll move it to a separate repo later.

pavelnikolov:
@romshark I completely agree with you. Being able to add this as a middleware would be a better solution. This way only people who need it can add it.

mattsrobot:
@romshark do you have plans to create a middle-ware library of GraphQL Shield? I'm very interested!

pavelnikolov:
duplicate of #578

PR #401, opened 2020-06-26
Implement query complexity analysis #324

ne0h12:
#324

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -101,12 +101,12 @@ func MaxDepth(n int) SchemaOpt {
 	}
 }
 
-// MaxQueryComplexity specifies the complexity a query.
+// MaxQueryComplexity specifies the complexity of a query.
 func MaxQueryComplexity(n int) SchemaOpt {
 	return ComplexityEstimator(validation.SimpleEstimator{n})
 }
 
-// MaxQueryRecursion specifies the recursion a query.
+// MaxQueryRecursion specifies the recursion of a query.
 func MaxQueryRecursion(n int) SchemaOpt {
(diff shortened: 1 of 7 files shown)


ne0h12:
@pavelnikolov can you review my PR, please?

graphql.go
```diff
@@ -100,6 +101,23 @@ func MaxDepth(n int) SchemaOpt {
 	}
 }
 
+// MaxQueryComplexity specifies the complexity a query.
```
pavelnikolov, line 104: (nit) "... the complexity **of** a query."

ne0h12: I fixed. Thanks.

pavelnikolov:
Hi, @ne0h12 
Sorry for the late reply on this PR. Thank you for your effort but I don't feel that the complexity implementation should look like this. Instead after merging #560, I would like it to be possible for users to define their own `@complexity` directive on `FIELD_DEFINITION` and `ARGUMENT_DEFINITION`. This way each user can assign custom complexity and calculate it based on their needs. This library should not impose a particular solution. Instead it should be extendable and allow the users to define this behavior themselves.

internal/validation/complexity.go
```diff
@@ -0,0 +1,136 @@
...
+func (s visitedSels) copy() visitedSels {
+	newSels := visitedSels{}
+	for index, value := range s {
+		newSels[index] = value
+	}
+
+	return newSels
+}
+
+func (e RecursionEstimator) doRecursivelyVisitSelections(
```
pavelnikolov, line 86: How is this different than the normal `MaxDepth` check?

PR #407, opened 2020-08-05
Fixed union inline fragments

### LINKED ISSUE #241: Interface inline fragment on union types results in panic.
## Error
Given:
```
interface Node {
  id: ID!
}

type A implements Node {
  id: ID!
  a: String!
}

type B implements Node {
  id: ID!
  b: String!
}

type C implements Node {
  id: ID!
  c: String!
}

union ABC = A | B | C

query {
  search(input: InputType!): ABC
}
```
If I make the following query to this mock api:
```
{
  search(input: { ... }) {
    ... on A {
      a
    }
    ... on B {
      b
    }
    ... on C {
      c
    }
    ... on Node {
      id
    }
  }
}
```
I get a panic error that Node does not implement ABC, originating from the following line of code: https://github.com/graph-gophers/graphql-go/blob/bb9738501bd42a6536227b96068349b814379d6e/internal/exec/selected/selected.go#L170

## Expected
I expect the Node inline fragment to resolve just fine, because A) relay makes queries of this nature via its api, and B) each type in the union implements the Node interface.

## Probable source of the error
Looking through the code I can see - correct me if I'm wrong - that the type assertion is only checking for possible types for the union. It is not checking if A) the inline fragment is an interface type, or if B) the union objects implement said interface.

### Update
I've decided to just make all my union types into interface types for the time being.

pzduniak:
This fixes #241. I haven't fully understood the problem yet, so no tests are present. I adopted code from https://github.com/graph-gophers/graphql-go/issues/241#issuecomment-406877915, modifying it to support the scenario in that issue. I did not remove the original comparsion added by @marksauter, I don't know what was the scenario that broke his code.

I'm casting the "fragment type" as an interface, if it works then I look into its possible types and check if there's a TypeAssertion for that interface. I think that's OK? I'll write tests for this patch if you think this is fine.

### DIFF
--- a/internal/exec/selected/selected.go
+++ b/internal/exec/selected/selected.go
@@ -187,18 +187,23 @@ func applyFragment(r *Request, s *resolvable.Schema, e *resolvable.Object, frag 
 		}}
 	}
 	if ok && len(face.PossibleTypes) > 0 {
+		sels := []Selection{}
 		for _, t := range face.PossibleTypes {
 			if t.Name == e.Name {
 				return applySelectionSet(r, s, e, frag.Selections)
 			}
+
 			if a, ok := e.TypeAssertions[t.Name]; ok {
-				return []Selection{&TypeAssertion{
+				sels = append(sels, &TypeAssertion{
 					TypeAssertion: *a,
 					Sels:          applySelectionSet(r, s, a.TypeExec.(*resolvable.Object), frag.Selections),
-				}}
+				})
 			}
 		}
-		panic(fmt.Errorf("%q does not implement %q", e.Name, frag.On)) // TODO proper error handling
+		if len(sels) == 0 {
+			panic(fmt.Errorf("%q does not implement %q", e.Name, frag.On)) // TODO proper error handling
+		}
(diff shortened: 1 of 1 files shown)


pzduniak:
Looks like the CI failed, perhaps the implementation is too eager in applying the selection sets? I'll investigate this later.

pzduniak:
Code behaves OK, but the test is still failing somehow.

pzduniak:
Ahh, finally! It behaves OK now, I messed up the cast before. Still not entirely sure if this is the correct approach.

sqs:
I took this PR and added some test cases in https://github.com/graph-gophers/graphql-go/pull/409. Feel free to steal them and incorporate them into this branch if you'd prefer. Thanks for the fix!

pavelnikolov:
Thank you for your contribution. I'm closing this in favour of #409 because it has tests as well.

pzduniak:
Thank you for taking care of that guys, it's tough for me to find any time for finalizing my opensource contributions. Great to see my changes finally adopted!

PR #409, opened 2020-09-25
fix #241

### LINKED ISSUE #241: Interface inline fragment on union types results in panic.

sqs:
Similar to https://github.com/graph-gophers/graphql-go/pull/407, but adds test cases.

fix #241

### DIFF
--- a/example/social/introspect.json
+++ b/example/social/introspect.json
@@ -219,6 +219,39 @@
         "enumValues": null,
         "fields": [
           {
+            "args": [],
+            "deprecationReason": null,
+            "description": null,
+            "isDeprecated": false,
+            "name": "name",
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "String",
+                "ofType": null
+              }
+            }
+          }
+        ],
+        "inputFields": null,
+        "interfaces": null,
+        "kind": "INTERFACE",
+        "name": "Person",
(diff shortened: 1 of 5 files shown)


pavelnikolov:
I'm merging this PR because it has tests. I'll close the other one. Thank you for your effort.

PR #411, opened 2020-10-01
Add context to validation tracing

dackroyd:
Context is needed for tracing to access the current span, in order to
 add tags to it, or create child spans. As presently defined (without a
 context), this cannot be done: new spans could be created, but they
 would not be associated with the existing request trace.

OpenTracingTracer implements the new interface (it never implemented the
 old one). Via this 'extension interface', the tracer configured (or the
 default tracer) will be used as the validation tracer if:

 * The tracer implements the (optional) interface; and
 * A validation tracer isn't provided using the deprecated option

What this means is that the deprecated option is _preferred_ as an
 override. This allows users to migrate in a non-breaking, non-behaviour
 changing way, until such time as they intentionally remove the use of
 the deprecated option. For those who are currently using the default
 tracer, and not supplying a validation tracer, validation will be traced
 immediately with no change required to configuration options.

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -25,14 +25,21 @@ import (
 // resolver, then the schema can not be executed, but it may be inspected (e.g. with ToJSON).
 func ParseSchema(schemaString string, resolver interface{}, opts ...SchemaOpt) (*Schema, error) {
 	s := &Schema{
-		schema:           schema.New(),
-		maxParallelism:   10,
-		tracer:           trace.OpenTracingTracer{},
-		validationTracer: trace.NoopValidationTracer{},
-		logger:           &log.DefaultLogger{},
+		schema:         schema.New(),
+		maxParallelism: 10,
+		tracer:         trace.OpenTracingTracer{},
+		logger:         &log.DefaultLogger{},
 	}
 	for _, opt := range opts {
 		opt(s)
+	}
+
+	if s.validationTracer == nil {
+		if tracer, ok := s.tracer.(trace.ValidationTracerContext); ok {
+			s.validationTracer = tracer
+		} else {
+			s.validationTracer = &validationBridgingTracer{tracer: trace.NoopValidationTracer{}}
+		}
 	}
 
 	if err := s.schema.Parse(schemaString, s.useStringDescriptions); err != nil {
@@ -68,7 +75,7 @@ type Schema struct {
 	maxDepth                 int
 	maxParallelism           int
 	tracer                   trace.Tracer
-	validationTracer         trace.ValidationTracer
+	validationTracer         trace.ValidationTracerContext
 	logger                   log.Logger
 	useStringDescriptions    bool
 	disableIntrospection     bool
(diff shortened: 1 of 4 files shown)


dackroyd:
Open to adjusting the implementation here. As I've currently implemented it, I'm trying to avoid making this a breaking change for anyone that _is_ actually using the validation tracer. As mentioned however, it's impossible for tracing to work properly without supplying the existing request context containing the current span, so it may be ok for this to be a 'breaking' change instead of trying to maintain backward compatibility...

The `ValidationTracerContext` interface _does not_ return a `context.Context` in its current form, as the validation itself doesn't accept a context. I'm unsure whether maybe it _should_ take a context in the future, in which case the `ValidationTracerContext` should make sure that it returns the context it creates so that it can be used with that call.

pavelnikolov:
@dackroyd There are merge conflicts. It seems that I have merged one of the other PRs first. Would you mind resolving these. I'm sorry for the inconvenience.

dackroyd:
@pavelnikolov conflicts have been addressed

graphql.go
```diff
@@ -117,9 +124,10 @@ func Tracer(tracer trace.Tracer) SchemaOpt {
 }
 
 // ValidationTracer is used to trace validation errors. It defaults to trace.NoopValidationTracer.
+// Deprecated: context is needed to support tracing correctly. Use a Tracer which implements trace.ValidationTracerContext.
```
pavelnikolov, line 127: This is perfect with the comment here.

trace/trace.go
```diff
@@ -67,6 +67,22 @@ func (OpenTracingTracer) TraceField(ctx context.Context, label, typeName, fieldN
...
+		if len(errs) > 0 {
+			msg := errs[0].Error()
+			if len(errs) > 1 {
+				msg += fmt.Sprintf(" (and %d more errors)", len(errs)-1)
+			}
+			ext.Error.Set(span, true)
+			span.SetTag("graphql.error", msg)
+		}
+		span.Finish()
+	}
```
pavelnikolov, line 83: Isn't this code duplicating the errors logged above? https://github.com/graph-gophers/graphql-go/pull/411/files#diff-87aab52b1f88b8febc7f0d09f51af16d5bfafb1763ddfc2e353ae47d8425de9cR36-R46

trace/trace.go
```diff
@@ -67,6 +67,22 @@ func (OpenTracingTracer) TraceField(ctx context.Context, label, typeName, fieldN
 	}
 }
 
+func (OpenTracingTracer) TraceValidation(ctx context.Context) TraceValidationFinishFunc {
```
pavelnikolov, line 70: What is calling this new method? I can't see where the method is being called.

pavelnikolov: Nevermind, I just noticed it.

graphql.go
```diff
@@ -186,7 +194,7 @@ func (s *Schema) exec(ctx context.Context, queryString string, operationName str
 		return &Response{Errors: []*errors.QueryError{qErr}}
 	}
 
-	validationFinish := s.validationTracer.TraceValidation()
```
pavelnikolov, line 189: I wonder how was this used before without a context 🤷

PR #413, opened 2020-10-12
Update logic to always check for nil pointer returns

OuranosSkia:
Update the selection set execution logic to always check for a nil value. This addresses an issue where further down the pointer can be dereferenced without validating it is not nil and will cause a panic.

A test has been added to validate that the fix is functioning properly.


tonyghita (approved):
Looks good, thank you for covering the changes with a unit test 💯

PR #415, opened 2020-10-17
Adding variables parameter for query validations.

smotavitaendava:
Adding an extra interface for the internal validation.Validate for validate queries with GraphQL variables. Refactoring the existing `schema.Validate` function as well.

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -146,12 +146,17 @@ type Response struct {
 
 // Validate validates the given query with the schema.
 func (s *Schema) Validate(queryString string) []*errors.QueryError {
+	return s.ValidateWithVariables(queryString, nil)
+}
+
+// ValidateWithVariables validates the given query with the schema and the input variables.
+func (s *Schema) ValidateWithVariables(queryString string, variables map[string]interface{}) []*errors.QueryError {
 	doc, qErr := query.Parse(queryString)
 	if qErr != nil {
 		return []*errors.QueryError{qErr}
 	}
 
-	return validation.Validate(s.schema, doc, nil, s.maxDepth)
+	return validation.Validate(s.schema, doc, variables, s.maxDepth)
 }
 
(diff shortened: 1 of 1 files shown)


graphql.go
```diff
@@ -144,14 +144,67 @@ type Response struct {
...
+		return []*errors.QueryError{errors.Errorf("%s", err)}, nil
+	}
+
+	if op.Type == query.Mutation {
+		if _, ok := s.schema.EntryPoints["mutation"]; !ok {
+			return []*errors.QueryError{{Message: "no mutations are offered by the schema"}}, nil
+		}
+	}
+
+	fields := []string{}
```
pavelnikolov, line 168: prefer default value, e.g. `var fields []string`

graphql.go
```diff
@@ -144,14 +144,67 @@ type Response struct {
...
+	mainField := op.Selections[0].(query.Field)
+	for _, arg := range mainField.Arguments {
+		fieldName := arg.Name.Name + "."
+		fields = append(fields, fieldsFromLiteral(fieldName, arg.Value)...)
+	}
+
+	return nil, nil
```
pavelnikolov, line 175: You don't seem to be using the `fields` slice. Am I missing something? Why do you return `nil, nil`?

graphql.go
```diff
@@ -144,14 +144,67 @@ type Response struct {
...
+}
+
+func fieldsFromLiteral(prefix string, lit common.Literal) []string {
+	fields := []string{}
```
pavelnikolov, line 179: `var fields []string`

pavelnikolov:
Can you, please, add some unit test?

smotavitaendava:
@pavelnikolov Thanks for the time you took reading this PR. Currently I'm working on this, I thought I've set the latest change (in the 'Adding field discovery' commit) out of this PR.
I'm looking forward to have 'Adding variables parameter for query validations.' revised.
Sorry any inconvenience.

PR #417, opened 2020-11-11
Fixed `reflect.Value.Type on zero Value` panic when subscription resolver itself panics

maoueh:
The internal exec Subscribe method had code to deal with subscription resolver panicking when being called. But when such handling happen, the error is attached to the request object and it never checked later on.

This leads to some zero checks to fail when we try to extract the type from the resolver's channel since this variable was never set. Doing this creates a second panic which is not handled and make the application die.

To fix the issue, we now check if there is errors on the request object before continuing with the rest of the check, if there is errors, it's because a panic occurs and we return the response right away.

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -37,7 +37,7 @@ type extensionser interface {
 }
 
 func makePanicError(value interface{}) *errors.QueryError {
-	return errors.Errorf("graphql: panic occurred: %v", value)
+	return errors.Errorf("panic occurred: %v", value)
 }
 
 func (r *Request) Execute(ctx context.Context, s *resolvable.Schema, op *query.Operation) ([]byte, []*errors.QueryError) {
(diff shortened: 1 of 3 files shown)


internal/exec/exec.go
```diff
@@ -37,7 +37,7 @@ type extensionser interface {
 }
 
 func makePanicError(value interface{}) *errors.QueryError {
-	return errors.Errorf("graphql: panic occurred: %v", value)
```
maoueh, line 40: This was causing panic errors to print like `graphql: graphql: ...` so I removed it. Not other tests failed, so I assume it was never checked.

Can revert is asked for it no problem.

pavelnikolov:
@maoueh there seems to be a conflict in this PR. Can you, please, fix it.

maoueh:
Conflicts fixed

pavelnikolov:
Thank you

PR #317, opened 2019-04-10
Allow propagation of errors from Subscriptions channels into Request.…

billettc:
Please see open issue: 
https://github.com/graph-gophers/graphql-go/issues/314

PR #418, opened 2020-11-11
Added possibility to customize subscription resolver timeout value

maoueh:
The previous value was hard-coded to 1 second. This is problematic for resolver that takes more time than this to return a result.

When parsing the schema, it's not possible to pass a custom value for the subscription resolver timeout.

Extracted from #317


pavelnikolov:
@maoueh Thank you for the PR

Issue #145, opened 2017-12-15
Separate resolvers for queries and mutations


nicksnyder:
It appears that the schemaResolver passed in to ParseSchema must contain all root resolvers for both queries and mutations. Since these are separate namespaces in the graphql interface, it would be great to have them be separate namespaces (resolver objects) in the Go implementation too.

Consider the following schema:

```graphql
schema {
  query: Query
  mutation: Mutation
}

type Query {
    something: Something
}

type Mutation {
    updateSomething(value: String): Something
}

type Something {
    value: String
}
```

This currently requires an implementation like this:
```go
func init() {
    graphql.MustParseSchema(schemaFromSnippetAbove, &schemaResolver{})
}

type schemaResolver struct{}

func (r *schemaResolver) Something() *somethingResolver {
	return &somethingResolver{}
}

func (r *schemaResolver) UpdateSomething(ctx context.Context, args *struct {
	Value string
}) *somethingResolver {
	return &somethingResolver{value: args.Value}
}

type somethingResolver struct {
	value string
}

func (r *somethingResolver) Value() string {
	return r.value
}
```

But I would prefer it to require something like this:
```go
func init() {
    graphql.MustParseSchema(schemaFromSnippetAbove, &schemaResolver{})
}

type schemaResolver struct{}

func (r *schemaResolver) Query() *queryResolver {
	return &queryResolver{}
}

func (r *schemaResolver) Mutation() *mutationResolver {
	return &mutationResolver{}
}

type queryResolver struct{}

func (r *queryResolver) Something() *somethingResolver {
	return &somethingResolver{}
}

type mutationResolver struct{}

func (r *mutationResolver) UpdateSomething(ctx context.Context, args *struct {
	Value string
}) *somethingResolver {
	return &somethingResolver{value: args.Value}
}

type somethingResolver struct {
	value string
}

func (r *somethingResolver) Value() string {
	return r.value
}
```

This would better model what is happening at the graphql layer, which would make the server code more understandable, and also avoid unnecessary limitations (e.g. GraphQL allows you to have a mutation and a query with the same name, but graphql-go does not).

bsr203:
https://github.com/neelance/graphql-go/pull/142
https://github.com/neelance/graphql-go/issues/128
:-)

nicksnyder:
#142 is not the right solution in my option. I think there should still be a single top level resolver and that resolver should explicitly model the query resolver and the mutation resolver, as described above.

[...]

PR #421, opened 2020-11-11
Feature/prefix roots

maoueh:
This enables Query, Mutation and Subscription to use the same name. This is backward compatible and activated through an option when parsing the schema.

By specifying `graphql.PrefixRootFunctions()` option (name up to discussion), a single `RootResolver` object can be used that implements three time the same name for different kind of operation (`Query`, `Mutation`, `Subscription`). 

This was extracted from #317 and it's a variant implementation of #145 (I don't think we realized such work already existed prior doing our own fork, sorry for stepping on other toes).

#### Example

```graphql
schema {
    query: Query
    mutation: Mutation
    subscription: Subscription
}

type Subscription { hello: String! }
type Query { hello: String! }
type Mutation { hello: String! }
```

And the resolver and schema instantiation:

```go
schema := graphql.MustParseSchema(document, &RootResolver{}, graphql.PrefixRootFunctions())

type RootResolver struct{}

func (r *RootResolver) QueryHello() string {
	return "Hello query!"
}

func (r *RootResolver) MutationHello() string {
	return "Hello mutation!"
}

func (sr *RootResolver) SubscriptionHello(context.Context) (chan string, error) {
	c := make(chan string)
	go func() {
		c <- "Hello subscription!"
		close(c)
	}()

	return c, nil
}
```

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -41,7 +41,7 @@ func ParseSchema(schemaString string, resolver interface{}, opts ...SchemaOpt) (
 		return nil, err
 	}
 
-	r, err := resolvable.ApplyResolver(s.schema, resolver)
+	r, err := resolvable.ApplyResolver(s.schema, resolver, s.prefixRootFunctions)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pavelnikolov:
Closing this in favor of https://github.com/graph-gophers/graphql-go/pull/561

PR #422, opened 2020-11-13
Added possibility to explore the actual field selection tree

maoueh:
Added possibility to explore the actual field selection tree, this is attached to the context that the resolver method receive. I'm opening this to gather feedback about this new feature.

Each resolver's context will now be able to access through the graph of elements that were selected as part of the execution of this resolve via the standard go context's value.

The returned field is an interface representing a subset of the functionality of the internal/exec/selected package, enough for the developer to know what was queried by the user.

I'm opening the PR to start a discussion about offering this feature to end user of the library. We have started using it in our own fork of the library.

##### Design

This introduces new API interfaces the would become part of the standard library.

The interface is currently a simple wrapper around elements of `internal/exec/selected/Selection` interface needed to "walk" the selected field. Hopefully, I got the structure right.

I've currently used `Identifier()` instead of `Name()` and `Aliased()` instead of `Alias` to not renamed the field's name since there would be conflict with struct fields in the `internal/exec/selected/Selection` package. If we agree, the internal struct could change so the interface has the best name.

A future addition that I forsee could be:
- A "path" retrieval syntax "a la" gjson to quickly extract some nodes of the selection, for example with the star wars element ".search[]" that would return all type assertion nodes below the `search` field. Exact semantic to be discussed.
- Maybe a visitor pattern to ease the walking of the selection tree.

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,0 +1,26 @@
+package graphql
+
+import (
+	"context"
+
+	gcontext "github.com/graph-gophers/graphql-go/internal/context"
+	"github.com/graph-gophers/graphql-go/selected"
+)
+
+type Context struct {
+	Field selected.Field
+}
+
+// GraphQLContext is used to retrieved the graphql from the context. If no graphql
+// is present in the context, the `fallbackGraphql` received in parameter
+// is returned instead.
+func GraphQLContext(ctx context.Context) *Context {
+	field, found := gcontext.GraphQL(ctx)
+	if !found {
+		return nil
+	}
+
+	return &Context{
+		Field: field.ToSelection().(selected.Field),
+	}
+}
--- a/example/starwars/starwars.go
+++ b/example/starwars/starwars.go
@@ -4,12 +4,14 @@
 package starwars
 
 import (
+	"context"
 	"encoding/base64"
 	"fmt"
 	"strconv"
 	"strings"
 
 	graphql "github.com/graph-gophers/graphql-go"
+	"github.com/graph-gophers/graphql-go/selected"
 )
 
 var Schema = `
@@ -93,6 +95,8 @@ var Schema = `
 		appearsIn: [Episode!]!
 		# This droid's primary function
 		primaryFunction: String
+
+
 	}
 	# A connection object for a character's friends
 	type FriendsConnection {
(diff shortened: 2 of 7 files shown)


context.go
```diff
@@ -0,0 +1,26 @@
...
+
+	gcontext "github.com/graph-gophers/graphql-go/internal/context"
+	"github.com/graph-gophers/graphql-go/selected"
+)
+
+type Context struct {
+	Field selected.Field
+}
+
+// GraphQLContext is used to retrieved the graphql from the context. If no graphql
```
maoueh, line 14: Wrong comment

example/starwars/starwars.go
```diff
@@ -4,12 +4,14 @@
 package starwars
 
 import (
+	"context"
```
maoueh, line 7: Changes here are just for demonstration purposes if someone wants to play with the feature a bit. To try it out, run the starwars example server and perform this `curl` call:

```
curl 'http://localhost:8080/query' -d '{"query":"query {search(text:\"C-3PO\") {... on Droid {name} ... on Starship {name}}}","variables":null}'
```

maoueh: I do not intend to ship that if the feature is merged, it will be removed.

selected/interface.go
```diff
@@ -0,0 +1,78 @@
...
+	Children() []Selection
+}
+
+type TypenameField interface {
+	Selection
+	Type() string
+	Aliased() string
+}
+
+func Dump(selection Selection) {
```
maoueh, line 51: This is also present mainly for debugging/exploration purposes, another method that could be removed.

maoueh:
The use case we have right now is to explore the type assertion provided by the user, like in the starwars search query:

```
query {
  search(text:\"C-3PO\") {
    ... on Droid { name } 
    ... on Starship { name }
  }
}
```

We use the elements as a "filter" in our backend so we do not search for everything, only for what was requested by the user.

internal/exec/selected/selected.go
```diff
@@ -56,16 +70,86 @@ type SchemaField struct {
 	FixedResult reflect.Value
 }
 
+func (f *SchemaField) Kind() selected.Kind {
+	return selected.FieldKind
+}
+
+func (f *SchemaField) Identifier() string {
```
maoueh, line 77: Couldn't use `Name()` method here yet since the backing `internal/exec/selected/*SchemaField` already has a field `Name` so it conflicts. 

At one point, if we all agree on the feature and the interface, I suggest the internal field changes to the interface looks better.

This is the same reasoning for `Aliased` below.

pavelnikolov: Why can't you use its own `Name` field then?

maoueh: Because the new public interface would exposes it, hence a field with the same name cannot exist.

PR #423, opened 2020-11-17
DisableIntrospection should skip __typename

Epsirom:
`__typename` is defined in GraphQL spec [4.4](http://spec.graphql.org/June2018/#sec-Type-Name-Introspection), and when we use [union types](https://graphql.org/learn/schema/#union-types), the `__typename` field is usually used for differentiating different data types from each other on the client.

It is safe to expose the `__typename` field when `DisableIntrospection` because attackers can not get the type fields with `__typename`.

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -2614,6 +2614,44 @@ func TestIntrospectionDisableIntrospection(t *testing.T) {
 				}
 			`,
 		},
+
+		{
+			Schema: starwarsSchemaNoIntrospection,
+			Query: `
+				{
+					search(text: "an") {
+						__typename
+						... on Human {
+							name
+						}
+						... on Droid {
+							name
+						}
+						... on Starship {
+							name
+						}
+					}
+				}
+			`,
+			ExpectedResult: `
+				{
+					"search": [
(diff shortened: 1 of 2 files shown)


Epsirom:
In the official implementation [graphql-js](https://github.com/graphql/graphql-js), when introspections are disabled, the `__typename` field is still valid, see https://github.com/graphql/graphql-js/blob/master/src/validation/__tests__/NoSchemaIntrospectionCustomRule-test.js#L33

```javascript
describe('Validate: Prohibit introspection queries', () => {
  it('ignores valid fields including __typename', () => {
    expectValid(`
      {
        someQuery {
          __typename
          someField
        }
      }
    `);
  });
```

jaylucas:
Is this change merged in? I would like to have this as well.

kainosnoema:
Just ran into this... also needing to have `__typename` available while disabling the rest of schema introspect. Apollo client uses `__typename` as part of it core caching logic, so it's considered a required property: https://github.com/apollographql/apollo-client/issues/1913

pavelnikolov:

PR #425, opened 2020-12-03
Adding modifying field extraction

smotavitaendava:
Adding a method for extract the fields that are being modifying in a mutation. The idea is to have a list of fields where each field is identified by the path (type.field). e.g.

Having as schema and query:
```
...
input PersonInput {
    firstName: String!,
    middleName: String,
    lastName: String!
    location: LocationInput
}

input LocationInput {
    name: String!,
    description: String
}
...

mutation addingPerson($name: String!) {
    addPerson(person: {
        firstName: $name,
        middleName: null,
        lastName: "Fixed Last Name",
        location: {
            name: "Fixed Location Name"
        }
    }) { id }
}
```

The list of fields that are being modified are:
`person.firstName`, `person.middleName`, `person.lastName`, `person.location.name`

Note that `person.location.description` is not provided, hence, it won't be listed.

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -183,9 +183,17 @@ func (s *Schema) ModifyingFields(queryString string, operationName string, varia
 	}
 
 	var fields []string
-	mainField := op.Selections[0].(*query.Field)
-	for _, arg := range mainField.Arguments {
-		fields = append(fields, fieldsFromLiteral(arg.Name.Name, arg.Value)...)
+	for _, selection := range op.Selections {
+		mainField, ok := selection.(*query.Field)
+
+		// not something we could process
+		if !ok {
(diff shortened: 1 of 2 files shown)


pavelnikolov:
@smotavitaendava thank you for your PR, however, I don't think this feature is ready to be merged the way it is right now. I'm going to close this PR and will consider working on this when I work on the selected fields. In any case, it would be better off to start with opening an issue where we can discuss the actual implementation before proceeding with a PR. The functionality of the current PR could be implemented by using [nullable types](https://github.com/graph-gophers/graphql-go/blob/master/nullable_types.go) where you check if the field has been provided and if yes, then track that field in an array.

PR #428, opened 2021-01-02
Expose query's selected fields to resolvers via context

keithmattix:
I submitted #373 a while back, but I will be closing it in favor of this PR which is off of my fork's master branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,17 @@ func (r *helloWorldResolver) Hello(ctx context.Context) (string, error) {
 }
 ```
 
+### Schema Options
+
+- `UseStringDescriptions()` enables the usage of double quoted and triple quoted. When this is not enabled, comments are parsed as descriptions instead.
+- `UseFieldResolvers()` specifies whether to use struct field resolvers.
+- `MaxDepth(n int)` specifies the maximum field nesting depth in a query. The default is 0 which disables max depth checking.
+- `MaxParallelism(n int)` specifies the maximum number of resolvers per request allowed to run in parallel. The default is 10.
+- `Tracer(tracer trace.Tracer)` is used to trace queries and fields. It defaults to `trace.OpenTracingTracer`.
+- `ValidationTracer(tracer trace.ValidationTracer)` is used to trace validation errors. It defaults to `trace.NoopValidationTracer`.
+- `Logger(logger log.Logger)` is used to log panics during query execution. It defaults to `exec.DefaultLogger`.
+- `DisableIntrospection()` disables introspection queries.
+
 ### Custom Errors
 
 Errors returned by resolvers can include custom extensions by implementing the `ResolverError` interface:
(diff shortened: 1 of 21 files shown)


aarondl0:
Thank you for this. I built on this work to also expose the mutation arguments. This is really useful for mutations in Go with this library since you can't tell what fields are `null` vs completely omitted when binding to a struct (and map binding isn't supported by this library).

https://github.com/ProlificLabs/graphql-go/pull/1

pavelnikolov:
@aarondl0 there is a different approach to that in #430

README.md
```diff
@@ -76,7 +76,7 @@ When using `UseFieldResolvers` schema option, a struct field will be used *only*
 
 The method has up to two arguments:
 
-- Optional `context.Context` argument.
+- Optional `context.Context` argument. If the graphql query had nested subfields, then use the `SelctedFieldsFromContext(ctx context.Context)` getter method
```
mitchlloyd, line 79: Typo here: Selcted -> Selected

ventsislav-georgiev:
@pavelnikolov can one of these PRs get merged? Any approach of implementing the resolver's access to the requested fields will be sufficient for the users of the lib. It is a relatively small change and there are 3 open PRs for it. The first one dates back to 2017.

mehran-prs:
Hi all, 

This PR is great, I have been checked graphql documents and implementation in js and I was thinking about a third param(instead of context) that provides info to the resolver.
Just like what [graphql-js](https://graphql.org/graphql-js/type/#graphqlobjecttype) does:

```typescript
// See below about resolver functions.
type GraphQLFieldResolveFn = (
  source?: any,
  args?: {[argName: string]: any},
  context?: any,
  info?: GraphQLResolveInfo
) => any
 
type GraphQLResolveInfo = {
  fieldName: string,
  fieldNodes: Array<Field>,
  returnType: GraphQLOutputType,
  parentType: GraphQLCompositeType,
  schema: GraphQLSchema,
  fragments: { [fragmentName: string]: FragmentDefinition },
  rootValue: any,
  operation: OperationDefinition,
  variableValues: { [variableName: string]: any },
}
```

Then we can provide a helper method to returns selections from the `GraphQLResolveInfo`.
I think this could be just like what graphql [documented](https://graphql.org/learn/execution/).

Users had this problem in js too, and [resolved it ](https://stackoverflow.com/questions/48004805/how-to-get-requested-fields-inside-graphql-resolver) by the info param.

keithmattix:
@mehran-prs There are essentially 4 different PRs open for this feature with 4 different designs. You're [not the only one ](https://github.com/graph-gophers/graphql-go/pull/373#issuecomment-571635199) who would like to see this option exposed as a third argument, but it's really up to the maintainers to pick which design they'd like to merge

mehran-prs:
Oh, I see,
Thanks, @keithmattix

pavelnikolov:
I am aware that this has been the most requested feature for this library.
@mehran-prs thank you for sharing this - it is similar to what I want for this library. I don't want the selected fields in the context. That is the reason why it hasn't been merged yet. Yes, most probably there will be a third parameter with the selected fields and their arguments.
I will try to focus on this next week. In the mean time PRs adding this a 3rd param are welcome.
Thank you for your patience!

pavelnikolov:
@ventsislav-georgiev I know that it is a small change, however, this PR is not complete. I've been collecting requirements and apparently we need the field arguments too because they are needed for filtering data, for example, for paging. You are correct that people have been waiting a long time for this feature. However, I want us to get it right because once people start using it then it is difficult to upgrade it. Thank you for your patience and I will try to add it soon.

keithmattix:
In light of @pavelnikolov's direction here, I feel like it would be best for me to close this PR in order to filter out a lot of the noise going on with this feature request. There are at least a couple of other PRs that take the argument approach, and I feel that the community would be better served by iterating on those approaches instead mine. Godspeed in your decision-making @pavelnikolov, and thank you for your hard work maintaining this library.

Issue #210, opened 2018-05-15
Support for Null Values in mutations


Hagbarth:
According to the GraphQL spec, there should be a distinction between explicitly set `null`values and just plain omitted values (spec here: http://facebook.github.io/graphql/draft/#sec-Null-Value).

It can be used to distinguish between ignoring a field and deleting the field value in a mutation fx. 

Is there support for this already and I'm just missing it?

derekpitt:
I am interested this as well.

It would be nice to support `json.RawMessage` for nullable fields in the schema and not require it to be a pointer..

or maybe use a struct tag to skip the runtime type checks against the schema?

I'll poke around and maybe work on a PR.

@neelance or @tonyghita any thoughts on this?

Hagbarth:
@derekpitt we had a pretty severe bug in our system, so I ended up forking and providing a pretty crude solution that enables you to add another argument to your resolver functions, that can tell you if a field was provided or not.

rinatio:
I'm also interested in this. An example would be an update mutation that updates only fields passed in the query (and some of them could be null). There seems to be no way to identify if particular argument was passed as null or simply omitted. Could you please suggest any workarounds?

desdic:
I second a solution on this. I have tried 2 different graphql projects written on go now and they both do not support null vs attribute not there. It would be really nice since this library has a pretty nice interface.

pavelnikolov:
This issue is not something that I am actively looking at. Indeed it is mentioned in the spec that `null` and a missing value should be treated differently. PRs are welcome.

pavelnikolov:
#430 resolved this

Slessi:
Following off of #430 if I wanted to handle the null vs omitted case for an array, e.g. `[String!]`, do I need the below? Or am I overlooking a more straight forward approach?

```go
type NullSliceString struct {
	Value *[]string
	Set   bool
}

func (NullSliceString) ImplementsGraphQLType(name string) bool {
	return name == "[String!]"
}

func (s *NullSliceString) UnmarshalGraphQL(input any) error {
	s.Set = true

	if input == nil {
		return nil
	}

	switch v := input.(type) {
	case []any:
		val := make([]string, len(v))
		for i := range v {
			val[i] = v[i].(string)
		}
		s.Value = &val
		return nil
	default:
		return fmt.Errorf("wrong type for [String!]: %T", v)
	}
}

func (s *NullSliceString) Nullable() {}
```

[...]

PR #430, opened 2021-01-15
Add support for null vs missing input value

Ackar:
This is an attempt to solve https://github.com/graph-gophers/graphql-go/issues/210 by adding support for nullable non-pointer types. This is inspired by the Null types in the `sql` package.

The idea is to have custom types where we can call `UnmarshalGraphQL` with a `nil` input, thus allowing to differentiate between no value (method not called) and a null value (method called with `nil`).

The main change is to allow a non-pointer type to represent a nullable value if it implements the following interface:
```go
type NullUnmarshaller interface {
	Unmarshaler
	Nullable()
}
```

Also added nullable types for existing scalars: `NullString`, `NullInt`, `NullFloat`, `NullBool` and `NullTime`.

**Example**

```graphql
input MyInput {
    a: String
}

type Query {
    myQuery(input: MyInput!): String!
}
```

```go
type MyInput struct {
	A graphql.NullString
}

type resolver struct{}

func (r *resolver) MyQuery(args struct {
	Input MyInput
}) string {
	if args.Input.A.Valid {
		if args.Input.A.Value == nil {
			// null value provided
		} else {
			// non-null value provided
		}
	} else {
		// field omitted
	}

	return "hello world"
}
```

### DIFF
--- a/nullable_types.go
+++ b/nullable_types.go
@@ -1,0 +1,150 @@
+package graphql
+
+import (
+	"fmt"
+)
+
+// NullString is a string that can be null. Use it in input structs to
+// differentiate a value explicitly set to null from an omitted value.
+// When the value is defined (either null or a value) Set is true.
+type NullString struct {
+	Value *string
+	Set   bool
+}
+
+func (NullString) ImplementsGraphQLType(name string) bool {
+	return name == "String"
+}
+
+func (s *NullString) UnmarshalGraphQL(input interface{}) error {
+	s.Set = true
+
+	if input == nil {
+		return nil
+	}
+
+	switch v := input.(type) {
+	case string:
+		s.Value = &v
+		return nil
+	default:
+		return fmt.Errorf("wrong type for String: %T", v)
+	}
+}
+
+func (s *NullString) Nullable() {}
+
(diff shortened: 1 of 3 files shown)


pavelnikolov:
Hi @Ackar
Thank you for your contribution. I'll take a closer look tomorrow.

nullable_types.go
```diff
@@ -0,0 +1,150 @@
...
+	Value *string
+	Valid bool
+}
+
+func (NullString) ImplementsGraphQLType(name string) bool {
+	return name == "String"
+}
+
+func (s *NullString) UnmarshalGraphQL(input interface{}) error {
+	s.Valid = true
```
pavelnikolov, line 20: I wonder if we need to leave this as `Valid`. Isn't `Set` a better choice for a name? I know that this is consistent with the `sql` package. But set sounds better, IMHO.

```go
if args.Input.Int.Valid {
}
```

vs 

```go
if args.Input.Int.Set {
}
```

Ackar: I agree, `Set` is better. I've updated the PR.

pavelnikolov:
Thank you

PR #433, opened 2021-02-12
Adding support for partial returns under context deadlines

placeybordeaux-remitly:
# Motivation

We'd like to support taking in a timeout/deadline at the request level and returning as many of the fields as we can within that deadline.

Initial investigations showed that if any resolvers take longer than the deadline then no data is returned.

# Within the scope of this PR

We add the ability to return all fields that have completed if a context times out or is canceled.
Adding two test cases that verify behavior for context timing out as well as context cancellation.

# Outside the scope of this PR

* Extending this support to cover the synchronous code path. This currently only supports the async code path.
* Add an error message that identifies which field timed out
* Supporting context canceling properly for subscriptions

# Misc
Additionally it looks like we've added support for building the response & returning it as soon as the context times out. Previously if we gave a context deadline but our resolvers did not respect it then we'd block until the http request is closed or the resolver returns, at which point we'd throw away the output & return an error.

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -51,7 +51,9 @@ func (r *Request) Execute(ctx context.Context, s *resolvable.Schema, op *query.O
 	}()
 
 	if err := ctx.Err(); err != nil {
-		return nil, []*errors.QueryError{errors.Errorf("%s", err)}
+		//If context has either been cancelled or timed out we still may want to return the features that have finished
+		// TODO properly attribute mark which features have timedout in the error field
+		return out.Bytes(), []*errors.QueryError{errors.Errorf("%s", err)}
 	}
 
 	return out.Bytes(), r.Errs
@@ -62,6 +64,7 @@ type fieldToExec struct {
 	sels     []selected.Selection
 	resolver reflect.Value
 	out      *bytes.Buffer
+	finished bool
 }
 
 func resolvedToNull(b *bytes.Buffer) bool {
@@ -85,7 +88,17 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 				execFieldSelection(ctx, r, s, f, &pathSegment{path, f.field.Alias}, true)
 			}(f)
 		}
-		wg.Wait()
+		// close a signal channel once the wait group is complete
+		c := make(chan struct{})
+		go func() {
+			defer close(c)
+			wg.Wait()
+		}()
+		// We want to block until either the context is done or the wait group is complete
+		select {
+		case <-c:
+		case <-ctx.Done():
+		}
 	} else {
 		for _, f := range fields {
 			f.out = new(bytes.Buffer)
@@ -111,6 +124,11 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 		out.WriteString(f.field.Alias)
 		out.WriteByte('"')
 		out.WriteByte(':')
+		// if this field hasn't finished yet, then it's timed out. Record it as null
+		if !f.finished {
+			out.WriteString("null")
+			continue
+		}
 		out.Write(f.out.Bytes())
 	}
 	out.WriteByte('}')
@@ -182,6 +200,9 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
 				err = makePanicError(panicValue)
 				err.Path = path.toSlice()
 			}
+		}()
+		defer func() {
+			f.finished = true
 		}()
 
 		if f.field.FixedResult.IsValid() {
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -53,7 +53,7 @@ func (r *Request) Execute(ctx context.Context, s *resolvable.Schema, op *query.O
 	if err := ctx.Err(); err != nil {
 		//If context has either been cancelled or timed out we still may want to return the features that have finished
 		// TODO properly attribute mark which features have timedout in the error field
-		return out.Bytes(), []*errors.QueryError{errors.Errorf("%s", err)}
+		return out.Bytes(), r.Errs
 	}
 
 	return out.Bytes(), r.Errs
@@ -90,14 +90,14 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 			}(f)
 		}
 		// close a signal channel once the wait group is complete
-		c := make(chan struct{})
+		done := make(chan struct{})
 		go func() {
-			defer close(c)
+			defer close(done)
 			wg.Wait()
 		}()
 		// We want to block until either the context is done or the wait group is complete
 		select {
-		case <-c:
+		case <-done:
 		case <-ctx.Done():
 		}
 	} else {
@@ -113,9 +113,14 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 		// If a non-nullable child resolved to null, an error was added to the
 		// "errors" list in the response, so this field resolves to null.
 		// If this field is non-nullable, the error is propagated to its parent.
-		if _, ok := f.field.Type.(*common.NonNull); ok && resolvedToNull(f.out) {
+		if _, ok := f.field.Type.(*common.NonNull); ok && (resolvedToNull(f.out) || !f.finished) {
 			out.Reset()
 			out.Write([]byte("null"))
+			if !f.finished { // if we haven't finished yet that means we haven't recorded this failure yet
+				err := errors.Errorf(ctx.Err().Error())
+				err.Path = append(path.toSlice(), f.field.Alias)
+				r.AddError(err)
+			}
 			f.lock.Unlock()
 			return
 		}
@@ -129,6 +134,9 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 		out.WriteByte(':')
 		// if this field hasn't finished yet, then it's timed out. Record it as null
 		if !f.finished {
+			err := errors.Errorf(ctx.Err().Error())
+			err.Path = append(path.toSlice(), f.field.Alias)
+			r.AddError(err)
 			out.WriteString("null")
 			continue
 		}
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -64,6 +64,7 @@ type fieldToExec struct {
 	sels     []selected.Selection
 	resolver reflect.Value
 	out      *bytes.Buffer
+	lock     sync.Mutex
 	finished bool
 }
 
@@ -108,12 +109,14 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 
 	out.WriteByte('{')
 	for i, f := range fields {
+		f.lock.Lock()
 		// If a non-nullable child resolved to null, an error was added to the
 		// "errors" list in the response, so this field resolves to null.
 		// If this field is non-nullable, the error is propagated to its parent.
 		if _, ok := f.field.Type.(*common.NonNull); ok && resolvedToNull(f.out) {
 			out.Reset()
 			out.Write([]byte("null"))
+			f.lock.Unlock()
 			return
 		}
 
@@ -130,6 +133,7 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 			continue
 		}
 		out.Write(f.out.Bytes())
+		f.lock.Unlock()
 	}
 	out.WriteByte('}')
 }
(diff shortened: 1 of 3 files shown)


placeybordeaux-remitly:
It looks like the case where there is a timeout on a non-nullable field is not handled properly. I'll be adding tests for that and fixing it shortly.

placeybordeaux-remitly:
This appears to introduce a race condition because modifying the out bytes.Buffer can happen before and after the waitgroup.Wait().

This means that to accomplish this we'd need more significant changes. I'll take a look at a reasonable path forward. If any maintainers have an opinion please let me know!

placeybordeaux-remitly:
`go test -race timeout_test.go` is passing, but `go test -race ./...` is failing. Working on that.

placeybordeaux-remitly:
`go test -race ./...` appeared to be failing erroneously due to insufficient separation between the timeout tests.

It doesn't look like there is any benchmarking code so I can't say definitely how much the locking & unlocking matters, but there should only be lock contention after a timeout has already occurred so I believe it is unlikely to matter in the happy case. Furthermore the lock is only acquired before performing tasks that are well timebounded (e.g. out.WriteString("{"))

internal/exec/exec.go
```diff
@@ -111,7 +127,13 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 		out.WriteString(f.field.Alias)
 		out.WriteByte('"')
 		out.WriteByte(':')
+		// if this field hasn't finished yet, then it's timed out. Record it as null
+		if !f.finished {
+			out.WriteString("null")
```
jonahglover, line 132: So here's something that makes me a bit nervous _theoretically_ though practically I think we can get around it. [The spec](http://spec.graphql.org/June2018/#sec-Type-System.Non-Null) denotes that when you add the `!` after a type, the output is non-nullable. If you look around some of our resolver methods, you'll see some return `*TYPE` and some just `TYPE`. 

It seems like our default behaviour is to return null if the context deadline is exceeded. I actually think this makes sense however I think it means this API no longer conforms to the GraphQL spec. We'd need to think of some solution to this.

jonahglover: At least internally, I think this would break marshaling logic.

timeout_test.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (m MetaMessage) Msg(args struct{ Timeout int32 }) string {
+	time.Sleep(time.Duration(args.Timeout) * time.Millisecond)
+	s := "MetaSuccess!"
+	return s
+}
+
+func TestPartialSuccessTimeout(t *testing.T) {
+	cxt, _ := context.WithDeadline(context.Background(), time.Now().Add(20*time.Millisecond)) // This test now depends on the simplest resolver returning within 10 milliseconds
```
jonahglover, line 37: nit: `cxt` vs canonical `ctx`

internal/exec/exec.go
jonahglover: I'm seeing a related comment [above](https://github.com/graph-gophers/graphql-go/pull/433/files#diff-87ea4cfcdd6bb8062908727b36ef05d305ddfbd5ed89322e2276f315ac63dd1bR115). 

I believe returning a `null` for a non-nullable field is actually prevented at run time. The schema parser will verify that matching resolver methods have the proper return types. Maybe something like that can be done here? I don't know.

internal/exec/exec.go
```diff
@@ -95,12 +109,14 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 
 	out.WriteByte('{')
 	for i, f := range fields {
+		f.lock.Lock()
```
jonahglover, line 112: Can you help me understand why we need this mutex?

internal/exec/exec.go
```diff
@@ -62,6 +64,8 @@ type fieldToExec struct {
 	sels     []selected.Selection
 	resolver reflect.Value
 	out      *bytes.Buffer
+	lock     sync.Mutex
```
jonahglover, line 67: We need this mutex to prevent race conditions where we might try to write the output and resolve the field at the same time. The lock prevents us from updating the `finished` bool or writing the output until whichever finished first is completed. 


Is this correct?

internal/exec/exec.go
```diff
@@ -85,7 +89,17 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
...
+		c := make(chan struct{})
+		go func() {
+			defer close(c)
+			wg.Wait()
+		}()
+		// We want to block until either the context is done or the wait group is complete
+		select {
+		case <-c:
+		case <-ctx.Done():
+		}
```
jonahglover, line 102: this is a nice solution

internal/exec/exec.go
```diff
@@ -85,7 +89,17 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
 				execFieldSelection(ctx, r, s, f, &pathSegment{path, f.field.Alias}, true)
 			}(f)
 		}
-		wg.Wait()
+		// close a signal channel once the wait group is complete
+		c := make(chan struct{})
```
jonahglover, line 93: nit: can we name this something like `done`? No hard and fast rules but it's slightly more canonical and readable.

internal/exec/exec.go
```diff
@@ -85,7 +89,17 @@ func (r *Request) execSelections(ctx context.Context, sels []selected.Selection,
...
+		go func() {
+			defer close(c)
+			wg.Wait()
+		}()
+		// We want to block until either the context is done or the wait group is complete
+		select {
+		case <-c:
+		case <-ctx.Done():
```
jonahglover, line 101: I wonder if we need to do anything if `ctx.Error()` is non nil?

jonahglover: I guess we're expecting this to only be hit when deadline is exceeded but what should we do if that's not the error?

internal/exec/exec.go
```diff
@@ -51,7 +51,9 @@ func (r *Request) Execute(ctx context.Context, s *resolvable.Schema, op *query.O
 	}()
 
 	if err := ctx.Err(); err != nil {
-		return nil, []*errors.QueryError{errors.Errorf("%s", err)}
+		//If context has either been cancelled or timed out we still may want to return the features that have finished
+		// TODO properly attribute mark which features have timedout in the error field
+		return out.Bytes(), []*errors.QueryError{errors.Errorf("%s", err)}
```
jonahglover, line 56: Is there a way to only add our new behaviour if the error was a context deadline timeout?

jonahglover: Is it too brittle to switch on the error being "context deadline exceeded"?

internal/exec/exec.go
jonahglover: EDIT: looks like this is handled above when we write the final output

jonahglover:
Wondering, how does this interact with the opentracing library? This library has really nice out-of-the-box open tracing and one of those is that fields that error out are called out in open tracing. If we aren't doing that with this PR we're going to lose that visibility in jaeger traces.

internal/exec/exec.go
```diff
@@ -51,7 +51,9 @@ func (r *Request) Execute(ctx context.Context, s *resolvable.Schema, op *query.O
 	}()
 
 	if err := ctx.Err(); err != nil {
-		return nil, []*errors.QueryError{errors.Errorf("%s", err)}
+		//If context has either been cancelled or timed out we still may want to return the features that have finished
+		// TODO properly attribute mark which features have timedout in the error field
```
jonahglover, line 55: is there a potential race condition where the deadline is exceeded and before we've responded a field resolves and writes an error to the output so we got the impression that a field failed for some reason but actually it was because the deadline exceeded?

jonahglover, line 55: It seems like this PR somehow silences all errors if the context deadline is exceeded, even if those errors showed up before the deadline. Can you confirm if that's true or not? This might have been what you were mentioning to me the other day but I think I hadn't dug into the PR enough to understand the implications.

jonahglover: I notice locally that if the context deadline is exceeded I only get that one error and if the context deadline is not exceeded I got all sorts of underlying errors.

internal/exec/exec.go
placeybordeaux-remitly: Yup!

internal/exec/exec.go
placeybordeaux-remitly: The behavior before this PR was to block until all resolvers have finished, aggregate the errors and then throw them away if the context deadline has been exceeded.

Currently if we get an error, and then another resolver makes the context deadline exceed then we will throw that error away.

Adding the ability to persist errors even in the face of a context deadline being exceeded seems like a reasonable scope for this PR.

internal/exec/exec.go
placeybordeaux-remitly: As of time of writing the only two errors that context.Error() will return are if the context deadline is exceeded or the context is canceled.

https://golang.org/pkg/context/#Context

internal/exec/exec.go
placeybordeaux-remitly: The third test I wrote TestTotalFailureOnNonNull shows that the current code will happily record null on non-null features if a timeout has occurred. It doesn't look like the schema parser stops this from happening.

jonahglover: > The third test I wrote TestTotalFailureOnNonNull shows that the current code will happily record null on non-null features if a timeout has occurred. It doesn't look like the schema parser stops this from happening.

On master, if you try to do this you'll get errors in your console when you try to query the graphql API so we probably don't want to introduce this unexpected behaviour.

placeybordeaux-remitly: I have changed it so the timeout case obeys the non-null behavior we want.

placeybordeaux-remitly:
Converting to draft because the following features are still not supported:

- Adding an error when we haven't even started processing the field before the context deadline is exceeded.
- Handling lists of fields properly

Other things that we don't support, but I wasn't going to include:

- Handling context timeouts when running synchronously properly.

placeybordeaux-remitly:
@pavelnikolov Apologies if this is just noise, but given the issues that cropped up above and the potential scope of necessary changes I was going to pause work on this until there is some guidance from you or another maintainer as to whether this change or a similar one could eventually get merged into master.

Obviously there are some outstanding issues and increasingly it's looking like there isn't a simple fix, I'd love to hear your thoughts as to whether this is a feature you'd like to have & directions as to how you'd like to accomplish it.

Thanks for all your work & cheers!

p.s. Please let me know if there is someone else that might have opinions on the matter! I pinged you because you were the last person to merge something, I couldn't find a maintainers document.

internal/exec/subscribe.go
```diff
@@ -154,7 +154,7 @@ func (r *Request) Subscribe(ctx context.Context, s *resolvable.Schema, op *query
 					}()
 
 					if err := subCtx.Err(); err != nil {
-						c <- &Response{Errors: []*errors.QueryError{errors.Errorf("%s", err)}}
+						c <- &Response{Data: nil, Errors: []*errors.QueryError{errors.Errorf("%s", err)}}
```
pavelnikolov, line 157: Why do you need to add the `Data: nil`?

timeout_test.go
```diff
@@ -0,0 +1,177 @@
...
+			},
+			Context: ctx,
+		})
+}
+
+func TestPartialSuccessCancel(t *testing.T) {
+
+	ctx2, cancelFunc := context.WithDeadline(context.Background(), time.Now().Add(50*time.Millisecond))
+	go func() {
+		time.Sleep(20 * time.Millisecond)
```
pavelnikolov, line 71: I don't like seeing `time.Sleep` in tests. You can mock the behaviour of a timeout. Or you can cancel the context which should result in similar behaviour, non?

placeybordeaux-remitly: That's a fair concern, I can switch to only doing cancels. They do give different error messages, but there is no switching based off of the actual error text.

internal/exec/subscribe.go
placeybordeaux-remitly: Hmm looking at that now it does look odd. I think I will need to go back and rethink the stuff in internal/exec/subscribe, now with greater context I am not actually sure how timeouts should really work with subscribe

internal/exec/exec.go
pavelnikolov: Shouldn't the `null` propagate to the nearest null-able parent if a resolver times out? We had this issue before with other errors and I believe that it was fixed, right? So timeout errors should be like any other errors.

placeybordeaux-remitly: The issue I was trying to solve here was that the field would not be set to null nor have an error attached to it if a context timeout occurred. We'd start marshaling our result before that resolver ever finished, that's why I am keying off of !finished. I tried to handle the case of propegating that error to the parent above here, this is the case where the field is nullable and a timeout occured.

placeybordeaux-remitly:
@pavelnikolov Appologies for the ping, but I'm currently trying to figure out whether I was going to put in the effort to fix all of the outstanding issues & get this working with the other parts (e.g. lists).

do you think that this feature & the general approach outlined in this PR is something you'd like to see merged?

placeybordeaux-remitly:
Closing this as it has conflicts with master, hasn't pasted CR & we've found a work around.

Issue #416, opened 2020-11-08
Add Support for Interfaces Implementing Interfaces


lukephillippi:
## Expected Behavior

As implemented via graphql/graphql-spec#373, [the GraphQL draft specification](https://spec.graphql.org/draft/#sec-Interfaces.Interfaces-Implementing-Interfaces) now indicates that schema definitions can include interfaces that implement other interfaces. As an example, the following should be considered valid schema definition syntax and thus be parsed successfully by the `ParseSchema` and `MustParseSchema` functions:

```graphql
interface Node {
  id: ID!
}

interface Resource implements Node {
  id: ID!
  url: String
}
```

## Actual Behavior

However, it appears that this syntax is not yet supported with the current version of the `graph-gophers/graphql-go` package ([`v0.0.0-20201027172035-4c772c181653`](https://github.com/graph-gophers/graphql-go/tree/4c772c181653467cd21faaca62cf0a97f5a799d8)).

Attempting to parse a schema definition using this new syntax results in an error:

```shell
...
panic: graphql: syntax error: unexpected "implements", expecting "{" (line 10, column 20)
...
```

## Reproduction Code ([Go Playground](https://play.golang.org/p/I8NAUUJYMys))

```go
package main

import "github.com/graph-gophers/graphql-go"

const schema = `
type Query {
  node: Node!
}

interface Node {
  id: ID!
}

interface Resource implements Node {
  id: ID!
  url: String
}
`

type resolver struct{}

func (_ *resolver) Node() *node {
	return &node{}
}

type node struct{}

func (_ *node) ID() graphql.ID {
	return graphql.ID("foo")
}

func (n *node) ToResource() (*resource, bool) {
	return &resource{n}, true
}

type resource struct {
	*node
}

func (_ *resource) URL() *string {
	u := "https://github.com/graph-gophers/graphql-go"
	return &u
}

func main() {
	graphql.MustParseSchema(schema, &resolver{})
}
```

Any chance that y'all could add support for the interfaces implementing interfaces syntax?

Thank you in advance and for the great framework!

pavelnikolov:
Closing this issue as it has been resolved by https://github.com/graph-gophers/graphql-go/pull/471

PR #436, opened 2021-03-05
Add support for interfaces implementing other interfaces

tomemmerson:
To fix issue [Add support for interfaces implementing other interfaces](https://github.com/graph-gophers/graphql-go/issues/416)

To match new draft spec here: http://spec.graphql.org/draft/#sec-Interfaces.Interfaces-Implementing-Interfaces
Allows the following functionality:
```graphql
interface Node {
  id: ID!
}

interface Resource implements Node {
  id: ID!
  url: String
}
```

The implementation here isn't massively efficient, I check multiple interfaces multiple times, mostly due to the requirement that

> Transitively implemented interfaces (interfaces implemented by the interface that is being implemented) must also be defined on an implementing type or interface. For example, Image cannot implement Resource without also implementing Node:
https://spec.graphql.org/draft/#sel-FAHbhBHCAACGB35P

There's definitely a more efficient solution than the one here, really just wanted an opinion as to whether it is worth it. 

Thanks
Tom

### DIFF
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -48,6 +48,7 @@ type Schema struct {
 	unions          []*Union
 	enums           []*Enum
 	extensions      []*Extension
+	interfaces      []*Interface
 }
 
 // Resolve a named type in the schema by its name.
@@ -101,11 +102,13 @@ type Object struct {
 //
 // http://facebook.github.io/graphql/draft/#sec-Interfaces
 type Interface struct {
-	Name          string
-	PossibleTypes []*Object
-	Fields        FieldList // NOTE: the spec refers to this as `FieldsDefinition`.
-	Desc          string
-	Directives    common.DirectiveList
+	Name           string
+	PossibleTypes  []*Object
+	Fields         FieldList // NOTE: the spec refers to this as `FieldsDefinition`.
+	Desc           string
+	Directives     common.DirectiveList
+	Interfaces     []*Interface // NOTE: http://spec.graphql.org/draft/#sec-Interfaces.Interfaces-Implementing-Interfaces
+	interfaceNames []string
 }
 
 // Union types represent objects that could be one of a list of GraphQL object types, but provides no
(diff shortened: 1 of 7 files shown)


internal/schema/schema.go
```diff
@@ -770,3 +832,30 @@ func parseFieldsDef(l *common.Lexer) FieldList {
...
+//
+// NOTE: https://spec.graphql.org/draft/#sel-FAHbhBHCAACGB35P
+func interfaceDependancies(iface Interface) []string {
+	var search func(requiredNames []string, iface Interface) []string
+
+	search = func(requiredNames []string, iface Interface) []string {
+		for _, f := range iface.Interfaces {
+			if f == nil {
+				continue
+			}
```
pavelnikolov, line 851: Can this happen? Is it possible that one of the `iface.Interfaces` is `nil`?

internal/schema/schema.go
```diff
@@ -340,6 +343,52 @@ func (s *Schema) Parse(schemaString string, useStringDescriptions bool) error {
...
+
+	for _, iface := range s.interfaces {
+		if iface == nil {
+			continue
+		}
+		expectedNames := interfaceDependancies(*iface)
+		actualNames := iface.interfaceNames
+
+		expMap := make(map[string]int)
+		actMap := make(map[string]int)
```
pavelnikolov, line 376: Could you, please, remove the `Map` suffix from the names. From the syntax below it is obvious that it is a map.

internal/schema/schema.go
tomemmerson: No, good point, ifaces should never be nil in the current implementation. Just used to checking pointer types, is there a reason all the types are using pointers or can I change `Schema.Interfaces` to a non-pointer type and remove this check?

pavelnikolov: I'm not sure what the reason is since I haven't written this code but my guess is that the author didn't want to copy structs by value for performance reasons. On the other hand, interfaces are not supposed to be huge. But we never know how people use it. So let's leave it consistent with all other types.

I'd prefer if both the `interfaceDependancies` and `search` functions accept a reference. Then at the beginning of the `search` function check if the passed pointer to the `Interface` struct is `nil` and if yes, then panic. This is never supposed to happen.

internal/schema/schema.go
```diff
@@ -340,6 +343,52 @@ func (s *Schema) Parse(schemaString string, useStringDescriptions bool) error {
...
+				if iface.Fields.Get(f) == nil {
+					return errors.Errorf("interface %q expects field %q but %q does not provide it", intfName, f, iface.Name)
+				}
+			}
+
+			iface.Interfaces[i] = intf
+		}
+	}
+
+	for _, iface := range s.interfaces {
```
pavelnikolov, line 368: In order to avoid checking interfaces multiple times, you might define a variable of type `map[string][]string` above this `for` statement that contains all the dependencies for each interface. Then you will check every interface exactly once if it is not in the map already. 
However, if I understand correctly, you will need to sort all the interfaces by `len(interfaceNames)` ascending. This way interfaces that implement multiple other interfaces will make use of that map efficiently. This map will be used only once at runtime and then it will be garbage collected.

internal/schema/schema.go
```diff
@@ -770,3 +832,30 @@ func parseFieldsDef(l *common.Lexer) FieldList {
 	}
 	return fields
 }
+
+// interfaceDependancies searches through the tree of interface relations
+// to build a slice of all the names that should be implemented on the
+// given interface
+//
+// TODO: Would be more efficient not to search every interface multiple
+// times.
```
pavelnikolov, line 841: See my comment above how to handle that.

pavelnikolov:
There are conflicts

suessflorian:
👋  hey, what's the status with this PR? Any way I can help out? I just read through https://github.com/graphql/graphql-spec/pull/373 this could actually be very useful on a project I'm working on so keen to see this getting implemented. Awesome work so far though @tomemmerson 👍

pavelnikolov:
@suessflorian I'd be happy to merge this change if you could resolve the conflicts

suessflorian:
No worries @tomemmerson, your start here helped a lot - I've opened up a new PR continuing on from here and so I think we can close this PR @pavelnikolov 👍

pavelnikolov:
Closing this PR because of #471

Issue #434, opened 2021-02-12
Allow access to the AST and add a Visitor API


rudle:
# Problem Statement

The AST created by graphql-go's parser is locked away inside of an internal package. This means that clients of this library cannot access their schema in an abstract form. Historically, several would-be contributors to this library have made hard forks in order to work around this decision ([1], [2]). These hard forks lead to duplicated work and a lack of cohesion across the tools available to GraphQL engineers working in Go. While not a perfect metric, 99designs/gqlgen has more stars and forks than this repo which suggests that would-be contributors to graphql-go are spending their energy working on gqlgen instead.

# Recommended Approach

I propose a new non-internal package containing a representation of the schema's AST and all of the types used to represent a Schema and an Operation. This package would be the canonical place for types, replacing the types in `internal/common`.

Additionally, graphql-go should expose a visitor pattern AST traversal library, similar to graphql-js[3] so that tools such as code generators, linters and formatters can be written based on graphql-go's AST.

## Implementation

I am currently working on the first portion of the approach recommended above. I hope to have a version of graphql-go based on a publicly accessible type system in the next couple of weeks. This change alone will allow my team to deprecate our internal fork and move all of our tooling to depend on the master branch of graphql-go.

We are also happy to contribute code to get us to the next goal described in the approach, but the timeline is less clear. Thinking optimistically, it's possible that some other members of the community will contribute Pull Requests towards this goal once the project's leadership approves of this direction.

# Alternatives

If graphql-go decides not to expose the AST via a publicly-accessible package, the alternatives available are:
1) customers can fork graphql-go into their own repos, which are likely to get out of sync with this repo
2) customers can use a different tool to understand the AST in their tooling and at runtime

The disadvantages to 1) are pretty clear to me, especially because I am working on bringing a divergent fork back up to date with this repo. This option wastes engineering time proportional proportional to all repos that depend on this one. As the number of dependents grows, the wasted time grows also.
Option 2) is worth discussing as it allows this repo to maintain flexibility in how it represents the AST and keep its API small. However, implementation differences between this tool's parser and other tools may cause unexpected results.

## Related Issues

#116 describes this issue and has gathered quite a bit of support but does not chart a path forward.

[1] https://github.com/99designs/gqlgen/issues/35#issuecomment-369055382
[2] https://github.com/DealTap/graphql-gen-go
[3] https://github.com/graphql/graphql-js/blob/main/src/language/visitor.js#L225

Issue #116, opened 2017-10-06
Parser is locked away in /internal/


shelbyramsey:
Recently for a project we've had a need to parse & store gql queries in a structured manner that can easily be queried in a database. After quite minimal effort I was able to accomplish this thanks to the well done parser in the /internal/ directory.

That being said the code can not be used outside of the repo. I can appreciate not wanting to support backwards compatibility on the API but it was not clear as I was digging through it whether this was a concern or not. I'd be happy to do the work to deal with the migration if the maintainers are open to it. 

Thanks again for the great contribution to the community!

Vektah:
I would love to see `internal/parser`, `internal/validation` and `internal/schema` pulled into stand alone library. 

They would be very useful for building all kinds of graphql cli tools and proxies.

emwalker:
It's a pity.  I was quite enthusiastic about this library, until I realized I would need to get ahold of a parsed document to do what I needed, and then I discovered that the code for this is in `internal`.  This friendly worded issue was filed in October of last year, and no response.

vvakame:
Hi, I'm very interested in this topic.
I want to know what members are thinking 🌹 

http://facebook.github.io/graphql/June2018/ is out.
[vgo](https://github.com/golang/go/wiki/vgo#current-state) supports [semver](https://semver.org/), we can also specify a breaking change.

bsr203:
I use couple of libs written by @vvakame and his contribution to graphql-go would be great!! Also agree with versioning this lib as per the graphql spec version.

emwalker:
For anyone encountering this, see gqlgen, which is a [hard fork](https://github.com/99designs/gqlgen/issues/35).

pavelnikolov:
#437 437 solves that

PR #437, opened 2021-03-16
add types package

rudle:
Part of #434 and related to #116 this change adds a new package containing all
types used by graphql-go in representing the GraphQL specification. The names
used in this package should match the specification as closely as possible.

In order to have cohesion, all internal packages that use GraphQL types have
been changed to use this new package.

This change is large but mostly mechanical. I recommend starting by reading
through the `types` package to build familiarity. I'll call out places in the
code where I made decisions and what the tradeoffs were.

No new tests were added but existing tests were changed to use the new types.

Regarding backward compatibility, the external `MustParseSchema` entrypoint still works as expected. We verified that no code changes were required by importing this change into our main GraphQL application.

### DIFF
--- a/internal/query/query.go
+++ b/internal/query/query.go
@@ -6,97 +6,19 @@ import (
 
 	"github.com/graph-gophers/graphql-go/errors"
 	"github.com/graph-gophers/graphql-go/internal/common"
+	"github.com/graph-gophers/graphql-go/types"
 )
-
-type Document struct {
-	Operations OperationList
-	Fragments  FragmentList
-}
-
-type OperationList []*Operation
-
-func (l OperationList) Get(name string) *Operation {
-	for _, f := range l {
-		if f.Name.Name == name {
-			return f
-		}
-	}
-	return nil
-}
-
-type FragmentList []*FragmentDecl
-
-func (l FragmentList) Get(name string) *FragmentDecl {
-	for _, f := range l {
-		if f.Name.Name == name {
-			return f
-		}
-	}
-	return nil
-}
-
-type Operation struct {
-	Type       OperationType
-	Name       common.Ident
-	Vars       common.InputValueList
-	Selections []Selection
-	Directives common.DirectiveList
-	Loc        errors.Location
-}
-
-type OperationType string
 
 const (
-	Query        OperationType = "QUERY"
-	Mutation                   = "MUTATION"
-	Subscription               = "SUBSCRIPTION"
+	Query        types.OperationType = "QUERY"
+	Mutation                         = "MUTATION"
+	Subscription                     = "SUBSCRIPTION"
 )
 
-type Fragment struct {
-	On         common.TypeName
-	Selections []Selection
-}
-
-type FragmentDecl struct {
-	Fragment
-	Name       common.Ident
-	Directives common.DirectiveList
-	Loc        errors.Location
-}
-
-type Selection interface {
-	isSelection()
-}
-
-type Field struct {
-	Alias           common.Ident
-	Name            common.Ident
-	Arguments       common.ArgumentList
-	Directives      common.DirectiveList
-	Selections      []Selection
-	SelectionSetLoc errors.Location
-}
-
-type InlineFragment struct {
-	Fragment
-	Directives common.DirectiveList
-	Loc        errors.Location
-}
-
-type FragmentSpread struct {
-	Name       common.Ident
-	Directives common.DirectiveList
-	Loc        errors.Location
-}
-
-func (Field) isSelection()          {}
-func (InlineFragment) isSelection() {}
-func (FragmentSpread) isSelection() {}
-
-func Parse(queryString string) (*Document, *errors.QueryError) {
+func Parse(queryString string) (*types.ExecutableDefinition, *errors.QueryError) {
 	l := common.NewLexer(queryString, false)
 
-	var doc *Document
+	var doc *types.ExecutableDefinition
 	err := l.CatchSyntaxError(func() { doc = parseDocument(l) })
 	if err != nil {
 		return nil, err
@@ -105,12 +27,12 @@ func Parse(queryString string) (*Document, *errors.QueryError) {
 	return doc, nil
 }
 
-func parseDocument(l *common.Lexer) *Document {
-	d := &Document{}
+func parseDocument(l *common.Lexer) *types.ExecutableDefinition {
+	d := &types.ExecutableDefinition{}
 	l.ConsumeWhitespace()
 	for l.Peek() != scanner.EOF {
 		if l.Peek() == '{' {
-			op := &Operation{Type: Query, Loc: l.Location()}
+			op := &types.OperationDefinition{Type: Query, Loc: l.Location()}
 			op.Selections = parseSelectionSet(l)
 			d.Operations = append(d.Operations, op)
 			continue
@@ -141,8 +63,8 @@ func parseDocument(l *common.Lexer) *Document {
 	return d
 }
 
-func parseOperation(l *common.Lexer, opType OperationType) *Operation {
-	op := &Operation{Type: opType}
+func parseOperation(l *common.Lexer, opType types.OperationType) *types.OperationDefinition {
+	op := &types.OperationDefinition{Type: opType}
 	op.Name.Loc = l.Location()
 	if l.Peek() == scanner.Ident {
 		op.Name = l.ConsumeIdentWithLoc()
@@ -163,18 +85,18 @@ func parseOperation(l *common.Lexer, opType OperationType) *Operation {
 	return op
 }
 
-func parseFragment(l *common.Lexer) *FragmentDecl {
-	f := &FragmentDecl{}
+func parseFragment(l *common.Lexer) *types.FragmentDefinition {
+	f := &types.FragmentDefinition{}
 	f.Name = l.ConsumeIdentWithLoc()
 	l.ConsumeKeyword("on")
-	f.On = common.TypeName{Ident: l.ConsumeIdentWithLoc()}
+	f.On = types.TypeName{Ident: l.ConsumeIdentWithLoc()}
 	f.Directives = common.ParseDirectives(l)
 	f.Selections = parseSelectionSet(l)
 	return f
 }
 
-func parseSelectionSet(l *common.Lexer) []Selection {
-	var sels []Selection
+func parseSelectionSet(l *common.Lexer) []types.Selection {
+	var sels []types.Selection
 	l.ConsumeToken('{')
 	for l.Peek() != '}' {
 		sels = append(sels, parseSelection(l))
@@ -183,15 +105,15 @@ func parseSelectionSet(l *common.Lexer) []Selection {
 	return sels
 }
 
-func parseSelection(l *common.Lexer) Selection {
+func parseSelection(l *common.Lexer) types.Selection {
 	if l.Peek() == '.' {
 		return parseSpread(l)
 	}
-	return parseField(l)
+	return parseFieldDef(l)
 }
 
-func parseField(l *common.Lexer) *Field {
-	f := &Field{}
+func parseFieldDef(l *common.Lexer) *types.Field {
+	f := &types.Field{}
 	f.Alias = l.ConsumeIdentWithLoc()
 	f.Name = f.Alias
 	if l.Peek() == ':' {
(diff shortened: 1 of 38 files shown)


rudle:
I started integrating this package with an internal tool we have and realized I made an oversight in my initial submission. The `*types.Schema` is not actually publicly accessible! I added a getter but I'm open to other solutions that preserve backward compatibility too https://github.com/graph-gophers/graphql-go/pull/437/files#diff-98bf38fb290ecb143188e1585622022aaa0711a888da5e74b254e01a66d0eb99R86

pavelnikolov:
Just out of curiosity - what is. the internal tool that you are using with this code?

types/query.go
```diff
@@ -0,0 +1,62 @@
...
+	Loc        errors.Location
+}
+
+type OperationType string
+
+// A Selection is a field requested in a GraphQL operation.
+//
+// http://spec.graphql.org/draft/#Selection
+type Selection interface {
+	isSelection()
```
pavelnikolov, line 32: What is the point of making this interface public if its only method is private?

internal/query/query.go
```diff
@@ -6,97 +6,19 @@ import (
...
-func (Field) isSelection()          {}
-func (InlineFragment) isSelection() {}
-func (FragmentSpread) isSelection() {}
-
-func Parse(queryString string) (*Document, *errors.QueryError) {
+func Parse(queryString string) (*types.ExecutableDefinition, *errors.QueryError) {
 	l := common.NewLexer(queryString, false)
 
-	var doc *Document
+	var doc *types.ExecutableDefinition
```
pavelnikolov, line 21: Why is `Document` replaced with `ExecutableDefinition`?
As the spec says a document might include other things other than just `ExecutableDefinition`,  no?

> A GraphQL Document describes a complete file or request string operated on by a GraphQL service or client. 

Although, it also says that:

> GraphQL services which only seek to provide GraphQL query execution may choose to only include ExecutableDefinition and omit the TypeSystemDefinition and TypeSystemExtension rules from Definition.

The question is do we want to exclude `TypeSystemDefinition` and `TypeSystemExtension`?

rudle: We felt that the type called Document did not match up with the specification of a Document. As you mention, a document may contain an `ExecutableDefinition` but also a `TypeSystemDefinition` and extensions. It's essentially the root type. The grammar describes this here: http://spec.graphql.org/June2018/#sec-Appendix-Grammar-Summary.Document and the type itself is documented here: https://github.com/graph-gophers/graphql-go/pull/437/files/130895f516efd0fc6c23dca24bcdf2a8cf0c24bf#diff-b349fbf81a907441c5bb68aa1c730e71b6492354cd594b63626d781dc0bd51e7R9

Rather than add more things to Document and refactor everything to work based on that root type, we thought that making this type more narrow (in name only) was a better choice as it makes this change a bit more surgical and less risky. Another advantage of this naming choice is that the name `types.Document` is free to use as the root of the AST to match the grammar and specification. I'm open to making this change, but it would mean changing a lot of code and likely retiring `types.Schema` and `schema.Schema`.

I noticed that the variable name (`doc`) did not get renamed along with the type. That's an oversight and I'll do a pass to check for this type of error later on today.

rudle:
> Just out of curiosity - what is the internal tool that you are using with this code?

Good question!

We wrote a linter to catch common errors made by contributors to our shared codebase. It checks for things like: 
- do mutations and their inputs have similar names?
- does the type and its fields repeat words (eg avoid: `FriendRequests { friendRequestID }`)
- is pagination implemented correctly
and so on.

internal/query/query.go
pavelnikolov: > I noticed that the variable name (doc) did not get renamed along with the type. That's an oversight and I'll do a pass to check for this type of error later on today.

That would be idea, to avoid confusion in future.

types/query.go
rudle: This interface was carried over from the old code where it had the same quirks of visibility. It needs to be public because there is code in `internal/common` and elsewhere that depends on it. As best as I can tell, the private function `isSelection()` is a clever way of imitating the Abstract Type pattern common in languages like Java. It doesn't really fit in Go. I don't think `IsSelection()` should be part of the public API but `type Selection` needs to be, otherwise we cannot represent all ASTs using the types package.

An alternative to this approach that I'd be interested to hear your thoughts on is to do away with this Abstract Type and do something like:

```
type Selection struct {
  Field *types.Field
  FragmentSpread *types.FragmentSpread
  InlineFragment *types.InlineFragment
}
```

where only one of the fields is set based on what the concrete type of Selection is.

This is more inline with the rest of the code in this repo, for example see `ExecutableDefinition` for a simpler way of representing the one-or-the-other pattern. The downside is that code that handles a Selection will need to use a chain of if statements instead of a `switch` case. Another downside is that it's not clear that a Selection will have exactly one of its fields set.

types/directive.go
```diff
@@ -0,0 +1,31 @@
...
+type Directive struct {
+	Name      Ident
+	Arguments ArgumentList
+}
+
+// DirectiveDefinition is a representation of the GraphQL DirectiveDefinition.
+//
+// http://spec.graphql.org/draft/#sec-Type-System.Directives
+type DirectiveDefinition struct {
+	Name      string
```
ssko1, line 15: we should probably be consistent with our usage of `Ident`. Some types have `Ident` while others are `string`s. All type definition names should use the `Ident` type so that its location can be exposed.

internal/query/query.go
pavelnikolov: Just one question re:
> Rather than add more things to Document and refactor everything to work based on that root type
Why do we need to add everything and why is refactoring required? Do you mean adding `ExecutableDefinition` under document?

types/directive.go
pavelnikolov: Can you give an example, please?

types/field.go
```diff
@@ -0,0 +1,37 @@
+package types
+
+// FieldDefinition is a representation of a GraphQL FieldDefinition.
+//
+// http://spec.graphql.org/draft/#FieldDefinition
+type FieldDefinition struct {
+	Alias      Ident
```
ssko1, line 7: FieldDefinitions should not have an Alias.

internal/query/query.go
rudle: > Why do we need to add everything and why is refactoring required? Do you mean adding ExecutableDefinition under document?

I think we have some crossed wires, so I'm going to try and re-establish my line of thinking.

It's my belief that a type called `types.Document` should match the specification for a Document as closely as is reasonable. The type called `Document` in the query package did not match the specification, as it was missing 2/3 of the fields. http://spec.graphql.org/June2018/

The type now called `types.ExecutableDefinition` matches the relevant part of the specification. It's clear to everyone what it does and which part of the specification it implements.

An orthogonal option available to us in this PR (or a followup) is to introduce this type:

```go
// http://spec.graphql.org/June2018/#sec-Language.Document
type Document struct {
  ExecutableDefinitions []ExecutableDefinition
  TypeSystemDefinitions []TypeSystemDefinition
  TypeSystemExtension []TypeSystemExtension
}
```

and

```
// http://spec.graphql.org/June2018/#TypeSystemDefinition
type TypeSystemDefinition struct {
  SchemaDefinition SchemaDefinition
  TypeDefinition       []TypeDefinition
  DirectiveDefinition []DirectiveDefinition
}
```

and so on.

The main work of this approach would be to replace references to `types.Schema` with `types.Document` or `types.TypeSystemDefinition`. Given how much code depends on `types.Schema` and its fields, I don't find this option very appealing but it is the most correct way to proceed. I'm interested to hear your thoughts on this.

types/field.go
rudle: good catch. fixed in 6537f29

rudle:
I found another issue and pushed a fix. https://github.com/graph-gophers/graphql-go/pull/437/commits/ab449f07e831915a3ce3a1a3c83c27f2eb504082 describes the issue

rudle:
Hi @pavelnikolov, there are still a couple of outstanding discussions on this PR that I'd like to get resolved soon. Thanks!

On our side, we have successfully integrated our GraphQL schema linter with this branch of the repo which, in my opinion, supports this as a good solution for allowing tooling to be built on top of the parser.

pavelnikolov:
@rudle which are the outstanding discussions? I'd be happy to help. Feel free to ask questions proactively and ping me if I don't respond promptly.

types/query.go
pavelnikolov: What is the common thing between the `Field`, `FragmentSpread` and `InlineFragment` that you want to use? Can't we come up with some reasonable single-method interface?

rudle: They don't have any behavior in common, they are different types. The reason they are related is because together they are types of `Selection`, which is defined in the specification here: http://spec.graphql.org/June2018/#Selection. Once they are inside of this interface, code can use a switch statement to unpack the abstract types into concrete types. [Here is an example](https://github.com/rudle/graphql-go/blob/types-leaf-package/internal/validation/validation.go#L247) of how they are used.

I understand that in Go, we often use an interface to abstract _behavior_ (ex. `Writer`, `Reader`) which is maybe where your question about what they have in common is coming from. This code, in my opinion, is a misuse of Go interfaces but it is also this way [on the master branch](https://github.com/graph-gophers/graphql-go/blob/master/internal/exec/selected/selected.go#L45-L47) and there isn't another obvious way to express this type of abstract type.

internal/query/query.go
rudle: @pavelnikolov any thoughts here? I feel pretty strongly that we shouldn't have a type called `Document` in the public API that doesn't match the specification.

rudle:
Hi @pavelnikolov there are two discussions that I think are blocking this PR from being merged to master.

- [ ] https://github.com/graph-gophers/graphql-go/pull/437#discussion_r596867675 this discussion about the Selection interface
- [ ] https://github.com/graph-gophers/graphql-go/pull/437#discussion_r603671832 this discussion about `types.Document` vs `types.ExecutableDefinition`

If you'd prefer to discuss this sort of thing in high-bandwidth & real-time, I'd be happy to schedule a video chat at a time that works for you.

internal/query/query.go
pavelnikolov: I agree with you. Your reason to use `types.ExecutableDefinition` sounds pretty compelling. Let's proceed with executable definition and come back to Document in a followup PR.

pavelnikolov:
@rudle thank you for your patience. I'm looking forward to a quick chat about the Selection interface. You can find me in he Kubernetes or Gophers Slack with the same username.

types/query.go
pavelnikolov: > I understand that in Go, we often use an interface to abstract behavior (ex. Writer, Reader) which is maybe where your question about what they have in common is coming from.
Yes, this is exactly where the question is coming from.

> is a misuse of Go interfaces
I wouldn't call it a misuse, but I wish there were a more elegant solution to this.

pavelnikolov: > [Here is an example](https://github.com/rudle/graphql-go/blob/types-leaf-package/internal/validation/validation.go#L247) of how they are used.
@rudle I reviewed your example and I think that this code can be significantly improved. It can have a method that returns the flattened selection. Then we can loop through all selections and append them to a slice. That way we'll use an interface with a common method and we'll not have to typecast. They do have something in common and we better use it properly to avoid the casting.

rudle: Hi @pavelnikolov is this what you were describing? This function could be used as an interface for Selections. We could improve on the names, this is just a proof of concept.

```
func (f *Field) FlatSelections(exDef ExecutableDefinition) []*Field {
	return []*Field{f}
}

func (f *FragmentSpread) FlatSelections(exDef ExecutableDefinition) []*Field {
	sels := exDef.Fragments.Get(f.Name.Name).Fragment.Selections
	flatFields := make([]*Field, 0, len(sels))
	for _, s := range sels {
		flatFields = append(flatFields, s.FlatSelections(exDef)...)
	}
	return flatFields
}

func (f *InlineFragment) FlatSelections(exDef ExecutableDefinition) []*Field {
	sels := f.Selections
	flatFields := make([]*Field, 0, len(sels))
	for _, s := range sels {
		flatFields = append(flatFields, s.FlatSelections(exDef)...)
	}
	return flatFields
}
```
After writing this function, I looked for places that took in `types.Selection` as an argument. I wasn't able to find any places where the need for type-casting would be removed by having this FlatSelection function available. The one place that this function would maybe help is in [validateMaxDepth](https://github.com/rudle/graphql-go/blob/types-leaf-package/internal/validation/validation.go#L238) if we added error handling for the case of nil `FragmentSpread`.

I'm not convinced that this complicated code simplifies the rest of the library enough to include it.

Issue #432, opened 2021-02-06
graphql.Time unmarshal time check


EchoUtopia:
i use graphq.Time as a insert data field, when the front end makes a mistake, and pass the unix time in millisecond, the unmarshaled time will be tens thousands years in the future, like: `55055-03-02 08:00:00+00`. in time.MarshalJSON, the max year is limited to 9999.

so it's a good idea to check the time max limit when parsing

panikgaul:
@EchoUtopia I think this issue can be closed, because the check are already implemented?

I tested it now:

> err := t.UnmarshalGraphQL("55055-03-01T15:04:05Z")
>    if err != nil {
>       fmt.Println(err)
>    }

`parsing time "55055-03-01T15:04:05Z" as "2006-01-02T15:04:05Z07:00": cannot parse "5-03-01T15:04:05Z" as "-"`

EchoUtopia:
@panikgaul sorry that i did not describe clearly.

the function `UnmarshalGraphQL` also accepts `int64`, `int32`, `float64` input type, so when the input is a unexpected large number, graphql-go will assume it's a normal time.

for example: 

```
	gt := &graphql.Time{}
	err := gt.UnmarshalGraphQL(time.Now().UnixNano())
	if err != nil {
		fmt.Println(err)
	}
	fmt.Println(gt)

```

pavelnikolov:
I verified several other unix timestamp converters and they all behave the same way. Since it is supposed to count number of seconds since Jan 1 1970 it is totally normal that the year is in tens of thousands for large numbers. We don't know how people use this library so we do not want to limit them just to year 9999. It would be up to each developer to make sure that they validate both on the client and server sides.

panikgaul:
I tried to fix this issue here: https://github.com/graph-gophers/graphql-go/pull/438

EchoUtopia:
@pavelnikolov that make sense,  but in the other way, it's still a little weird, successfully unmarshaled time can not be marshaled, it's not consistent

PR #438, opened 2021-03-22
Fix issue graphql.Time unmarshal time check #432

panikgaul:
Now we can use both unix timestamps:

```
t.UnmarshalGraphQL(time.Now().UnixNano())
t.UnmarshalGraphQL(time.Now().Unix())
```

### DIFF
--- a/time.go
+++ b/time.go
@@ -35,7 +35,13 @@ func (t *Time) UnmarshalGraphQL(input interface{}) error {
 		t.Time = time.Unix(int64(input), 0)
 		return nil
 	case int64:
-		t.Time = time.Unix(input, 0)
+		if input >= 1e10 {
+			sec := input / 1e9
+			nsec := input - (sec * 1e9)
+			t.Time = time.Unix(sec, nsec)
+		} else {
+			t.Time = time.Unix(input, 0)
+		}
 		return nil
 	case float64:
 		t.Time = time.Unix(int64(input), 0)
--- a/time_test.go
+++ b/time_test.go
@@ -1,0 +1,50 @@
+package graphql
+
+import (
+	"testing"
+	"time"
+)
+
+var timeTestCases = []interface{}{
+	time.Now(),
+	int32(1),
+	time.Now().Unix(),
+	time.Now().UnixNano(),
+	float64(-1),
+	"2006-01-02T15:04:05Z",
+}
+
+func TestImplementsGraphQLType(t *testing.T) {
+	time := Time{}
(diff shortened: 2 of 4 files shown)


time.go
```diff
@@ -35,7 +35,13 @@ func (t *Time) UnmarshalGraphQL(input interface{}) error {
 		t.Time = time.Unix(int64(input), 0)
 		return nil
 	case int64:
-		t.Time = time.Unix(input, 0)
+		if input >= 1e10 {
```
pavelnikolov, line 38: I wonder if you need the the `if` statement at all... If `input` is smaller than `1e10` then `nsec` will be `0` anyway, no?

pavelnikolov:
Can you, please, add unit tests and I'd be happy to merge this PR

panikgaul:
> Can you, please, add unit tests and I'd be happy to merge this PR. Feel free to create a `time_test.go` file in the root folder.

Of course, I added a couple of tests, the coverage of time_test.go should be 100% now.

time_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"time"
+)
+
+var timeTestCases = []interface{}{
+	time.Now(),
+	int32(1),
+	time.Now().Unix(),
+	time.Now().UnixNano(),
+	float64(-1),
+	"2006-01-02T15:04:05Z",
```
pavelnikolov, line 14: Could you, pelase, check the expected values too? So that each test has `got`, `want` pair. This way you will not only check that marshalling and unmarshalling work but also that they return what you expect.

pavelnikolov (changes requested):

time_test.go
pavelnikolov: Also, I wouldn't use `time.Now()` instead I prefer constants, please

pavelnikolov:
@panikgaul could you, please, check the results too in the tests?

time_test.go
panikgaul: Now I added the checks for expeced values with assertion library and remove time.Now() usage.
Can you please check if it is ok for you, Thx!

pavelnikolov:
I'm ready to merge this PR, but I don't want `testify` as part of the library's dependencies. You only use two of it's methods, so it doesn't really bring any significant value. Thank you for your understanding and I'm sorry for delaying the merge.

pavelnikolov:
@panikgaul are your intentions to remove testify and continue working on this PR?

pavelnikolov:
#486 solves this problem. Thank you

PR #442, opened 2021-03-25
Add exported IsNullable method to nullable types

Kalbi:
I was very happy to see https://github.com/graph-gophers/graphql-go/pull/430 merged and am now using it. However, in my project, I felt that having an exported `IsNullable` method would be useful, especially when implementing `Nullable` custom scalar types.

This will allow one to verify that custom `Nullable` implementations correctly implement the `NullUnmarshaller` interface.

It also allows us to verify the type of the input field for things like gRPC field masks.

E.g., previously, if one had a traditional nullable field, we would not be able to generate gRPC field masks accurately for partial updates because if a field was `nil` we would not add the field mask. Now, users can check if a field `IsNullable`, and if it is, check if the `Set` field is `true` or `false`, in order to set a field mask accurately.

Hopefully this method is safe to export; it would be helpful for our project and I hope it will be useful to others who want to take advantage of the new Nullable functionality.

### DIFF
--- a/nullable_types.go
+++ b/nullable_types.go
@@ -2,6 +2,9 @@ package graphql
 
 import (
 	"fmt"
+	"reflect"
+
+	"github.com/graph-gophers/graphql-go/internal/exec/packer"
 )
 
 // NullString is a string that can be null. Use it in input structs to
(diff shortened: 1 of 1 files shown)


pavelnikolov:
Hi @Kalbi,
all these types already have a public `Nullable() {}` method. This means that you can create an interface that has this method and test for yourself if the type implements it. I am not convinced that we need to expose an additional static method `IsNullable`. This doesn't add any functionality that is not already available. At the cost of one additional interface, you can check for yourself.

Issue #151, opened 2018-01-14
Ability to add our own directives


renoirb:
May be related to #116.

## CONTEXT

Author of this request is proficient JavaScript, but not in Go. Some statements in this issue may skewed perceptions based as a total beginner in Go.

## REQUEST/QUESTION

Assuming we have a schema that supports `Time` scalar (like you've done in [TestTime](https://github.com/neelance/graphql-go/blob/master/graphql_test.go#L1475)), and we want to use a custom formatter using GraphQL directives.

**Is there plans for our own `neelance/graphql-go` GraphQL server implementation to create our own directives?**

For example, we could get a `n days ago`, by defining our own `@date(...)` directive;

```
query {
  ago: addHour("2001-02-03T04:05:06Z") @date(as: "days ago")
}
// => { "ago": "6189 days ago" }
```

I'm aware not all types of Directives would be useful to ALL packages extending `neelance/graphql-go`.

It wouldn't make sense to have *scalar Time* exported by default.

But what I enjoyed in `tests/`, [and the file `time.go`](https://github.com/neelance/graphql-go/blob/master/time.go), was that I could see how I could implement my own scalar.

That would be awesome to have another example where one could see how to implement properly his own directive (e.g. `@date`).

### OBSERVATIONS

As said in #116, it looks like the place we could append our own directives would be implemented are hidden away. I assume it's because `lowerCased` functions aren't exported in Go.

I see in [internal/exec/selected/selected.go at line 194](https://github.com/neelance/graphql-go/blob/master/internal/exec/selected/selected.go#L194) how you’ve implemented `@skip` directive

And also how (black magic?) you’ve implemented `@onQuery` at [internal/tests/testdata/tests.json `{"schemas": []}`](https://github.com/neelance/graphql-go/blob/master/internal/tests/testdata/tests.json#L3), and ran from [internal/tests/all_test.go **TestAll**](https://github.com/neelance/graphql-go/blob/master/internal/tests/all_test.go#L26)

### LINKS

Maybe you'd like to see some directives implemented in JavaScript, here is one implementation sample I found useful; [*lirown/graphql-custom-directives* Date Formatting Directives](https://github.com/lirown/graphql-custom-directives#date-formatting-directives)

renoirb:
I've forked the project and started experimenting on how to make it happen

See [**renoirb/graphql-go**, at branch **issue-151**](https://github.com/renoirb/graphql-go/)

[...]

PR #446, opened 2021-04-08
Added ability to use your own directives usage

eko:
Hi,

## Description

This PR goal is to bring ability to use so custom directives, as mentioned in issue https://github.com/graph-gophers/graphql-go/issues/151.

I would be interested in having feedbacks, mainly on the way that I am currently re-using the `internal/common` package exported to be used publicly in `pkg/common`: maybe I could just copy objects I need (but I need a lot of these defined in package ;-)).

## Example usage

Assuming that I have the following schema:

```graphql
directive @customDirective(
	customAttribute: String!
) on FIELD_DEFINITION

schema {
	query: Query
}

type Query {
	myTestQuery: String! @customDirective(customAttribute: hi)
}
```

I can use the directive in my resolver by adding a new argument:

```go
func (r *customDirectiveResolver) MyTestQuery(ctx context.Context, directives common.DirectiveList) string {
	customDirective := directives.Get("customDirective")
	customAttribute, _ := customDirective.Args.Get("customAttribute")

	return fmt.Sprintf(
		"Hello custom directive '%s' with attribute value '%s'!",
		customDirective.Name.Name,
		customAttribute.String(),
	)
}
```

Thank you!

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/graph-gophers/graphql-go/gqltesting"
 	"github.com/graph-gophers/graphql-go/introspection"
 	"github.com/graph-gophers/graphql-go/trace"
+	"github.com/graph-gophers/graphql-go/types"
 )
 
 type helloWorldResolver1 struct{}
@@ -46,6 +47,27 @@ func (r *helloSnakeResolver2) HelloHTML(ctx context.Context) (string, error) {
 
 func (r *helloSnakeResolver2) SayHello(ctx context.Context, args struct{ FullName string }) (string, error) {
 	return "Hello " + args.FullName + "!", nil
+}
+
+type customDirectiveVisitor struct {
+	beforeWasCalled bool
+}
+
+func (v *customDirectiveVisitor) Before(ctx context.Context, directive *types.Directive, input interface{}) error {
+	v.beforeWasCalled = true
+	return nil
+}
+
+func (v *customDirectiveVisitor) After(ctx context.Context, directive *types.Directive, output interface{}) (interface{}, error) {
+	if v.beforeWasCalled == false {
+		return nil, errors.New("Before directive visitor method wasn't called.")
+	}
+
+	if value, ok := directive.Arguments.Get("customAttribute"); ok {
+		return fmt.Sprintf("Directive '%s' (with arg '%s') modified result: %s", directive.Name.Name, value.String(), output.(string)), nil
+	} else {
+		return fmt.Sprintf("Directive '%s' modified result: %s", directive.Name.Name, output.(string)), nil
+	}
 }
 
 type theNumberResolver struct {
@@ -191,7 +213,6 @@ func TestHelloWorld(t *testing.T) {
 				}
 			`,
 		},
-
 		{
 			Schema: graphql.MustParseSchema(`
 				schema {
@@ -210,6 +231,67 @@ func TestHelloWorld(t *testing.T) {
 			ExpectedResult: `
 				{
 					"hello": "Hello world!"
+				}
+			`,
+		},
+	})
+}
+
+func TestCustomDirective(t *testing.T) {
+	t.Parallel()
+
+	gqltesting.RunTests(t, []*gqltesting.Test{
+		{
+			Schema: graphql.MustParseSchema(`
+				directive @customDirective on FIELD_DEFINITION
+
+				schema {
+					query: Query
+				}
+
+				type Query {
+					hello_html: String! @customDirective
+				}
+			`, &helloSnakeResolver1{},
+				graphql.DirectiveVisitors(map[string]types.DirectiveVisitor{
+					"customDirective": &customDirectiveVisitor{},
+				})),
+			Query: `
+				{
+					hello_html
+				}
+			`,
+			ExpectedResult: `
+				{
+					"hello_html": "Directive 'customDirective' modified result: Hello snake!"
+				}
+			`,
+		},
+		{
+			Schema: graphql.MustParseSchema(`
+				directive @customDirective(
+					customAttribute: String!
+			    ) on FIELD_DEFINITION
+
+				schema {
+					query: Query
+				}
+
+				type Query {
+					say_hello(full_name: String!): String! @customDirective(customAttribute: hi)
+				}
+			`, &helloSnakeResolver1{},
+				graphql.DirectiveVisitors(map[string]types.DirectiveVisitor{
+					"customDirective": &customDirectiveVisitor{},
+				})),
+			Query: `
+				{
+					say_hello(full_name: "Johnny")
+				}
+			`,
+			ExpectedResult: `
+				{
+					"say_hello": "Directive 'customDirective' (with arg 'hi') modified result: Hello Johnny!"
 				}
 			`,
 		},
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -25,6 +25,7 @@ type Request struct {
 	Logger                   log.Logger
 	PanicHandler             errors.PanicHandler
 	SubscribeResolverTimeout time.Duration
+	Visitors                 map[string]types.DirectiveVisitor
 }
 
 func (r *Request) handlePanic(ctx context.Context) {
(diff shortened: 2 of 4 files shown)


pavelnikolov:
Thank you for your contribution. I really want custom directives to be supported in the library! There are few things to discuss, though.
1. Does your implementation support only directives on `FIELD_DEFINITION`s? Do you support directives on objects, fragments and inline fragments? 
2. Regarding the private types that are made public - take a look at this PR #437 it already exports some types and it will be merged soon.
3. The final thing that I want to get resolved before merging this is refactor the signature of the resolvers. (Opt-in and backwards compatible, of course). Now we support no args, ctx and  ctx + args. With this implementation we now have one more optional argument - directives. But in another PR that I want to get in soon(ish) is selected fields that allows people to prefetch the data of fields beforehand. So I was thinking about an optional signature that has a `Request` argument, for example: 
```go
func (r *Resolver) Something(ctx context.Context, req *graphql.Request) (*SomethingResponse, error) {
    // req contains input args, directives and selected fields. Potentially we can add more things in future without changing the signature.
}
```

eko:
@pavelnikolov,

Thanks for your quick & clear reply!

1. You're right, actually it only supports `FIELD_DEFINITION`, maybe we could add support for `OBJECT` and `FRAGMENT` too in a second time? I've tried to take a look and I don't think I am too smart with reflection to do this actually, or maybe you can point me to the right direction?

2. Thanks for pointing me to this PR #437, this is exactly what I need so I will wait for it to be merged and will use it afterwards.

3. I understand your arguments and what you want but I find it pity to loose the typing of the arguments because we will retrieve a `common.Request` object so users will have to do casting like this in order to retrieve their arguments and we will not benefit of schema validation anymore?

```go
type PaginationArg struct {
    Offset int32
}

func (r *Resolver) Something(ctx context.Context, req *graphql.Request) (*SomethingResponse, error) {
    myDirective := req.Directives.Get("myCustomDirective") // No problem here, directives are graphql-go types

    myPaginationArg, ok := req.Args.Pagination.(*PaginationArg)
    if ok {
        offset := myPaginationArg.Offset
    }
}
```

Anyway, maybe it's acceptable, I will try to implement this in the coming days and will keep you informed.

eko:
@pavelnikolov After thinking about your suggestions, a colleague @iotafr 👋 point me to what Apollo is doing and I think this is a good way: https://www.apollographql.com/docs/apollo-server/schema/creating-directives/

I've updated the PR to use this kind of Visitor pattern. I think it takes the following advantages:
* We don't touch to the current resolvers objects,
* We can easily create a single directive visitor to act on multiple fields,
* We could easily support objects & fragments directives (I will try to take a look at this now that I have something working).

Directive visitor must implement the following interface and can act before the resolver to be called or after:

```go
type DirectiveVisitor interface {
	Before(directive *Directive, input interface{}) error
	After(directive *Directive, output interface{}) (interface{}, error)
}
```

What do you think? I've kept the old previous logic on a branch in case you don't like this one ;)

pkg/common/directive.go
```diff
@@ -1,5 +1,10 @@
 package common
 
+type DirectiveVisitor interface {
```
iotafr, line 3: Maybe in some cases we need to access to the context at the directive level. What do you think to add `ctx context.Context` in `After` and `Before` definitions ?

eko: You're right, I just added the context

pavelnikolov: One such case would be with a caching directive, since caching info can be stored in the context.

pavelnikolov:
@eko I really like the visitor pattern!

pavelnikolov:
@eko  #437 has been merged

eko:
@pavelnikolov Great thanks! I've just updated the PR with types exported by #437.

I've tried to take a look at how I can retrieve the object directives but without success actually, I've to spend much time to understand the internal workflow.

pavelnikolov:
There are some conflicts. @eko can I help you to get this working?

eko:
Hi @pavelnikolov,

Sorry for the delay, I was off for a couple of days.

I just rebased the pull request in case you want to merge the first part. Unfortunately, I will not have any much time to work on other directives implementations.

Thank you

graphql_test.go
```diff
@@ -4113,7 +4195,9 @@ func TestTracer(t *testing.T) {
 		"id": "1002",
 	}
 
-	_ = schema.Exec(ctx, doc, opName, variables)
+	visitors := map[string]types.DirectiveVisitor{}
+
+	_ = schema.Exec(ctx, doc, opName, variables, visitors)
```
pavelnikolov, line 4200: I do believe that directive visitors should be passed at schema parse time and need to be a member of the schema itself and not members of the requests. IMHO, the resolvers do not need to be a member of the `Request` struct. and here you have access to the schema too.

pavelnikolov:
@eko thank you very much for rebasing this code. I left one comment re where to pass the directive visitors. I like this pattern but it will require some more work.
Regarding your comment about "on other directives implementations" what do you mean with "other"? Do you mean object, interface, union and argument directives (anything other than fields)?

eko:
Hello @pavelnikolov,

Thank you, you're right, this is so much better to have them at the schema parse section, I just updated the PR in order to add a new `DirectiveVisitors()` Schema option to pass them.

Concerning the other directives implementations: yes, this implementation works with FIELD_DEFINITION type. That's a first step but other types have to also be implemented later.

pavelnikolov:
@eko OK, this is much better now that the directive visitors are not passed with `Exec`. I was thinking about adding `DirectiveVisitors()` method to the root resolver, but maybe your idea is better to be passed as a schema option. I like it.

eko:
I also found it was interesting to have them declared as schema options but let me know if you prefer to have them on root resolver.

pavelnikolov:
Merci beaucoup, @eko! I was thinking the same thing. Then I thought to myself "What if there is a resolver field named 'DirectiveVisotrs'". I know that this is highly unlikely. The notion of Schema options is that they are optional and we can turn the options on and off. However, in the case of directives, is it possible that they are included in the schema, but not included in the options? Then what do we do? Do we error out or do we ignore them? If we ignore them, some people might be confused in the other case other people will get confused. If we leave them in the resovler, how do we pass them around so that the method for directives doesn't interfere with resolver fields/methods? 🤔

pavelnikolov:
These before/after visitors are only going to be invoked if the fields have method resolvers. How about field-resolvers?
https://github.com/graph-gophers/graphql-go/pull/446/files#diff-87ea4cfcdd6bb8062908727b36ef05d305ddfbd5ed89322e2276f315ac63dd1bR204

internal/exec/exec.go
```diff
@@ -208,8 +209,47 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
...
+
+						if visitorErr := visitor.Before(ctx, directive, values); err != nil {
+							err := errors.Errorf("%s", visitorErr)
+							err.Path = path.toSlice()
+							err.ResolverError = visitorErr
+							return err
+						}
+					}
+				}
+			}
```
pavelnikolov, line 230: These visitors will only be invoked if we use method resolvers. How about when we use struct fields as resolvers?

pavelnikolov: I'm talking about [lines 263 to 269 below](https://github.com/graph-gophers/graphql-go/pull/446/files#diff-87ea4cfcdd6bb8062908727b36ef05d305ddfbd5ed89322e2276f315ac63dd1bR263-R269)

internal/exec/exec.go
```diff
@@ -208,8 +209,47 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
...
+			// Before hook directive visitor
+			if len(f.field.Directives) > 0 {
+				for _, directive := range f.field.Directives {
+					if visitor, ok := r.Visitors[directive.Name.Name]; ok {
+						var values = make([]interface{}, 0)
+						for _, inValue := range in {
+							values = append(values, inValue.Interface())
+						}
+
+						if visitorErr := visitor.Before(ctx, directive, values); err != nil {
```
fajrifernanda, line 222: this should be 
`if visitorErr := visitor.Before(ctx, directive, values); visitorErr != nil {`

internal/exec/exec.go
```diff
@@ -208,8 +209,47 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
...
+			// Call method
 			callOut := res.Method(f.field.MethodIndex).Call(in)
 			result = callOut[0]
+
+			// After hook directive visitor (when no error is returned from resolver)
+			if !f.field.HasError && len(f.field.Directives) > 0 {
+				for _, directive := range f.field.Directives {
+					if visitor, ok := r.Visitors[directive.Name.Name]; ok {
+						returned, visitorErr := visitor.After(ctx, directive, result.Interface())
+						if err != nil {
```
fajrifernanda, line 241: this should be `if visitorErr != nil`

fajrifernanda (changes requested):

pavelnikolov:
I am closing this PR in favor of https://github.com/graph-gophers/graphql-go/pull/543

PR #452, opened 2021-04-29
Fix dir in readme

jinleileiking:


pavelnikolov:
Thank you @jinleileiking

PR #453, opened 2021-04-29
Coerce float64 to int32 in `NullInt` and vice versa in `NullFloat`

Kalbi:
I have been using the new `NullInt` type in one of my projects and received an error `wrong type for Int: float64`.

According to the GraphQL spec (http://spec.graphql.org/June2018/#sec-Scalars) and the rest of the code that coerces a `float64` type to `int32`, we should also be doing it for the built-in `NullInt` and `NullFloat`. 

We know that when Go unmarshals JSON (https://golang.org/pkg/encoding/json/#Unmarshal), that any JSON number is stored in a `float64`, so it is definitely necessary for `NullInt`. Therefore, the `NullFloat` changes made here are probably unnecessary.

I think this makes sense; please let me know if you have any feedback.

### DIFF
--- a/nullable_types.go
+++ b/nullable_types.go
@@ -2,6 +2,7 @@ package graphql
 
 import (
 	"fmt"
+	"math"
 )
 
 // NullString is a string that can be null. Use it in input structs to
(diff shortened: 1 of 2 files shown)


Kalbi:
I am unsure why `TestSchemaSubscribe_CustomResolverTimeout` is failing in the CI build, I don't seem to have touched anything related to that at all?

Issue #444, opened 2021-03-31
Add an errors.Location field to all GraphQL type definitions


ssko1:
# Problem Statement

For some context on the problem statement, please refer to the following comment!

https://github.com/graph-gophers/graphql-go/pull/437#issuecomment-802130304
> We wrote a linter to catch common errors made by contributors to our shared codebase. It checks for things like:
> 
> * do mutations and their inputs have similar names?
> * does the type and its fields repeat words (eg avoid: `FriendRequests { friendRequestID }`)
> * is pagination implemented correctly
>   and so on.

One of the goals of our linter is to show meaningful error messages to our contributors whenever a GraphQL schema change produces errors. An important component of a good error message is the source location of the problem. The source location allows contributors to quickly identify issues in their schema so they can correct it. Additionally, IDEs can parse the source location to provide quick links to the affected code.

Currently, GraphQL type definitions ([`Object`](https://github.com/graph-gophers/graphql-go/blob/master/internal/schema/schema.go#L87) for example) have no fields that provide the location of the type in the source schema.

# Recommended Approach
All GraphQL type definitions should provide a `Loc errors.Location` field.

Examples:
```diff
type DirectiveDecl struct {
	Name string
	Desc string
	Locs []string
	Args common.InputValueList
+       Loc errors.Location
}
```

```diff
type Object struct {
	Name       string
	Interfaces []*Interface
	Fields     FieldList
	Desc       string
	Directives common.DirectiveList

	interfaceNames []string
+       Loc errors.Location
}
```

### replace `Name: string` with `Name: Ident`
`Ident` is primarily used in query fields, e.g. [Field](https://github.com/graph-gophers/graphql-go/pull/437/files#diff-c98ebe423d42119dc93aeddb287ff4c4082eff72280c9c6a5d0a0efeaf556585L72-L73) and [Fragment Spreads](https://github.com/graph-gophers/graphql-go/pull/437/files#diff-c98ebe423d42119dc93aeddb287ff4c4082eff72280c9c6a5d0a0efeaf556585L87).  

We could extend this functionality and replace type names with `Ident` instead. 

However, this change introduces a significant amount of stuttering, particularly where the string variant of name is needed (`field.Name.Name`, `directive.Name.Name`). This is especially true in the construction of maps: `map[string]SomeType`.

## Alternative approaches
I currently see no alternatives to providing location other than rolling out a completely separate parser that provides the location for GraphQL type definitions.

PR #454, opened 2021-04-29
Add location fields to type definitions

ssko1:
Addresses #444 

# Overview
This change adds a `Loc: errors.Location` field to all definition structs defined in the `types` package. These struct fields are populated when the schema is parsed; the `Line` and `Column` fields in `errors.Location` represents the first character of the *identifier* for that definition.

For example:
```graphql
directive @foo ON FIELD_DEFINITION # line: 1 column: 12
type Foo { }                       # line: 2 column: 6
input FooInput { }                 # line: 3 column: 7
# ...etc
```

This change strictly adds the `Loc: errors.Location` field to definitions that don't already have it. Definitions that already have location fields such as [`InputValueDefinition`](https://github.com/graph-gophers/graphql-go/blob/8a96404e06f14f259c40f43b4ebf1864565e) were not touched.

## Motiviation and Examples
Tools built against a schema will most certainly need line/column source mapping as an additional form of identification when the `name` of a definition isn't enough. 

For a specific example: as explained in #444, location data is important in our usage of graphql-go because a majority of schema validation occurs after parsing the schema (i.e. the schema syntax is valid, but there are other issues that aren't caught by the parser). Without location data on definitions, we won't be able to surface the lines and columns of domain-specific schema issues in our error messages. 

Below are a few examples that I hope can showcase the importance of location data in definitions

### Loc on ObjectTypeDefinition
Given the syntax-valid schema below:
```graphql
type uncapitalizedtypename { 
  a: String
}
```

The `Loc` field in `ObjectTypeDefinition` will allow us to produce the following error:
```
schema.graphql:1:6: uncapitalizedtypename: Type names must be TitleCase.
```

### Loc on FieldDefinition
```graphql
type Foo {
  accountVerified: Boolean
}
```

The `Loc` field in `FieldDefinition` will allow us to produce the following error:
```
schema.graphql:2:3: accountVerified: Boolean field names should be prefixed by "is", "can", ...
```

### DIFF
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -348,9 +348,10 @@ func parseSchema(s *types.Schema, l *common.Lexer) {
 			s.Types[input.Name] = input
 
 		case "scalar":
+			loc := l.Location()
 			name := l.ConsumeIdent()
 			directives := common.ParseDirectives(l)
-			s.Types[name] = &types.ScalarTypeDefinition{Name: name, Desc: desc, Directives: directives}
+			s.Types[name] = &types.ScalarTypeDefinition{Name: name, Desc: desc, Directives: directives, Loc: loc}
 
 		case "directive":
 			directive := parseDirectiveDef(l)
@@ -368,7 +369,7 @@ func parseSchema(s *types.Schema, l *common.Lexer) {
 }
 
 func parseObjectDef(l *common.Lexer) *types.ObjectTypeDefinition {
-	object := &types.ObjectTypeDefinition{Name: l.ConsumeIdent()}
+	object := &types.ObjectTypeDefinition{Loc: l.Location(), Name: l.ConsumeIdent()}
 
 	for {
 		if l.Peek() == '{' {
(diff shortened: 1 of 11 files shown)


pavelnikolov:
Thank you for your contribution. The only thing that is kind of annoying is that the location is in the `errors` package but we already had it there for some types and not for all of them.

ssko1:
@pavelnikolov Thank you for the quick merge!

I'm open to moving `errors.Location` somewhere else as its function is now no longer specifically tied to producing GraphQL errors. I'm hoping that, with the recent move towards semantic versioning, breaking changes such as moving the `errors.Location` type  is less impactful and easier to reason about. 

Could you explain a little more on your thoughts on how semantic versioning would work in this project? Would the `master` branch be open to experimental and/or frequent breaking changes?

pavelnikolov:
Yes, master branch is open to experiments. Breaking changes are never a nice thing - we aim to avid them.

PR #462, opened 2021-06-10
internal/exec: assign parent type name to __typename fields

### LINKED ISSUE #461: trace: field __typename field loses parent typeName in TraceField
Given an implementation of `trace.Tracer` that logs out the parent type name and field type name

```go
type MyTracer struct {}

func (t MyTracer) TraceQuery(ctx context.Context, _, _ string, _ map[string]interface{}, _ map[string]*introspection.Type) (context.Context, trace.TraceQueryFinishFunc)
) {
  // no-op
  return ctx, func([]*errors.QueryError) {}
}

func (t MyTracer) TraceField(ctx context.Context, label, typeName, fieldName string, trivial bool, args map[string]interface{}) (context.Context, trace.TraceFieldFinishFunc) {
  fmt.Printf("type: %q, field: %q\n", typeName, fieldName)
  return ctx, func(*errors.QueryError) {}
}
```

and schema

```graphql
type User {
  id: ID!
}

type Query {
  user(name: String!): User
}
```

and the operation

```graphql
{ 
  user(name: "tonyghita") {
    __typename
    id
  }
}
```

I would expect to see the following output when the query operation is executed
```
type: "Query", field: "user"
type: "User", field: "__typename"
type: "User", field: "id"
```

Instead, I see

```
type: "Query", field: "user"
type: "", field: "__typename"
type: "User", field: "id"
```

The `__typename` field loses its associated parent type.

tonyghita:
This change modifies `internal/exec` to attach the parent type name when a `__typename` field is executed so that it is available to any `trace.Tracer` implementations.

Fixes #461.

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"testing"
 	"time"
 
@@ -11,6 +12,8 @@ import (
 	gqlerrors "github.com/graph-gophers/graphql-go/errors"
 	"github.com/graph-gophers/graphql-go/example/starwars"
 	"github.com/graph-gophers/graphql-go/gqltesting"
+	"github.com/graph-gophers/graphql-go/introspection"
+	"github.com/graph-gophers/graphql-go/trace"
 )
 
 type helloWorldResolver1 struct{}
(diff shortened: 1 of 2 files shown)


pavelnikolov:
lgtm

Issue #292, opened 2018-11-20
Query variables are not validated


wiktor256:
There seems to be a bug in schema validation of query variables.
For this schema:

```
input SearchFilter {
  required: String!
  optional: String
}

type SearchResults {
  match: String
}

type Query {
  search(filter: SearchFilter!): [SearchResults!]!
}
```

the following query correctly fails schema validation:
```
 query  {
   search(filter: {}) {
     match
   }
} 
```
with the error message:
```
{
  "errors": [
    {
      "message": "Argument \"filter\" has invalid value {}.\nIn field \"required\": Expected \"String!\", found null.",
      "locations": [
        {
          "line": 2,
          "column": 17
        }
      ]
    }
  ]
}
```

but the same query using variables doesn't fail at all:

```
query s($filter:SearchFilter!) {
  search(filter: $filter){
    match
  }
}

variables:
{
  "filter": {}
}
```

pavelnikolov:
Hi @wiktor256, thank you for reporting this issue!
PRs are welcome!

xiaocode:
I found the same bug too.+1.

suessflorian:
I'll give this one a crack 👍

suessflorian:
Interesting, I'm having struggle to reproduce this example as I was trying to write tests to reproduce this issue - then went to a minimal server example;

```go
package main

import (
	"context"
	"log"
	"net/http"

	"github.com/graph-gophers/graphql-go"
	"github.com/graph-gophers/graphql-go/relay"
)

type queryVarResolver struct{}
type filterArgs struct {
	Required string
	Optional *string
}
type filterSearchResults struct {
	Match *string
}

func (r *queryVarResolver) Search(ctx context.Context, args *struct{ Filter filterArgs }) []filterSearchResults {
	return []filterSearchResults{}
}

func main() {
	schema := graphql.MustParseSchema(`
        input SearchFilter {
          required: String!
          optional: String
        }

        type SearchResults {
          match: String
        }

        type Query {
          search(filter: SearchFilter!): [SearchResults!]!
        }
      `, &queryVarResolver{}, graphql.UseFieldResolvers())
	http.Handle("/query", &relay.Handler{Schema: schema})
	log.Fatal(http.ListenAndServe(":8080", nil))
}
```

Mod file;
```go
module test

go 1.16

require (
	github.com/graph-gophers/graphql-go v1.2.0
	github.com/stretchr/testify v1.7.0 // indirect
)
```

[...]

PR #470, opened 2021-10-02
Tests showing query variables are validated correctly

suessflorian:
Closes https://github.com/graph-gophers/graphql-go/issues/292


pavelnikolov:
Thank you, @suessflorian

PR #471, opened 2021-10-08
Support "Interfaces Implementing Interfaces"

suessflorian:
Continues work done in https://github.com/graph-gophers/graphql-go/pull/436. Implement https://spec.graphql.org/draft/#sec-Interfaces.Interfaces-Implementing-Interfaces

Key differences:
- Avoids changing the core `Schema` type, sticks to using `Schema.Types` as a means of tracking interface validation.
- Single loop validation of all interface types.
- Improved test coverage.

### DIFF
--- a/internal/schema/schema_test.go
+++ b/internal/schema/schema_test.go
@@ -875,7 +875,7 @@ Second line of the description.
 }
 
 func TestInterfaceImplementsInterface(t *testing.T) {
-	for _, test := range []struct {
+	for _, tt := range []struct {
 		name                  string
 		sdl                   string
 		useStringDescriptions bool
@@ -890,8 +890,8 @@ func TestInterfaceImplementsInterface(t *testing.T) {
 			}
 			interface Bar implements Foo {
 				field: String!
-      }
-      `,
+			}
+			`,
 			validateSchema: func(s *types.Schema) error {
 				const implementedInterfaceName = "Bar"
 				typ, ok := s.Types[implementedInterfaceName].(*types.InterfaceTypeDefinition)
@@ -922,7 +922,7 @@ func TestInterfaceImplementsInterface(t *testing.T) {
 		{
 			name: "Parses interface transitively implementing an interface that implements an interface",
 			sdl: `
-      interface Foo {
+			interface Foo {
 				field: String!
 			}
 			interface Bar implements Foo {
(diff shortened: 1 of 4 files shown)


internal/schema/schema_test.go
```diff
@@ -873,3 +873,137 @@ Second line of the description.
 		})
 	}
 }
+
+func TestInterfaceImplementsInterface(t *testing.T) {
+	for _, test := range []struct {
+		name                  string
+		sdl                   string
+		useStringDescriptions bool
```
pavelnikolov, line 881: It seems that `useStringDescriptions` is not really used in your tests, so you could just hard-code the value below. It's just noise here.

internal/schema/schema_test.go
```diff
@@ -873,3 +873,137 @@ Second line of the description.
...
+					return fmt.Errorf("interface %q not found", implementingInterfaceName)
+				}
+
+				return nil
+			},
+		},
+		{
+			name: "Parses interface transitively implementing an interface that implements an interface",
+			sdl: `
+      interface Foo {
```
pavelnikolov, line 925: (nitpick) alignment

internal/schema/schema_test.go
```diff
@@ -873,3 +873,137 @@ Second line of the description.
...
+					return fmt.Errorf("second interface %q not found", secondImplementingInterfaceName)
+				}
+
+				return nil
+			},
+		},
+		{
+			name: "Transitively implemented interfaces must also be defined on an implementing type or interface",
+			sdl: `
+		  interface A {
```
pavelnikolov, line 970: (nitpick) alignment

internal/schema/schema_test.go
```diff
@@ -873,3 +873,137 @@ Second line of the description.
 		})
 	}
 }
+
+func TestInterfaceImplementsInterface(t *testing.T) {
+	for _, test := range []struct {
+		name                  string
+		sdl                   string
+		useStringDescriptions bool
+		validateError         func(err error) error
```
pavelnikolov, line 882: Could you, please, replace 
```go
validateError func(err error) error
```

with just:
```go
expectedError string
```
This way at the end of each test we'll be comparing the received error with the expected one.

internal/schema/schema_test.go
```diff
@@ -873,3 +873,137 @@ Second line of the description.
...
+	}{
+		{
+			name: "Parses interface implementing other interface",
+			sdl: `
+			interface Foo {
+				field: String!
+			}
+			interface Bar implements Foo {
+				field: String!
+      }
```
pavelnikolov, line 893: indentation

internal/schema/schema_test.go
pavelnikolov: I just noticed that the previous tests are the same. I'll take care of this later and will leave the test for consistency the way you've written it.

pavelnikolov:
Thank you @suessflorian

PR #472, opened 2021-10-11
Added the query to context to use it for could not marshal error

SharoonAustin6696:
Added the query to context to use it for could not marshal error


pavelnikolov:
What exactly is it that you are trying to achieve with this PR? What problem are you solving here? 
Also all your import paths are messed up.

PR #477, opened 2021-12-09
internal/exec/resolvable: include struct field name in errors

agnivade:
We were only adding method name, which meant that it was taking
an empty string if the resolver was a struct field. This was
making the error messages hard to parse as the user can't know
which field has the error.

Added a check to use the correct variable.

Earlier:

```
can not use int64 as Float
        used by (*api4.channelMember).
```

Now:

```
can not use int64 as Float
        used by (*api4.channelMember).LastViewedAt
```

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -3074,6 +3074,8 @@ type inputArgumentsObjectMismatch2 struct{}
 
 type inputArgumentsObjectMismatch3 struct{}
 
+type fieldNameMismatch struct{}
+
 type helloInput struct {
 	Name string
 }
@@ -3082,7 +3084,7 @@ type helloOutput struct {
 	Name string
 }
 
-func (*helloOutput) Hello() helloOutput {
+func (*fieldNameMismatch) Hello() helloOutput {
 	return helloOutput{}
 }
 
@@ -3228,7 +3230,7 @@ func TestInputArguments_failSchemaParsing(t *testing.T) {
 		},
 		"Struct field name inclusion": {
 			Args: args{
-				Resolver: &helloOutput{},
+				Resolver: &fieldNameMismatch{},
 				Opts:     []graphql.SchemaOpt{graphql.UseFieldResolvers()},
 				Schema: `
 					type Query {
@@ -3239,7 +3241,7 @@ func TestInputArguments_failSchemaParsing(t *testing.T) {
 					}
 				`,
 			},
-			Want: want{Error: "string is not a pointer\n\tused by (graphql_test.helloOutput).Name\n\tused by (*graphql_test.helloOutput).Hello"},
+			Want: want{Error: "string is not a pointer\n\tused by (graphql_test.helloOutput).Name\n\tused by (*graphql_test.fieldNameMismatch).Hello"},
 		},
 	}
 
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -3078,6 +3078,14 @@ type helloInput struct {
 	Name string
 }
 
+type helloOutput struct {
+	Name string
+}
+
+func (*helloOutput) Hello() helloOutput {
+	return helloOutput{}
+}
+
 type helloInputMismatch struct {
 	World string
 }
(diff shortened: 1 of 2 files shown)


graphql_test.go
```diff
@@ -3217,14 +3226,29 @@ func TestInputArguments_failSchemaParsing(t *testing.T) {
...
+		"Struct field name inclusion": {
+			Args: args{
+				Resolver: &helloOutput{},
+				Opts:     []graphql.SchemaOpt{graphql.UseFieldResolvers()},
+				Schema: `
+					type Query {
+						hello(): HelloOutput!
+					}
+					type HelloOutput {
+						name: Int
```
pavelnikolov, line 3238: Why are you using an `Int` here when the type of the `Name` field is `string`? I do not understand the problem that you are trying to solve.

graphql_test.go
```diff
@@ -3217,14 +3226,29 @@ func TestInputArguments_failSchemaParsing(t *testing.T) {
 			},
 			Want: want{Error: "field \"Input\": *struct { Thing string } does not define field \"name\" (hint: missing `args struct { ... }` wrapper for field arguments, or missing field on input struct)\n\tused by (*graphql_test.inputArgumentsObjectMismatch3).Hello"},
 		},
+		"Struct field name inclusion": {
+			Args: args{
+				Resolver: &helloOutput{},
```
pavelnikolov, line 3231: According to your schema you need to have a field or a method named `Hello`, which returns an instance of `helloOutput`. Instead you are using the `helloOutput` instance as a resolver. This doesn't make sense to me.

internal/exec/resolvable/resolvable.go
```diff
@@ -256,7 +256,13 @@ func (b *execBuilder) makeObjectExec(typeName string, fields types.FieldsDefinit
...
 		fe, err := b.makeFieldExec(typeName, f, m, sf, methodIndex, fieldIndex, methodHasReceiver)
 		if err != nil {
-			return nil, fmt.Errorf("%s\n\tused by (%s).%s", err, resolverType, m.Name)
+			var resolverName string
+			if methodIndex != -1 {
+				resolverName = m.Name
+			} else {
+				resolverName = sf.Name
+			}
+			return nil, fmt.Errorf("%s\n\tused by (%s).%s", err, resolverType, resolverName)
```
pavelnikolov, line 265: Can you give me a good example before and after this change so that I can understand what exactly you are trying to achieve.

graphql_test.go
```diff
@@ -3078,6 +3078,14 @@ type helloInput struct {
 	Name string
 }
 
+type helloOutput struct {
+	Name string
+}
+
+func (*helloOutput) Hello() helloOutput {
+	return helloOutput{}
+}
```
pavelnikolov, line 3087: Why would the `helloOutput` have a `Hello()` method that returns a new instance of `helloOutput`?

internal/exec/resolvable/resolvable.go
pavelnikolov: In what case the error message would be ambiguous?

graphql_test.go
agnivade: Woops, messed up that while writing the test. Fixing it.

graphql_test.go
agnivade: I tried to explain the problem clearly in the PR message, but let me explain with an example:

Currently, when you have a type mismatch, the error message will say something like:

```
 can not use int64 as Float
        used by (*api4.channelMember).
```

Notice that there is nothing after the `.`? That is because when resolving struct methods, we don't print the field name. We only print `m.Name` always instead of `m.Name` or `sf.Name` depending on the condition. That's what this PR fixes.

Now, with the fix, it will print as:
```
can not use int64 as Float
        used by (*api4.channelMember).LastViewedAt
```

Does that help?

internal/exec/resolvable/resolvable.go
agnivade: Explained above and also added in the PR message.

agnivade:
@pavelnikolov - Thanks for the review! The test was my bad. I pushed with a separate struct. PTAL.

PR #478, opened 2021-12-10
all: fix golangci lint errors in the codebase

agnivade:
Added a base golangci-config to the codebase to get
started. Some more changes are pending, and those
checks are commented out in the config.

Once this gets merged, we can work on enabling those
gradually.

For full reference, here are the errors which were fixed:

```
example/scalar_map/types/map.go:5: File is not `gofmt`-ed with `-s` (gofmt)
type Map map[string]interface {}
internal/common/lexer_test.go:97: File is not `gofmt`-ed with `-s` (gofmt)
var multilineStringTests = []consumeTestCase {
internal/schema/schema_test.go:472: File is not `gofmt`-ed with `-s` (gofmt)
                                        "Coloured": struct{}{},
                                        "Named":    struct{}{},
                                        "Numbered": struct{}{},
relay/relay.go:43:30: unnecessary conversion (unconvert)
        return json.Unmarshal([]byte(s[i+1:]), v)
                                    ^
internal/exec/resolvable/meta.go:45:15: S1039: unnecessary use of fmt.Sprintf (gosimple)
                TraceLabel: fmt.Sprintf("GraphQL field: __typename"),
                            ^
internal/exec/resolvable/meta.go:53:15: S1039: unnecessary use of fmt.Sprintf (gosimple)
                TraceLabel: fmt.Sprintf("GraphQL field: __schema"),
                            ^
internal/exec/resolvable/meta.go:61:15: S1039: unnecessary use of fmt.Sprintf (gosimple)
                TraceLabel: fmt.Sprintf("GraphQL field: __type"),
                            ^
internal/exec/selected/selected.go:179:7: shadow: declaration of "ok" shadows declaration at line 177 (govet)
                        a, ok := e.TypeAssertions[frag.On.Name]
                           ^
internal/schema/schema_test.go:864:8: shadow: declaration of "err" shadows declaration at line 859 (govet)
                                if err := test.validateError(err); err != nil {
                                   ^
graphql.go:135:30: SA1019: trace.ValidationTracer is deprecated: use ValidationTracerContext. (staticcheck)
func ValidationTracer(tracer trace.ValidationTracer) SchemaOpt {
                             ^
graphql.go:300:9: SA1019: trace.ValidationTracer is deprecated: use ValidationTracerContext. (staticcheck)
        tracer trace.ValidationTracer
               ^
internal/query/query.go:13:2: SA9004: only the first constant in this group has an explicit type (staticcheck)
        Query        types.OperationType = "QUERY"
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,36 @@
+run:
+  timeout: 5m
+
+linters-settings:
+  gofmt:
+    simplify: true
+  govet:
+    check-shadowing: true
+    enable-all: true
+    disable:
+      - fieldalignment
+      - composites # remove later
+      - deepequalerrors # remove later
+
+linters:
+  disable-all: true
+  enable:
+    - deadcode
+    - gofmt
+    - gosimple
+    - govet
+    - ineffassign
+    - exportloopref
+    - structcheck
+    - staticcheck
+    - unconvert
+    - unused
+    - varcheck
+    # - misspell add later
+    - goimports
+
+issues:
+  exclude-rules:
+    - linters:
+      - unused
+      path: "graphql_test.go"
--- a/graphql.go
+++ b/graphql.go
@@ -132,7 +132,7 @@ func Tracer(tracer trace.Tracer) SchemaOpt {
 
 // ValidationTracer is used to trace validation errors. It defaults to trace.NoopValidationTracer.
 // Deprecated: context is needed to support tracing correctly. Use a Tracer which implements trace.ValidationTracerContext.
-func ValidationTracer(tracer trace.ValidationTracer) SchemaOpt {
+func ValidationTracer(tracer trace.ValidationTracer) SchemaOpt { //nolint:staticcheck
 	return func(s *Schema) {
 		s.validationTracer = &validationBridgingTracer{tracer: tracer}
 	}
@@ -297,7 +297,7 @@ func (s *Schema) validateSchema() error {
 }
 
 type validationBridgingTracer struct {
-	tracer trace.ValidationTracer
+	tracer trace.ValidationTracer //nolint:staticcheck
 }
 
 func (t *validationBridgingTracer) TraceValidation(context.Context) trace.TraceValidationFinishFunc {
--- a/example/scalar_map/server.go
+++ b/example/scalar_map/server.go
@@ -10,19 +10,17 @@ import (
 	"github.com/graph-gophers/graphql-go/relay"
 )
 
-
 type Args struct {
 	Name string
 	Data types.Map
 }
 
-
 type mutation struct{}
 
-func (_ *mutation) Hello(args Args) string { 
+func (_ *mutation) Hello(args Args) string {
 
 	fmt.Println(args)
-	
+
 	return "Args accept!"
 }
 
--- a/example/scalar_map/types/map.go
+++ b/example/scalar_map/types/map.go
@@ -2,10 +2,10 @@ package types
 
 import "fmt"
 
-type Map map[string]interface {}
+type Map map[string]interface{}
 
 func (Map) ImplementsGraphQLType(name string) bool {
-	return name == "Map" 
+	return name == "Map"
 }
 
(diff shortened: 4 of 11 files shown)


agnivade:
@pavelnikolov - I don't have access to semaphoreCI. Would you be able to add the golangci step in that? Or alternatively, we can setup GitHub actions to allow contributors to add CI steps themselves.

Let me know. Thanks :)

.golangci.yml
```diff
@@ -0,0 +1,36 @@
+run:
```
tonyghita, line 1: How did you arrive at this configuration?

.golangci.yml
```diff
@@ -0,0 +1,36 @@
...
+    - structcheck
+    - staticcheck
+    - unconvert
+    - unused
+    - varcheck
+    # - misspell add later
+    - goimports
+
+issues:
+  exclude-rules:
```
tonyghita, line 33: Why exclude these rules?

tonyghita (commented):
This seems like a good improvement. I left a few questions about how the lint configuration was chosen.

tonyghita:
I'm looking into updating SemaphoreCI. It seems like they've changed quite a bit. They also only seem to support extremely old versions of Go, at least in the version of the product that this project is currently usig.

graphql.go
```diff
@@ -132,7 +132,7 @@ func Tracer(tracer trace.Tracer) SchemaOpt {
 
 // ValidationTracer is used to trace validation errors. It defaults to trace.NoopValidationTracer.
 // Deprecated: context is needed to support tracing correctly. Use a Tracer which implements trace.ValidationTracerContext.
-func ValidationTracer(tracer trace.ValidationTracer) SchemaOpt {
+func ValidationTracer(tracer trace.ValidationTracer) SchemaOpt { //nolint:staticcheck
```
pavelnikolov, line 135: Why ignoring `statickcheck`? What rule does this func violate?

.golangci.yml
```diff
@@ -0,0 +1,36 @@
...
+    - linters:
+      - unused
```
pavelnikolov, line 35: I guess that if something is unused we have to remove it, right?

graphql.go
agnivade: I mentioned this in the PR message:

```
graphql.go:135:30: SA1019: trace.ValidationTracer is deprecated: use ValidationTracerContext. (staticcheck)
func ValidationTracer(tracer trace.ValidationTracer) SchemaOpt {
                             ^
graphql.go:300:9: SA1019: trace.ValidationTracer is deprecated: use ValidationTracerContext. (staticcheck)
        tracer trace.ValidationTracer
```

This function is deprecated, so we need to add the exception.

agnivade:
@tonyghita - Thanks for the review. Let me answer your questions here:

> How did you arrive at this configuration?

We use golangci heavily at Mattermost, and I just lifted our default [configuration](https://github.com/mattermost/mattermost-server/blob/master/.golangci.yml) and made some adjustments.

@pavelnikolov 

> I guess that if something is unused we have to remove it, right?

> Why exclude these rules?

Because the struct is actually passed as a resolver to the schema. So the resolver method needs to be there for the schema validation to work. Nothing calls the method, but that's expected.

pavelnikolov:
> Because the struct is actually passed as a resolver to the schema. So the resolver method needs to be there for the schema validation to work. Nothing calls the method, but that's expected.

But then you can only add `//nolint:` comment to the method and remove the `exclude-rules` section from the config, right?

agnivade:
> But then you can only add //nolint: comment to the method and remove the exclude-rules section from the config, right?

Yes, of course. But then, that file is meant to have methods like this and will have more in future. So it will be filled with `//nolint:` comments. That's why I chose the file route to exclude the file altogether.

pavelnikolov:
Ok, that makes sense. Then I'll merge this PR since it looks like a good start and we'll improve it in future.

agnivade:
Lol, too fast :smile: We should have added it to SemaphoreCI first. I'll send another PR :)

PR #481, opened 2021-12-14
Improve Sempahore CI

agnivade:
- Added separate block to run golangci
- Removed `export GO111MODULE=on`. This is unnecessary now when we have a go.mod in the repo.
- Removed `export GOPATH=~/go`. Setting GOPATH is not necessary for modern Go projects.
- Removed `go get ./...`. Test and Build will both download dependencies automatically.
- Configured go test and go build to happen in parallel.


agnivade:
@tonyghita - Can you check why the CI is not being run?

relay/relay.go
```diff
@@ -24,7 +24,7 @@ func UnmarshalKind(id graphql.ID) string {
 	if err != nil {
 		return ""
 	}
-	i := strings.IndexByte(string(s), ':')
+	i := strings.IndexByte(string(s),  ':')
```
pavelnikolov, line 27: Accidental space here?

agnivade: It's intended. I am testing the CI. That's why it's at a draft stage :slightly_smiling_face:

tonyghita:
@agnivade the SemaphoreCI configuration was not set to run for forked PRs. I enabled them for a small allow list of contributor names. I _think_ the next commit pushed to this branch should trigger a CI run.

.semaphore/semaphore.yml
```diff
@@ -1,21 +1,38 @@
...
+  - name: Style Check
+    task:
+      jobs:
+        - name: fmt
+          commands:
+            - sem-version go 1.17
+            - checkout
+            # Recommended way to install golang-ci
+            # https://golangci-lint.run/usage/install/#other-ci.
+            - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.42.1
```
tonyghita, line 17: One thing I always do with these kinds of `curl | sh` scripts is add them to the repository so that the script contents are version controlled. Otherwise you never know what's on the other side of that remote call.

pavelnikolov: I'd be happy to vendor the script, although, this particular script I also tend to download in some pipelines

agnivade: Yep, fully agree. Will vendor. :+1:

relay/relay.go
pavelnikolov:
```
golangci-lint run ./...Failed in 
00:07
relay/relay.go:27: File is not `gofmt`-ed with `-s` (gofmt)00:07
	i := strings.IndexByte(string(s),  ':')00:07
internal/exec/subscribe.go:85:5: reflectvaluecompare: avoid using == with reflect.Value (govet)00:07
	if result == reflect.Zero(result.Type()) {00:07
	   ^00:07
graphql.go:200:5: reflectvaluecompare: avoid using == with reflect.Value (govet)00:07
	if s.res.Resolver == (reflect.Value{}) {00:07
	   ^00:07
subscriptions.go:24:5: reflectvaluecompare: avoid using == with reflect.Value (govet)00:07
	if s.res.Resolver == (reflect.Value{}) {00:07
	   ^00:07
internal/schema/schema_test.go:998:8: shadow: declaration of "err" shadows declaration at line 993 (govet)00:07
				if err := tt.validateError(err); err != nil {
```

The above is caught and there are some other issues too. I guess you have to fix it now.

agnivade:
@tonyghita / @pavelnikolov - PTAL. I had to fix some more issues because my local golangci was on a previous version, and the later version had some more checks. You can see this issue (https://github.com/golang/go/issues/43993) to understand the changes I made.

CI works now. Verified manually. Test and build stages happen in parallel now.

relay/relay.go
agnivade: Yep, fixed those as well. My local tool was on a previous version.

PR #493, opened 2022-01-20
Add OpenTelemetry Support

steve-gray:
Adds OpenTelemetry support. The old one, shortsightedly, lives in /trace and if it's moved out to a sub-package it'll break existing code/users - but at least this one's in an opentel package. Have mapped relevant values and statuses to span attributes and error status/codes, so there's some conceptual differences versus opentracing version, but these seem to be the relevant constructs to use.

Feel free to tweak. Am using this verbatim in my own code but would happily just switch to a mainline version. 

Works beautifully.

image


trace/opentel/opentelemetry_tracer.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+
+	"github.com/graph-gophers/graphql-go/errors"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+
+	"github.com/graph-gophers/graphql-go/introspection"
+	gogqltrace "github.com/graph-gophers/graphql-go/trace"
```
pavelnikolov, line 14: I'd rather reave this package as `trace` and name the other `oteltrace`. Most devs might be used with `trace` being part of OTEL, however, in the context of this library I would prefer if we didn't use `gogqltrace`. Also we use the `oteltrace` in only one place and the other one in four places.

pavelnikolov:
@steve-gray thank you very much for this contribution!

steve-gray:
@pavelnikolov before it gets tagged for release, maybe change the package name in my file to `oteltrace` to align with the folder? Not near a computer, but super-quick improvement that'll make gopls/other developer tools have an easier time with it. The `package trace` was just because that was what it was called in my local hacked up version temporarily.

pavelnikolov:
I hadn't noticed that. I'll change it. Thank you.

PR #494, opened 2022-01-20
Add maximum query length schema option

pavelnikolov:
Allow maximum query length schema option to prevent parsing too long queries.


graphql.go
```diff
@@ -122,6 +123,13 @@ func MaxParallelism(n int) SchemaOpt {
 	}
 }
 
+// MaxQueryLength specifies the maximum allowed query length in bytes. The default is 0 which disables max length checking.
+func MaxQueryLength(n int) SchemaOpt {
```
agnivade, line 127: Nit: `MaxQueryLengthBytes` might be more readable to the user to avoid having to check what is the unit.

agnivade:
Hey @pavelnikolov - just checking on this one. Let's get this over the finish line?

PR #495, opened 2022-01-21
Fix the OTEL tracer package name

pavelnikolov:
There were two trace packages. Now both the OpenTracing and the OpenTelemetry tracers are in the same `trace` package.

### DIFF
--- a/trace/otel_trace.go
+++ b/trace/otel_trace.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/graph-gophers/graphql-go/errors"
 	"github.com/graph-gophers/graphql-go/introspection"
-	"github.com/graph-gophers/graphql-go/trace"
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
@@ -15,7 +14,7 @@ import (
 )
 
 // DefaultOpenTelemetryTracer creates a tracer using a default name
-func DefaultOpenTelemetryTracer() trace.Tracer {
+func DefaultOpenTelemetryTracer() Tracer {
 	return &OpenTelemetryTracer{
 		Tracer: otel.Tracer("graphql-go"),
 	}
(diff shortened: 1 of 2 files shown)


steve-gray:
What I'd really meant was:

It's bad that the OpenTracing one is in "trace" at all. It forces people to pull a tree dependency on opentracing-go, which is going to go away as OpenTelemetry ramps up. These two tracers should *both* be in their own packages, but the old one has to stay where it is in case anyone referenced it (until you hit 2.x.y on versioning, then can break it, unless you want to relax the Semver rules).

What I'd really suggest is:

- Near term - Old tracer in "trace" new one in "oteltrace" (so you only pick up a vendor/dep of that if you're using opentelemetry).
- Near Future - Migrate older tracer to new package and leave only the interfaces in "trace".
- Distant future - Migrate both out to graph-gophers/contrib-tracing-x packages

Nothing irks me more than pulling a dep on a range of possible platforms and then only one of them's getting used. Even today people who aren't using any tracing are picking up OpenTracing packages in their tree.

pavelnikolov:
I was thinking about using the OpenTelemetry in a different package that implements the interface. I agree with you and this is the reason I haven't cut the next release. Maybe we should start moving them to a new packages as of now and the OpenTelemetry one should be in a new package.

pavelnikolov:
This library doesn't need to have this baked in. It's worth adding it as an example only and have the two implementations in a separate package.

steve-gray:
It wouldn't download both if it was split into 3 packages (trace, graph-gophers/tracing-opentelemetry and graph-gophers/tracing-opentracing). Code that only references graphql-go would only download that package.

The real question is do you want to wear the "breaking change" for moving the OpenTracing one out of the main folder. If you can do that, then you could just go balls-out and do the two sub-projects for the tracers and be done with the problem forevermore.

pavelnikolov:
I'm not willing to introduce a breaking change without a new major version. This is not planned for anytime soon. So at the moment I can move the OTel one into a new subpackage. I'd personally remove both of them from this project in a new major version and add them as an example.

steve-gray:
So, new one and old one both to packages, but leave copy of old one where it is with a deprecation notice?

pavelnikolov:
I can move the OTEL one like this https://github.com/graph-gophers/graphql-go/pull/496 (this is your original suggestion I beleive). However, I cannot introduce a breaking change, although, I agree with you that the two implementations need to be in two separate sub packages.

pavelnikolov:
Actually, the `opentracing-go` library is deprecated now. They have left this notice on their repo:
> OpenTracing API for Go. 🛑 This library is DEPRECATED!

This means that indeed I can add the deprecation notice and leave them both in the same package. Then later completely delete the OpenTracing one or move it to an example or to a separate package. The thing is that there are many organisations which are still using the OpenTracing package and haven't switched to OTEL.

PR #496, opened 2022-01-24
Fix otel trace package

pavelnikolov:
Move OTEL tracer to a new sub package. This would prevent people from downloading it as a dependency unless they have explicitly used it in their project.


pavelnikolov:
https://github.com/graph-gophers/graphql-go/pull/513

PR #498, opened 2022-02-01
Split tracing packages

pavelnikolov:
This PR:
* Creates a new `tracer` package
* Deprecates the old `trace` package
* Moves the `opentracing` and `otel` implementations of the tracer to their own packages
* Moves the dependencies into their corresponding submodules leaving the root library free of external dependencies if the user doesn't use any OpenTracing or OpenTelemetry.
* The default tracer is `tracer.Noop`


maoueh:
Is this a new `tracer` API that can have some breaking changes? Asking because I tweaked a bit the old one with breaking changes that I used in our fork for easier tracing of what's happening inside the library.

Here the interface as part of my PR:

```
type TraceRequestFinishFunc func([]*errors.QueryError)
type TraceOperationFinishFunc func()
type TraceFieldFinishFunc func(*errors.QueryError)

type Tracer interface {
	TraceRequest(ctx context.Context, queryString string, opType string, operationName string, variables map[string]interface{}, varTypes map[string]*introspection.Type) (context.Context, TraceRequestFinishFunc)

	TraceQuery(ctx context.Context, root selected.Field) (context.Context, TraceOperationFinishFunc)
	TraceMutation(ctx context.Context, root selected.Field) (context.Context, TraceOperationFinishFunc)
	TraceSubscription(ctx context.Context, root selected.Field) (context.Context, TraceOperationFinishFunc)

	TraceField(ctx context.Context, label, typeName, fieldName string, trivial bool, args map[string]interface{}) (context.Context, TraceFieldFinishFunc)
}
```

Here the commit message I worth:

```
Added TraceSubscription to Tracer interface and re-worked the interface

This is a breaking change, but I think the previous interface had flawed like limited
support for root operation, name was not totally right since `TraceQuery` was picking
up `mutation` change also and no support to make some analysis on the selection of field.

The new interface makes a distinction between a Request and actual root "operation". In a
request, the new tracing format supports the following nesting:

    Request (query)
     Root query (type, alias)
       Field1
       Field2
     Root query (type, alias)
       Field1
       Field2
    End

The `TraceRequest` could be renamed `TraceOperation` with a specific kind, but `TraceRequest`
felt better.

This PR is dependent on having the actual selection of fields, so we can actually extract fields information from selection for effective tracing.
```

I would like to get your opinion on some changes here, it improved our tracing quite well and I would like to see that included somehow.

maoueh:
I'm talking specifically about the new `Tracer` package here, I never open a PR to update the old exactly because I knew it was a breaking change and would not be accepted.

Now that we have a new package, I would like to update it to provide a better interface.

>  Is there a way to add your improvements to the library and not to the tracer

Not sure I understand your comment.

maoueh:
Here what I see as improvements that were added in our forked version:

### Ability to determine the type actual type of the operation: query, mutation, subscription

The current `TypeQuery` is equivalent to the new proposed `TraceRequest` here. This stays the same but new tracing element are then added `TraceQuery`, `TraceMutation` and `TraceSubscription`. 

By having that, it's now possible to have find grained metrics per type like how many active subscriptions there is currently. 

In addition, those new three elements received the `root selected.Field` value which enables them to extract the actual name of what is going to be performed, super useful to better understand what is happening within a request.

### Multi root tracer

If a request is performing multiple top-level root operations like this:

```
    Request (query)
     Root resolveTimestamp (type, alias)
       Field1
       Field2
     Root resolveTimestamp (type, alias)
       Field1
       Field2
    End
```

Now it's possible to create a tracing span for each root which will usually map to some a particular resolver, this again enable better tracking of what happen as well as grouping field trace later can be achieved since their context is tied to the correct span.

A part from the interface breaking change to add new tracing method for the "root" operation I'm talking here, the rest is mostly if not identical to before, if people still just want to trace the top-level request, they can simply no-op on `TraceQuery`, `TraceMutation` and `TraceSubscription` and it becomes roughly the same as before.

For those that would like augmented metrics about what kind of operations is performed and finer grain tracing, they can implement the extra three methods and do something.

For sake of completeness, here is our tracer implementation with the new interface:

```
package dgraphql

import (
	"context"
	"fmt"
	"strings"
	"time"

	"github.com/streamingfast/dtracing"
	"github.com/graph-gophers/graphql-go/introspection"
	"github.com/graph-gophers/graphql-go/selected"
	gqtrace "github.com/graph-gophers/graphql-go/trace"
	"github.com/streamingfast/dgraphql/metrics"
	"go.opencensus.io/trace"
)

type Tracer struct {
}

func (t *Tracer) TraceRequest(
	ctx context.Context,
	queryString string,
	opType string,
	operationName string,
	variables map[string]interface{},
	varTypes map[string]*introspection.Type,
) (context.Context, gqtrace.TraceRequestFinishFunc) {
	metrics.InflightRequestCount.Inc()
	metrics.TotalRequestCount.Inc()

	spanCtx, span := dtracing.StartSpan(ctx, "GraphQL Request",
		"graphql.operationType", opType,
		"graphql.variablesCount", len(variables),
	)

	return spanCtx, func(errs []*errors.QueryError) {
		if len(errs) > 0 {
			setSpanError(span, errs)
		}

		span.End()
		metrics.InflightRequestCount.Dec()
	}
}

func (*Tracer) TraceQuery(ctx context.Context, root selected.Field) (context.Context, gqtrace.TraceOperationFinishFunc) {
	fieldName := root.Identifier()
	if fieldName == "__schema" {
		// It's an introspection schema request, let's ignore them
		return ctx, func() {}
	}

	startTime := time.Now()
	metrics.InflightQueryCount.Inc(fieldName)
	metrics.TotalQueryCount.Inc(fieldName)

	spanCtx, span := dtracing.StartSpan(ctx, "GraphQL Query",
		"graphql.query.type", fieldName,
	)

	return spanCtx, func() {
		span.End()

		metrics.InflightQueryCount.Dec(fieldName)
		metrics.QueryResponseTimes.ObserveSince(startTime, fieldName)
	}
}

func (*Tracer) TraceMutation(ctx context.Context, root selected.Field) (context.Context, gqtrace.TraceOperationFinishFunc) {
	fieldName := root.Identifier()

	startTime := time.Now()
	metrics.InflightMutationCount.Inc(fieldName)
	metrics.TotalMutationCount.Inc(fieldName)

	spanCtx, span := dtracing.StartSpan(ctx, "GraphQL Mutation",
		"graphql.mutation.type", fieldName,
	)

	return spanCtx, func() {
		span.End()

		metrics.InflightMutationCount.Dec(fieldName)
		metrics.MutationResponseTimes.ObserveSince(startTime, fieldName)
	}
}

func (*Tracer) TraceSubscription(ctx context.Context, root selected.Field) (context.Context, gqtrace.TraceOperationFinishFunc) {
	fieldName := root.Identifier()

	metrics.InflightSubscriptionCount.Inc(fieldName)
	metrics.TotalSubscriptionCount.Inc(fieldName)

	spanCtx, span := dtracing.StartSpan(ctx, "GraphQL Subscription",
		"graphql.subscription.type", fieldName,
	)

	return spanCtx, func() {
		span.End()
		metrics.InflightSubscriptionCount.Dec(fieldName)
	}
}

func (t *Tracer) TraceField(
	ctx context.Context,
	label, typeName, fieldName string,
	trivial bool,
	args map[string]interface{},
) (context.Context, gqtrace.TraceFieldFinishFunc) {
	if trivial {
		return ctx, noop
	}

	var i = 0
	keyedAttributes := make([]interface{}, len(args)*2)
	for name, value := range args {
		keyedAttributes[i] = "graphql.args." + name
		i++
		keyedAttributes[i] = sanitizeTraceAttributeValue(value)
		i++
	}

	keyedAttributes = append(keyedAttributes,
		"graphql.type", typeName,
		"graphql.field", fieldName,
	)

	spanCtx, span := dtracing.StartSpan(ctx, "GraphQL Field", keyedAttributes...)

	return spanCtx, func(err *errors.QueryError) {
		if err != nil {
			setSpanError(span, []*errors.QueryError{err})
		}

		span.End()
	}
}

func sanitizeTraceAttributeValue(value interface{}) interface{} {
	switch v := value.(type) {
	case bool, int, int8, int16, int32, int64, uintptr, uint, uint8, uint16, uint32, uint64:
		return v
	case string, fmt.Stringer:
		return v
	case map[string]interface{}:
		pairs := make([]string, len(v))

		i := 0
		for k, v := range v {
			pairs[i] = fmt.Sprintf("%s: %s", k, v)
			i++
		}

		return strings.Join(pairs, ", ")
	default:
		return fmt.Sprintf("%T", v)
	}
}

func setSpanError(span *trace.Span, errs []*errors.QueryError) {
	if len(errs) > 0 {
		msg := errs[0].Error()
		if len(errs) > 1 {
			msg += fmt.Sprintf(" (and %d more errors)", len(errs)-1)
		}

		// What's the correct way to track error in the span?
		span.SetStatus(trace.Status{
			Code:    trace.StatusCodeUnknown,
			Message: msg,
		})

		span.AddAttributes(trace.StringAttribute("graphql.error", msg))
	}
}

func noop(*errors.QueryError) {}
```

pavelnikolov:
> By having that, it's now possible to have find grained metrics per type like how many active subscriptions there is currently.

You can see the number of active subscriptions in your WebSocket implementation. 

>In addition, those new three elements received the root selected.Field value which enables them to extract the actual name of what is going to be performed, super useful to better understand what is happening within a request.

I find this redundant and I'm still not convinced that this is necessary. When you trace the rest of the request you can see which fields are selected. The trivial fields are usually not interesting and only the ones that require fetching data from some backend storage system are the ones of interest.

> Now it's possible to create a tracing span for each root which will usually map to some a particular resolver, this again enable better tracking of what happen as well as grouping field trace later can be achieved since their context is tied to the correct span.

I can't imagine yet the benefit. Can you possibly show me a trace graph from your tracer vs the ones from the library if it is not too much work to clear sensitive info. I'd be curious to see the comparison. 

As for the:
```go
	metrics.InflightRequestCount.Inc()
```
I tend to handle these in the HTTP middleware, not in the tracer of the GraphQL library because the request is "in flight" until we have written the response. In your case you decrease the in-flight requests but they are not written yet to the consumer of the API.
But again, this is your choice to do whatever you want in your tracer implementation.

I would consider adding more methods which are by default "no-op" for most people and you could do something with them with some `graphql.SchemaOpt` that allows you to add custom tracer methods.

I see that you sanitise attribute values - aren't you afraid that you might end up with PII in your traces? It is nice to have everything in the traces but some customers might want to filter out sensitive information from their traces.


Back to your original question - adding tracing methods for metrics of the queries, mutations and subscriptions in flight might be a reasonable thing to do. It is a non-breaking change and if it is a no-op for most users then I'd consider it, however, it is outside of the scope of this PR. The intention here is to make sure we deprecate the OpenTracing tracer since the project has been deprecated by the CNCF and make sure we add a default OTEL tracer implementation. Again voiding breaking changes as much as possible in the `1.x` versions.

maoueh:
> You can see the number of active subscriptions in your WebSocket implementation.

You cannot see which actual subscription it is. Seeing 1000 active subscriptions is something, but it's much more interesting to see which actual subscription it is. This is possible only if you have the right parsed selection so it's possible to extract the "root" identifier.

> I find this redundant and I'm still not convinced that this is necessary. When you trace the rest of the request you can see which fields are selected. The trivial fields are usually not interesting and only the ones that require fetching data from some backend storage system are the ones of interest.

I know you get the fully query at the request level, but is it really a good idea to actually try to extract some semantic information from there, not really, much easier to actually use the already parsed selection.  

> I can't imagine yet the benefit. Can you possibly show me a trace graph from your tracer vs the ones from the library if it is not too much work to clear sensitive info. I'd be curious to see the comparison.

I agree it's more subjective, I found it interesting to have grouping by "top-level" graphql element, mostly because the "top-level" element, at least in our case, is always the start of some external resource loading and it's the way I find to easily get the `kind` of the actual "graphql" element. 

> I can't imagine yet the benefit. Can you possibly show me a trace graph from your tracer vs the ones from the library if it is not too much work to clear sensitive info. I'd be curious to see the comparison.

I don't have them handy because we stop collecting them for now. Roughly, using StarWars graph and query:

```
query {
  heroNewHope:hero(episode: NEWHOPE) { name }
  human(id: "1000") { name }
}
```


old way was looking like:

```
request starting {"operation_name": ""}
 field starting
 field terminated
 field starting
 field terminated
request terminated
```

And with the added "top-level" element:

```
request starting {"operation_name": ""}
 query starting {"name": "human"}
  field starting
  field terminated
 query terminated
 query starting {"name": "hero"}
  field starting
  field terminated
 query terminated
request terminated
```

For field tracing with the same name but from two different "root", it was hard to see where they were coming from. 

> As for `metrics.InflightRequestCount.Inc()`

Indeed, we could argue that for "request", HTTP middleware would be better, it's was all grouped in the tracer for those actually using field name like `metrics.InflightMutationCount.Dec(fieldName)` which use the actual knowledge of the GraphQL selection.

> I see that you sanitise attribute values - aren't you afraid that you might end up with PII in your traces? It is nice to have everything in the traces but some customers might want to filter out sensitive information from their traces.

This is to query blockchain data, so everything is already public there, so it was not a concern in this case.

> Back to your original question - adding tracing methods for metrics of the queries, mutations and subscriptions in flight might be a reasonable thing to do.

Ultimately my goal is to be able to to get if it's a top-level "node" and it's kind in this case `Query`, `Mutation`, `Subscription`. It could be added differently, like modifying the `TraceField` with extra information. I could simply then make some logic based on the field.

But if I recall right, it was much easier to create new tracing method for each "root" element and pass the type. The selection of field was something I added because I had access to it in my fork, but it's not strictly required and only the kind and the field type would be enough.

> It is a non-breaking change and if it is a no-op for most users then I'd consider it, however, it is outside of the scope of this PR.

It's technically a breaking change IMO because all consumers of `graphql-go` which implement the `tracer` interface would have needed to be updated. 

> The intention here is to make sure we deprecate the OpenTracing tracer since the project has been deprecated by the CNCF and make sure we add a default OTEL tracer implementation. Again voiding breaking changes as much as possible in the 1.x versions.

I don't want to be a blocker on this PR, but I find it interesting that a new `tracer` package exists, this is for me means it's the right time to change the new `tracer` API if you accept (of course, ensuring the old `trace` one can continue working).

PR #502, opened 2022-02-26
add support for repeatable directives

speezepearson:
[The October 2021 spec](https://spec.graphql.org/October2021/#sec-Directives-Are-Unique-Per-Location) ([changelog](https://github.com/graphql/graphql-spec/blob/main/changelogs/October2021.md), [commit](https://github.com/graphql/graphql-spec/commit/be33a640a8d12eb177f7cce0c61a1fad770a3430)) introduced an optional keyword `repeatable` in directive definitions. ([The June 2018 spec](https://spec.graphql.org/June2018/#sec-Directives-Are-Unique-Per-Location) requires that all directives be unique; graphql-go doesn't enforce that, but nor does it recognize the `repeatable` keyword.)

Context: I'm working in a codebase where I want to repeat a directive, and a different tool requires that we mark it `repeatable`; but if I do that, then graphql-go SyntaxErrors.

### DIFF
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -524,7 +524,7 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
 
 	for {
 		loc := l.ConsumeIdent()
-		if ok, err := regexp.MatchString("^[A-Z_]+$", loc); err != nil || !ok {
+		if ok, err := regexp.MatchString("^[A-Z][A-Z_]+$", loc); err != nil || !ok {
 			l.SyntaxError(fmt.Sprintf("expected directive location-spec to be SNAKE_CASE, but got %q", loc))
 		}
 		d.Locations = append(d.Locations, loc)
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -514,7 +514,7 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
 
 	switch x := l.ConsumeIdent(); x {
 	case "on":
-		d.Repeatable = false
+		// no-op; Go doesn't fallthrough by default
 	case "repeatable":
 		d.Repeatable = true
 		l.ConsumeKeyword("on")
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -513,6 +513,7 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
 
 	switch x := l.ConsumeIdent(); x {
 	case "on":
+		d.Repeatable = false
 	case "repeatable":
 		d.Repeatable = true
 		l.ConsumeKeyword("on")
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -511,7 +511,13 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
 		l.ConsumeToken(')')
 	}
 
-	l.ConsumeKeyword("on")
+	switch x := l.ConsumeIdent(); x {
+	case "on":
+	case "repeatable":
+		l.ConsumeKeyword("on")
+	default:
+		l.SyntaxError(fmt.Sprintf(`unexpected %q, expecting "on" or "repeatable"`, x))
+	}
 
 	for {
 		loc := l.ConsumeIdent()
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -514,6 +514,7 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
 	switch x := l.ConsumeIdent(); x {
 	case "on":
 	case "repeatable":
+		d.Repeatable = true
 		l.ConsumeKeyword("on")
 	default:
 		l.SyntaxError(fmt.Sprintf(`unexpected %q, expecting "on" or "repeatable"`, x))
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -524,8 +524,8 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
 
 	for {
 		loc := l.ConsumeIdent()
-		if _, ok := legalDirectiveFieldLocations[loc]; !ok {
-			l.SyntaxError(fmt.Sprintf("%q is not a legal directive location (options: %v)", loc, legalDirectiveFieldLocationsSlice))
+		if _, ok := legalDirectiveLocationNames[loc]; !ok {
+			l.SyntaxError(fmt.Sprintf("%q is not a legal directive location (options: %v)", loc, legalDirectiveLocationNames))
 		}
 		if ok, err := regexp.MatchString("^[A-Z][A-Z_]*$", loc); err != nil || !ok {
 			l.SyntaxError(fmt.Sprintf("expected directive location-spec to be SNAKE_CASE, but got %q", loc))
@@ -600,7 +600,7 @@ func parseFieldsDef(l *common.Lexer) types.FieldsDefinition {
 	return fields
 }
 
-var legalDirectiveFieldLocations = map[string]struct{}{
+var legalDirectiveLocationNames = map[string]struct{}{
 	"SCHEMA":                 {},
 	"SCALAR":                 {},
 	"OBJECT":                 {},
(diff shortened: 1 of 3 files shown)


internal/schema/schema_test.go
```diff
@@ -834,6 +835,8 @@ Second line of the description.
 			}
 
 			union Union @uniondirective = Photo | Person
+
+			scalar Repeatable @repeatabledirective @repeatabledirective
```
pavelnikolov, line 839: I'd rather use some other name for this scalar as I don't want to repeat the word `Repeatable`.

internal/schema/schema.go
```diff
@@ -511,7 +511,13 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
 		l.ConsumeToken(')')
 	}
 
-	l.ConsumeKeyword("on")
+	switch x := l.ConsumeIdent(); x {
+	case "on":
+	case "repeatable":
+		l.ConsumeKeyword("on")
```
pavelnikolov, line 517: How does the above work if in your example you have two words `repeatable on`:
```
directive @repeatabledirective repeatable on SCALAR
```
 with one space between them and in the code above you consume one indent, then any of the words `on` or `repeatable`. Does this mean that you would consume the keyword `"on"` twice?

internal/schema/schema_test.go
```diff
@@ -864,6 +867,11 @@ Second line of the description.
 				if len(unionDirectives) != 1 || unionDirectives[0].Name.Name != "uniondirective" {
 					return fmt.Errorf("missing directive on Union union, expected @uniondirective but got %v", unionDirectives)
 				}
+
+				repeatableDirectives := s.Types["Repeatable"].(*types.ScalarTypeDefinition).Directives
+				if len(repeatableDirectives) != 2 || repeatableDirectives[0].Name.Name != "repeatabledirective" || repeatableDirectives[1].Name.Name != "repeatabledirective" {
+					return fmt.Errorf("missing directive on Repeatable scalar, expected @repeatabledirective @repeatabledirective but got %v", repeatableDirectives)
+				}
```
pavelnikolov, line 874: In your test you do not verify that the `Rpeatable` property of your directive is set to `true`. Please, test if it is set properly.

internal/schema/schema.go
speezepearson: Ah, yeah -- the `case "on"` is a no-op since Go doesn't fallthrough-by-default, but this is really easy to misread. I've tried to fix this by adding a comment (and tests to verify that `on on` (and `repeatable repeatable` for good measure) doesn't parse.)

internal/schema/schema_test.go
speezepearson: 😬 Thanks, yep, I was not setting that field. I wrote a separate test for this because it didn't feel like it fit in with this test, but I'll combine it if you prefer!

internal/schema/schema_test.go
speezepearson: 😛 👍

internal/schema/schema.go
```diff
@@ -511,10 +512,21 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
...
+	default:
+		l.SyntaxError(fmt.Sprintf(`unexpected %q, expecting "on" or "repeatable"`, x))
+	}
 
 	for {
 		loc := l.ConsumeIdent()
+		if ok, err := regexp.MatchString("^[A-Z_]+$", loc); err != nil || !ok {
+			l.SyntaxError(fmt.Sprintf("expected directive location-spec to be SNAKE_CASE, but got %q", loc))
+		}
```
pavelnikolov, line 529: Can this start with `_`? Also, why do we need the regex at all?

internal/schema/schema_test.go
```diff
@@ -864,6 +867,55 @@ Second line of the description.
...
+
+				massDirectives := s.Types["Mass"].(*types.ScalarTypeDefinition).Directives
+				if len(massDirectives) != 2 || massDirectives[0].Name.Name != "repeatabledirective" || massDirectives[1].Name.Name != "repeatabledirective" {
+					return fmt.Errorf("missing directive on Repeatable scalar, expected @repeatabledirective @repeatabledirective but got %v", massDirectives)
+				}
+				return nil
+			},
+		},
+		{
+			name: "Sets Directive.Repeatable iff `repeatable` keyword is given",
```
pavelnikolov, line 879: Maybe `iff` -> `if`?

internal/schema/schema.go
speezepearson: Hrrrm, yeah, it would accept `_WHATEVER` (or just `_`, for that matter). Fixed.

The two obvious approaches, to me, are "regex-match" or "enumerate legal directive locations" -- I leaned towards regex-matching so that this library won't need updating if/when the GraphQL spec adds new legal locations in future revisions. But if you'd prefer the enumeration approach, I could re-instate [this commit](https://github.com/graph-gophers/graphql-go/pull/502/commits/10dda978ed415d5017ad6a5a41b032a8c77a937f) which I ended up reverting.

pavelnikolov: I would prefer the white list, please. It is much stricter and gives a clear error message. The regex is not preferred. What is more in your example, the regex is not pre-compiled and it is going to be compiled on every directive.

pavelnikolov:
Please, remove the regex.

internal/schema/schema.go
speezepearson: All righty, sounds good! Done.

internal/schema/schema.go
```diff
@@ -511,10 +512,24 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
 		l.ConsumeToken(')')
 	}
 
-	l.ConsumeKeyword("on")
+	switch x := l.ConsumeIdent(); x {
+	case "on":
+		// no-op; Go doesn't fallthrough by default
```
pavelnikolov, line 517: Shouldn't you `falltthrough` here? What if a directive has only `on`, without `repeatable`?

internal/schema/schema.go
```diff
@@ -584,3 +599,32 @@ func parseFieldsDef(l *common.Lexer) types.FieldsDefinition {
...
+	"QUERY":                  {},
+	"MUTATION":               {},
+	"SUBSCRIPTION":           {},
+	"FIELD":                  {},
+	"FRAGMENT_DEFINITION":    {},
+	"FRAGMENT_SPREAD":        {},
+	"INLINE_FRAGMENT":        {},
+	"VARIABLE_DEFINITION":    {},
+}
+var legalDirectiveLocationNamesSlice = func() []string {
```
pavelnikolov, line 624: This variable is never used. Please, delete it!

internal/schema/schema.go
```diff
@@ -511,10 +512,24 @@ func parseDirectiveDef(l *common.Lexer) *types.DirectiveDefinition {
...
+	default:
+		l.SyntaxError(fmt.Sprintf(`unexpected %q, expecting "on" or "repeatable"`, x))
+	}
 
 	for {
 		loc := l.ConsumeIdent()
+		if _, ok := legalDirectiveLocationNames[loc]; !ok {
+			l.SyntaxError(fmt.Sprintf("%q is not a legal directive location (options: %v)", loc, legalDirectiveLocationNames))
+		}
+		if ok, err := regexp.MatchString("^[A-Z][A-Z_]*$", loc); err != nil || !ok {
```
pavelnikolov, line 530: This `if` statement is not needed, please, remove it!

speezepearson: Oops, thanks for the catch.

internal/schema/schema.go
speezepearson: `on` got consumed by the `ConsumeIdent()` on the previous line, didn't it? Looking at the implementations of `ConsumeIdent` / `ConsumeKeyword`, it looks like they both call `TokenText()` and then `ConsumeWhitespace()`, so I expect they will consume the same number of tokens; and if this code didn't consume `on`, then I would expect `on` to be interpreted as the location, which would make [this check](https://github.com/speezepearson/graphql-go/blob/05a893c8fd6e44c183b68e2d2672f35522d3ca01/internal/schema/schema.go#L526) fail in [this test](https://github.com/speezepearson/graphql-go/blob/05a893c8fd6e44c183b68e2d2672f35522d3ca01/internal/schema/schema_test.go#L880).

pavelnikolov:
Thank you for your contribution. I'll fix few more things and will release it soon in version `v1.4.0`

speezepearson:
Awesome, thanks! Pleasure doing business with you!

PR #504, opened 2022-03-01
Fix remove checkNilCase test helper function

cnnrrss:


pavelnikolov:
Thank you, this makes sense

Issue #120, opened 2017-10-30
Feature Request: Schema Stitching. Combining multiple schemas into one


arquio:
Hi,

are there any plan for implemente this feature or similar??

Thanks a lot for all yours work.

Ref: https://dev-blog.apollodata.com/graphql-schema-stitching-8af23354ac37

mikeifomin:
Nice approach for composing a large application.

jackielii:
I was looking at this just now. I have an idea:

e.g. I have `UserSchema`
```
type UserResolver struct{}
func (UserResolver) UserByID(id string) *User
```
and then `OrderSchema`
```
type OrderResolver struct{}
func (OrderResolver) OrderByID(id string) *Order
```

then the combined schema resolver I can use:
```
type RootResolver{
  UserResolver
  OrderResolver
}
```

albertorestifo:
@jackielii on top of that, schema stitching should allow for more complex merging:

If I have the following two remote GraphQL schemas:

```
type Event {
  date: String
  location: String
}


Query {
   event(date: String): Event
}
```

```
type Weather {
  temperature Int
  condition String
  windSpeed: Int
}

Query {
  forecast(location: String): Weather
}
```

I should be able to stich them toghether and perform the following query:

```
query GetEvent($date: String) {
  event(date: $date) {
    location {
      temperature
    }
  }
}
```

That's where the power of stitching comes in use

shreyansh-zazz:
Any updates on this?

mattdamon108:
Actually, I built a tool to merge and stitch modularized `*.graphql` files into one schema. I hope my tool to help you out.

https://github.com/mattdamon108/gqlmerge

obukhov:
I've made some research and found out that schema stitching is now deprecated feature of graphql-tools and is replaced with apollo federation: https://blog.apollographql.com/apollo-federation-f260cf525d21

obukhov:
So maybe now effort should be more on making build-in compatibility for federation as described here: https://www.apollographql.com/docs/apollo-server/federation/federation-spec/. 
What do you think?

jakubknejzlik:
@obukhov I definitely agree. Also as we are moving our "federated" services to golang, it would be awesome if I can help 👍

jakubknejzlik:
If anyone interested, I've found already implemented solution (for "http link" check the whole gateway repo): https://github.com/nautilus/gateway/blob/master/merge.go

[...]

PR #507, opened 2022-03-17
Apollo Federation Spec: Fetch service capabilities

aeramu:
Related to this issue #120, to integrate with Apollo Federation we need to fulfill their specification at minimal with fetch service capabilities
`Schema composition at the gateway requires having each service's schema, annotated with its federation configuration. This information is fetched from each service using _service, an enhanced introspection entry point added to the query root of each federated service.`
With these changes, we can use graphql-go as subgraph behind Apollo Federation Gateway even though it's still not fully implements the specification, but at least it's one step closer
 
This already works with simple schema stitching

### DIFF
--- a/example/apollo_federation/README.md
+++ b/example/apollo_federation/README.md
@@ -1,6 +1,6 @@
-### Apollo Federation
+# Apollo Federation
 
-A simple example of integration with apollo federation as subgraph.
+A simple example of integration with apollo federation as subgraph. Tested with Go v1.18, Node.js v16.14.2 and yarn 1.22.18.
 
 To run this server
 
@@ -13,3 +13,23 @@ To run this server
 `yarn start`
 
 and go to localhost:4000 to interact
+
+Execute the query:
+
+```
+query {
+  hello
+  hi
+}
+```
+
+and you should see a result similar to this:
+
+```json
+{
+  "data": {
+    "hello": "Hello from subgraph one!",
+    "hi": "Hi from subgraph two!"
+  }
+}
+```
--- a/example/apollo_federation/README.md
+++ b/example/apollo_federation/README.md
@@ -1,0 +1,15 @@
+### Apollo Federation
+
+A simple example of integration with apollo federation as subgraph.
+
+To run this server
+
+`go run ./example/apollo_federation/subgraph_one/server.go`
+
+`go run ./example/apollo_federation/subgraph_two/server.go`
+
+`cd example/apollo_federation/gateway`
+
+`yarn start`
+
+and go to localhost:4000 to interact
--- a/example/apollo_federation/gateway/.gitignore
+++ b/example/apollo_federation/gateway/.gitignore
@@ -1,0 +1,1 @@
+/node_modules
--- a/example/apollo_federation/gateway/index.js
+++ b/example/apollo_federation/gateway/index.js
@@ -1,0 +1,20 @@
+const { ApolloServer }  = require('apollo-server')
+const { ApolloGateway, IntrospectAndCompose } = require('@apollo/gateway');
+
+const gateway = new ApolloGateway({
+    supergraphSdl: new IntrospectAndCompose({
+        subgraphs: [
+            { name: 'one', url: 'http://localhost:4001' },
+            { name: 'two', url: 'http://localhost:4002' },
+        ],
+    }),
+});
+
+const server = new ApolloServer({
+    gateway,
+    subscriptions: false,
+});
+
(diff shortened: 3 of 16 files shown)


pavelnikolov:
@aeramu thank you for your contribution. Could you, please, add a minimalistic example with two subgraphs in the examples folder. That would be really helpful for people who are just starting with Apollo Federation and this library. 
In addition to that, there has been already work started on implementing directives https://github.com/graph-gophers/graphql-go/pull/446, which could allow for satisfying the rest of the specification. Just out of curiosity, what else is remaining?

suessflorian:
@pavelnikolov, the base introspection spec by default doesn't expose directives (motivation being, it's an implementation detail). Looks like Apollo Federation spec encourages servers to extend introspection capabilities to pseudo show these directives? In contrast to https://github.com/movio/bramble, which deals with this problem by making servers simply plain string show their schema. Both approaches seem to have flaws. But Bramble at least works to focus on compatibility.

pavelnikolov:
I thought that you'd need to use directives like: `@key, @provides and @external`.

aeramu:
@pavelnikolov I already add simple example.

The remaining main spec that I think quite complex is [resolve request for entites](https://www.apollographql.com/docs/federation/federation-spec#resolve-requests-for-entities). It's used to reference and extend types across subgraphs. I still don't know the best way to implement it because it's need user implementation to get each entities.

pavelnikolov:
@aeramu I pulled you branch locally but unfortunately I couldn't get the node server to start.
Initially it complained that I had newer version of Node.js than what Apollo supported. Then I downgraded Node to v14 LTS but now I get the following error:
```
$ yarn start
yarn run v1.22.18
warning package.json: No license field
$ node index.js
/Users/pavel/go/src/github.com/graph-gophers/graphql-go/example/apollo_federation/gateway/index.js:5
    supergraphSdl: new IntrospectAndCompose({
                   ^

TypeError: IntrospectAndCompose is not a constructor
    at Object.<anonymous> (/Users/pavel/go/src/github.com/graph-gophers/graphql-go/example/apollo_federation/gateway/index.js:5:20)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:75:12)
    at internal/main/run_main_module.js:17:47
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

It seems to me that it would be much more appropriate to have this declared in a `docker-compose.yml` file so that users would only need to run `docker compose up` in order to see the example working.
I'm sorry for demanding yet another change, but would you mind adding a `Dockerfile` in each of the 3 folders and then a `docker-compose.yaml` in the root folder of the example? This way we would ensure that this example works on everybody's machine. Let's not forget that some Go developers might not necessarily familiar with Node.js and having everything declared in a `Dockerfile` would make their life much easier.

pavelnikolov:
I did test a little bit more and it seems that the apollo/gateway version that you are using is really old. I did upgrade the module to `"^0.49.0"` and I got a different error:
```
node_modules/@apollo/gateway/dist/supergraphManagers/IntrospectAndCompose/loadServicesFromRemoteEndpoint.js:50
            throw new Error(errorMessage);
                  ^

Error: Couldn't load service definitions for "one" at http://localhost:4001: Expected JSON response body, but received:
<!DOCTYPE html>
<html>
...
```
This seemed like the GraphiQL interface HTML. That is why I changed the URLs of the two servers in order to get your example working:
```diff
const gateway = new ApolloGateway({
    supergraphSdl: new IntrospectAndCompose({
        subgraphs: [
-            { name: 'one', url: 'http://localhost:4001' },
+            { name: 'one', url: 'http://localhost:4001/query' },
-            { name: 'two', url: 'http://localhost:4002' },
+            { name: 'two', url: 'http://localhost:4002/query' },
        ],
    }),
});
```
And finally, I got your example working:
Screen Shot 2022-03-20 at 3 23 43 pm

pavelnikolov:
@aeramu I made the changes for you. Thank you for your contribution! I'm looking forward to having the rest of the Apollo Federation working properly. It seems that I need to merge a PR that improves the way we handle custom directives and focus on the `@key`, `@provides` and `@external` directives.

Issue #497, opened 2022-01-25
Unexpected panic in (*PrimitiveValue).Deserialize


jupenur:
The following query, when run against the starwars example, triggers an unexpected panic that is only recovered by `net/http`:

```
query ($x: String = -"") {
  search(text: $x) {
    __typename
  }
}
```

Full stack trace:

```
2022/01/24 10:30:58 http: panic serving [::1]:64496: invalid syntax
goroutine 23 [running]:
net/http.(*conn).serve.func1(0xc0000c2d20)
        /usr/local/go/src/net/http/server.go:1804 +0x153
panic(0x7255c0, 0xc000098010)
        /usr/local/go/src/runtime/panic.go:971 +0x499
github.com/graph-gophers/graphql-go/types.(*PrimitiveValue).Deserialize(0xc000186c60, 0x0, 0x97e700, 0x1)
        /home/user/Repositories/graphql-go/types/value.go:50 +0x259
github.com/graph-gophers/graphql-go.(*Schema).exec(0xc000109c00, 0x7f1538, 0xc0000db2c0, 0xc0001b8000, 0x338, 0x0, 0x0, 0xc000186de0, 0xc0001201c0, 0xc0000e9080)
        /home/user/Repositories/graphql-go/graphql.go:245 +0x3e5
github.com/graph-gophers/graphql-go.(*Schema).Exec(0xc000109c00, 0x7f1538, 0xc0000db2c0, 0xc0001b8000, 0x338, 0x0, 0x0, 0x0, 0x79fd18)
        /home/user/Repositories/graphql-go/graphql.go:202 +0x94
github.com/graph-gophers/graphql-go/relay.(*Handler).ServeHTTP(0xc00015c6c0, 0x7f0d80, 0xc000120620, 0xc00010e700)
        /home/user/Repositories/graphql-go/relay/relay.go:61 +0x169
net/http.(*ServeMux).ServeHTTP(0x9b29a0, 0x7f0d80, 0xc000120620, 0xc00010e700)
        /usr/local/go/src/net/http/server.go:2428 +0x1ad
net/http.serverHandler.ServeHTTP(0xc0001202a0, 0x7f0d80, 0xc000120620, 0xc00010e700)
        /usr/local/go/src/net/http/server.go:2867 +0xa3
net/http.(*conn).serve(0xc0000c2d20, 0x7f15e0, 0xc0000db000)
        /usr/local/go/src/net/http/server.go:1932 +0x8cd
[...]

PR #515, opened 2022-04-13
Adding primitive value validation

DallasP9124:
This is in response to issue #497 Unexpected panic in (*PrimitiveValue).Deserialize.

Tracing through validation showed that primitives were not being completely validated but depending on the value `Type` and not the value itself. I added in value validation to the simplest degree I could think of and included tests to demonstrate. Should no longer produce panics and return errors to the client.

### DIFF
--- a/internal/validation/testdata/tests.json
+++ b/internal/validation/testdata/tests.json
@@ -1458,6 +1458,125 @@
       "errors": []
     },
     {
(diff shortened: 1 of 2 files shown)


pavelnikolov:
@DallasP9124 thank you for your contribution

PR #516, opened 2022-05-04
Improve type assertion method argument validation (require zero)

kainosnoema:
It's tempting to include a context argument (or think it's allowed), but not discover that this will fail until a query is executed. Validating the resolver during schema parsing reduces the chance of inadvertent errors here.

We've run into this several times and even considered opening a PR to support passing `context.Context` into type assertion methods before realizing that it's better to handle loading and errors in the parent resolver instead. This validation should help avoid this confusion for others.

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -3666,6 +3666,106 @@ func TestErrorPropagation(t *testing.T) {
 	})
 }
 
+type assertionResolver struct{}
+
+func (r *assertionResolver) ToHuman() (*struct{ Name string }, bool) {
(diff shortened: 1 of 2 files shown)


pavelnikolov:
@kainosnoema could you, please, add some unit tests?

kainosnoema:
@pavelnikolov took me a bit to figure out a reasonable place to put the tests. Let me know if there's a better approach.

pavelnikolov:
@kainosnoema this is fine. Thank you!

PR #524, opened 2022-07-16
Optional support for querying fields yet unknown in the schema.

philikon:
Background: Most GraphQL servers assume that the server just needs to be
backwards-compatible, not forwards-compatible. There are scenarios, however, where
a server might be on an older software version, and therefore schema, than a client.
A distributed microservices landscape, for instance, could give rise to such a
scenario.

Now you might have a client (microservice C) querying a server (microservice S)
for a field that S does not yet know about. For instance because C was deployed
before S, or S had to be reverted to a previous version due to a bug. In such a
a case you'd want S to degrade gracefully when a query for a field comes in that
it doesn't know about. Presumably that field was added to the schema in a future
version that C has but S doesn't.

This patch adds support for just this scenario and this behavior. If
AllowUnknownFields is enabled, graphql-go will still resolve the query as best
as it can, fill in 'null' for the unknown field(s), and produce a corresponding
entry in the errors list, rather than failing the entire query altogether.

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -109,6 +109,14 @@ func UseFieldResolvers() SchemaOpt {
 	}
 }
 
+// AllowUnknownFields specifies whether a query containing unknown fields
+// is permitted or not.
+func AllowUnknownFields() SchemaOpt {
+	return func(s *Schema) {
+		s.schema.AllowUnknownFields = true
+	}
+}
+
 // MaxDepth specifies the maximum field nesting depth in a query. The default is 0 which disables max depth checking.
 func MaxDepth(n int) SchemaOpt {
 	return func(s *Schema) {
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -4460,3 +4460,87 @@ func TestQueryService(t *testing.T) {
 		},
 	})
 }
+
+func TestUnknownField(t *testing.T) {
+	schemaString := `
+	  schema {
+	    query: Query
+	  }
+
+	  type Query {
+	    hello: String!
+	  }`
+	gqltesting.RunTests(t, []*gqltesting.Test{
+		{
+			Schema: graphql.MustParseSchema(schemaString, &helloWorldResolver1{}),
+			Query: `
+				{
+					hello
+					world
+				}
+			`,
+			ExpectedErrors: []*gqlerrors.QueryError{{
+				Message:   `Cannot query field "world" on type "Query".`,
+				Locations: []gqlerrors.Location{{Column: 6, Line: 4}},
+				Rule:      "FieldsOnCorrectType",
+			}},
+		},
+		{
+			Schema: graphql.MustParseSchema(schemaString, &helloWorldResolver1{}, graphql.AllowUnknownFields()),
+			Query: `
+				{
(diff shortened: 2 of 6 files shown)


philikon:
ping?

pavelnikolov:
I'm reluctant to add a feature that is not in the official spec and nobody else has requested. In my opinion this is more a problem with the development and release process than a problem with the library. If you for example use feature flags then you can handle this business as usual. No need to rollback. You just disable the feature and return `null` for the field. If you enable the feature you return the actual data. I am closing this as it assumes how you are going to deploy the library and handles a hypothetical case which can be handled in a much more elegant way.

philikon:
I don't see how you've addressed the scenario I described, though I might be missing something. How does a feature flag help me in a distributed services landscape where one service does not yet have the latest and greatest version of the software + schema while others do? Those services running an older version of the software can't anticipate what fields might be added to the schema in the future, so there's no way to return `null` for something they can't anticipate.

philikon:
In REST or GRPC, it's easy for a new API schema to specify a new field or endpoint, and new consumers to start making optional use of it while retaining useful degraded performance if services don't yet support that field. If a client ends up making a request to an endpoint that doesn't exist yet, that specific request will fail, but they can still query other parts of the system. If the response doesn't yet contain a field recently added to the schema, the client still gets the rest of the data.

How do you envision this should work in GraphQL?

pavelnikolov:
The example you gave was:
> Now you might have a client (microservice C) querying a server (microservice S)
for a field that S does not yet know about. For instance because C was deployed
before S, or S had to be reverted to a previous version due to a bug.
The way I see this working is to always add new resolvers to your servers before you add them to the clients. Then your resolver might look something like this:

```go
func (r Resolver) SomeField(ctx context.Context, args ArgType) (*Response, error) {
    if !r.ft.Enabled("feature_name") { // once your clients are released, don't revert to a previous version! You can toggle the feature flag instead.
        return nil, nil
    }
    // calculate response
    return res, nil
}

```

philikon:
> The way I see this working is to always add new resolvers to your servers before you add them to the clients.

I explicitly listed examples where this is not possible. "Simply don't do the thing you need to do" isn't really an answer to my question.

One example I gave was that regulatory requirements or a bug might keep one of the microservices stuck at an older version of the software and thus an older version of the schema. Your suggestion would require that the older version already anticipated the new schema or a feature flag or whatever. This simply isn't possible in some environments.

REST, gRPC, and others offer workable semantics here. GraphQL through its query mechanism IMHO should have mode where it gracefully handles unknown fields. Otherwise one has to conclude that GraphQL, at least the way you're interpreting it to be, isn't suited for or meant to be used in distributed microservices environments where complex ownership, regulations, and deployment strategies can result in a non-hierarchical rollout of a schema change. In other words, atomicity cannot be guaranteed across the fleet. Is that what you're saying? I'd say that's pretty lame.

pavelnikolov:
@philikon is this implemented in the official GraphQL spec? Try adding an issue to the spec. If they accept it then, I would reconsider and add this "feature" to this library.

PR #525, opened 2022-07-18
Disallow repeat of non repeatable directives

### LINKED ISSUE #509: Return an error if a schema has repeated directives without the repeatable keyword
It was reported in https://github.com/graph-gophers/graphql-go/pull/502 that this library doesn't enforce uniqueness of non-repeatable directives. This is a bug and such a restriction needs to be enforced unless the `repeatable` keyword is present in the directive definition.

For example, if we are given two directives:

```
directive @repeatabledirective repeatable on SCALAR
directive @otherdirective on SCALAR
```
Then only the first one (i.e. `@repeatabledirective`) can be repeated multiple times on the same scalar type. If we repeat the `@otherdirective` more than once, then that should result in an error.

The issue is considered resolved when only directives declared as repeatable can be repeated.

ostrea:
Fixes #509.

### DIFF
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -289,6 +289,7 @@ func resolveField(s *types.Schema, f *types.FieldDefinition) error {
 }
 
 func resolveDirectives(s *types.Schema, directives types.DirectiveList, loc string) error {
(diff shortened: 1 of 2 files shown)


pavelnikolov:
@ostrea thank you for your contribution

PR #540, opened 2022-12-12
Fix __type queries sometimes not returning data

### LINKED ISSUE #539: __type(name:"") queries don't return data if the named type is unknown
Given the example schema:
```graphql
type Query {
  hello: String!
}
```
When you run a query for an unknown type
```graphql
{
  __type(name:"Bar") {
    kind
    name
    description
  }
}
```

I would expect to get 
```graphql
{
  "data": {
    "__type": null
  }
}
```

Instead I get 
```graphql
{
  "data": {}
}
```

simhnna:
If the queried type was not defined it would return an empty data section instead of `__type: null`

Fixes #539

### DIFF
--- a/internal/exec/selected/selected.go
+++ b/internal/exec/selected/selected.go
@@ -107,7 +107,9 @@ func applySelectionSet(r *Request, s *resolvable.Schema, e *resolvable.Object, s
 						return nil
 					}
 
+					var resolvedType *introspection.Type
 					t, ok := r.Schema.Types[v.String()]
+
 					var resolvedType *introspection.Type
 					if ok {
 						resolvedType = introspection.WrapType(t)
--- a/internal/exec/selected/selected.go
+++ b/internal/exec/selected/selected.go
@@ -109,8 +109,6 @@ func applySelectionSet(r *Request, s *resolvable.Schema, e *resolvable.Object, s
 
 					var resolvedType *introspection.Type
 					t, ok := r.Schema.Types[v.String()]
-
-					var resolvedType *introspection.Type
 					if ok {
 						resolvedType = introspection.WrapType(t)
 					}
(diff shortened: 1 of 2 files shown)


pavelnikolov:
@simhnna thank you for opening the issue and contributing with a PR. Does the response in the current version of the library actually violate the GraphQL Spec?

simhnna:
https://spec.graphql.org/October2021/#sec-Schema-Introspection
Defines `__type(name: String!): __Type`

And https://spec.graphql.org/October2021/#sec-Executing-Fields
states that 
> Each field requested in the grouped field set that is defined on the selected objectType will result in an entry in the response map.


So yes I would say the current implementation violates the spec. I know that Apollo server returns a field with null value for such queries. If you want I can double check the JavaScript reference implementation tomorrow on my work PC, but I'm pretty certain it also returns a field with a null if asked for an unknown type.

internal/exec/selected/selected.go
```diff
@@ -108,16 +108,17 @@ func applySelectionSet(r *Request, s *resolvable.Schema, e *resolvable.Object, s
 					}
 
 					t, ok := r.Schema.Types[v.String()]
```
pavelnikolov, line 110:
```suggestion
						var resolvedType *introspection.Type
						t, ok := r.Schema.Types[v.String()]

```

internal/exec/selected/selected.go
```diff
@@ -107,17 +107,20 @@ func applySelectionSet(r *Request, s *resolvable.Schema, e *resolvable.Object, s
 						return nil
 					}
 
+					var resolvedType *introspection.Type
 					t, ok := r.Schema.Types[v.String()]
-					if !ok {
-						return nil
+
+					var resolvedType *introspection.Type
```
pavelnikolov, line 113:
```suggessstion

					var resolvedType *introspection.Type
```

pavelnikolov (approved):

PR #543, opened 2022-12-22
Add Custom Directive Support for Fields

rudle:
based on graph-gophers/graphql-go#446 and work by @eko

I intend to revive the stalled conversation from #446 with this PR. I addressed the open comments on that change.


graphql_test.go
```diff
@@ -48,6 +49,27 @@ func (r *helloSnakeResolver2) SayHello(ctx context.Context, args struct{ FullNam
...
+func (v *customDirectiveVisitor) After(ctx context.Context, directive *types.Directive, output interface{}) (interface{}, error) {
+	if v.beforeWasCalled == false {
+		return nil, errors.New("Before directive visitor method wasn't called.")
+	}
+
+	if value, ok := directive.Arguments.Get("customAttribute"); ok {
+		return fmt.Sprintf("Directive '%s' (with arg '%s') modified result: %s", directive.Name.Name, value.String(), output.(string)), nil
+	} else {
+		return fmt.Sprintf("Directive '%s' modified result: %s", directive.Name.Name, output.(string)), nil
+	}
```
pavelnikolov, line 70: No need for an `else` statement here since you return inside the `if`. Please, remove the `else`.
```suggestion
	}

	return fmt.Sprintf("Directive '%s' modified result: %s", directive.Name.Name, output.(string)), nil
```

rudle:
I added a commit that addresses your comment. I will keep my changes in a separate commit and then we can squash the whole thing down when we are ready merge.

graphql_test.go
```diff
@@ -216,6 +240,158 @@ func TestHelloWorld(t *testing.T) {
 	})
 }
 
+func TestHelloWorldStructFieldResolver(t *testing.T) {
```
pavelnikolov, line 243: this doesn't call `t.Parallel()` on purpose?

graphql_test.go
```diff
@@ -216,6 +240,158 @@ func TestHelloWorld(t *testing.T) {
...
+				}
+			`,
+			ExpectedResult: `
+				{
+					"say_hello": "Directive 'customDirective' (with arg 'hi') modified result: Hello Johnny!"
+				}
+			`,
+		},
+
+		// tests for struct field resolvers
```
pavelnikolov, line 330: Why do we have this comment here? What type of test did you mean to add here?

graphql_test.go
rudle: oops. Oversight. Added to this and the other new test in this PR.

graphql_test.go
rudle: Agreed. This isn't helpful. This comment made sense when I was trying to put all the custom directive tests into this test function but decided against it.

internal/exec/exec.go
```diff
@@ -208,8 +209,48 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
...
+					if visitor, ok := r.Visitors[directive.Name.Name]; ok {
+						returned, visitorErr := visitor.After(ctx, directive, result.Interface())
+						if visitorErr != nil {
+							err := errors.Errorf("%s", visitorErr)
+							err.Path = path.toSlice()
+							err.ResolverError = visitorErr
+							return err
+						} else {
+							result = reflect.ValueOf(returned)
+						}
```
pavelnikolov, line 249:
```suggestion
						modified, visitorErr := visitor.After(ctx, directive, result.Interface())
						if visitorErr != nil {
							err := errors.Errorf("%s", visitorErr)
							err.Path = path.toSlice()
							err.ResolverError = visitorErr
							return err
						} else {
							result = reflect.ValueOf(modified)
						}
```

I'd prefer using the name `modified` to show the next person who reads this code that the directives can actually override the returned value.

internal/exec/exec.go
```diff
@@ -225,7 +266,38 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
...
```
pavelnikolov, line 297:
```suggestion
						modified, visitorErr := visitor.After(ctx, directive, result.Interface())
						if visitorErr != nil {
							err := errors.Errorf("%s", visitorErr)
							err.Path = path.toSlice()
							err.ResolverError = visitorErr
							return err
						} else {
							result = reflect.ValueOf(modified)
						}
```

types/directive.go
```diff
@@ -24,6 +28,11 @@ type DirectiveDefinition struct {
 
 type DirectiveList []*Directive
 
+type DirectiveVisitor interface {
+	Before(ctx context.Context, directive *Directive, input interface{}) error
+	After(ctx context.Context, directive *Directive, output interface{}) (interface{}, error)
+}
```
pavelnikolov, line 34: Do you think we could use generics here to make this type-safe and avoid interface casting throughout the code? 🤔 
Something like:
```suggestion
type DirectiveVisitor[I any, O any] interface {
	Before(ctx context.Context, directive *Directive, input I) error
	After(ctx context.Context, directive *Directive, output O) (O, error)
}
```

pavelnikolov: I know that this would require changes in this line:
```go
	visitors                 map[string]types.DirectiveVisitor
```
and we wouldn't be able to easily loop through all visitors, however, it would make the life of our users much better. All visitors behave the same and they differ just by the input and output types. Therefore, it'd be a nice candidate for generics. What do you think?

rudle: Sure. I will try this and see how it feels.

pavelnikolov: Yes, it is high time we updated that.

rudle: I pushed up a commit (`a7c50d4`) that tries a pretty naive approach with generics. It doesn't compile. It seems like when combining the DirectiveVisitor interface with generics, we don't gain much in the way of type safety. The interface rules in Go mean that the particular implementations must match the type signature of the interface which leads to compile errors like this.

```./graphql_test.go:290:25: cannot use &customDirectiveVisitor{} (value of type *customDirectiveVisitor) as type types.DirectiveVisitor[any, any] in map literal:
	*customDirectiveVisitor does not implement types.DirectiveVisitor[any, any] (wrong type for After method)
		have After(ctx context.Context, directive *types.Directive, output string) (string, error)
		want After(ctx context.Context, directive *types.Directive, output any) (any, error)
```

rudle: [1506179](https://github.com/graph-gophers/graphql-go/pull/543/commits/15061797b04c9e4a89b1427e81a26cd896df45da)

This latest commit adds a new file `types/directives.go` which will allow you to play around with the type system here. I don't have an intuition for Go's generics yet, so having this playground was helpful for me. I'm not sure there is a workable path to compile-time type safety because Go decided not to implement type covariance in its generics.

To be clear, the thing that's missing here is the ability to collect/abstract all Directives in a slice because we can't refer to them without specifying a type parameter. I'm going to keep working on this, just wanted to share my progress.

pavelnikolov: I'm going to sleep since it's late here. I'll think some more on this tomorrow. I was thinking that we can still use `interface{}` inside of the library to combine all the visitors in a single slice, however, this will be behind the scenes. The users of the library will only use concrete types for input and output, right? I haven't tested if this would work, though. I might check tomorrow. In any case your initial PR was fine. I'm trying to think if we can improve it before we make it available for everyone to use.

rudle: I did enough goofing around with generics to convince myself that they won't help us here. For posterity, here is a playground that demonstrates the two issues I see: https://go.dev/play/p/kIQdCQM1_cz

1) We cannot form a collection of concrete visitors with varied type signatures, there is no way to express their type. Go really wants us to specify a concrete type (instantiation). `interface{}` doesn't help us here, because the generic type checker expects the concrete values to have `type interface{}`
2) It's not clear how the library code (ie. exec.go) would discover what type to use when it calls the hooks. Since the hooks have concrete types, we must pass them values that type-check. We could do some risky run-time reflecty stuff here but that doesn't seem better than what we have now.

types/directive.go
```diff
@@ -24,6 +28,11 @@ type DirectiveDefinition struct {
 
 type DirectiveList []*Directive
 
+type DirectiveVisitor interface {
```
rudle, line 31: I don't think this interface actually belongs in the `types` package. This package, according to its docs, is here to map the specification into Go code 1:1. This is an implementation detail and not part of the spec.

We could move this to another public package called `directives` without losing any functionality. This would be especially valuable if we add more code to support directives in the future.

go.mod
```diff
@@ -1,9 +1,14 @@
 module github.com/graph-gophers/graphql-go
 
-go 1.13
+go 1.18
```
pavelnikolov, line 3: If we don't use generics, no need to upgrade this.

go.mod
```diff
@@ -1,9 +1,14 @@
...
 
 require (
 	github.com/opentracing/opentracing-go v1.2.0
 	go.opentelemetry.io/otel v1.6.3
 	go.opentelemetry.io/otel/trace v1.6.3
 )
+
+require (
+	github.com/go-logr/logr v1.2.3 // indirect
+	github.com/go-logr/stdr v1.2.2 // indirect
```
pavelnikolov, line 13: Why do we need these dependencies for?

go.mod
pavelnikolov:
```suggestion
go 1.13
```

types/directive.go
pavelnikolov: I agree

go.mod
```diff
@@ -7,3 +7,8 @@ require (
...
+)
```
pavelnikolov, line 14:
```suggestion
```

go.mod
rudle: https://go.dev/doc/go1.17#go-command This change in 1.17 is the likely cause of this change to go.mod

directives/visitor.go
```diff
@@ -0,0 +1,17 @@
...
+	"context"
+
+	"github.com/graph-gophers/graphql-go/types"
+)
+
+// Visitor defines the interface that clients should use to implement a Directive
+// see the graphql.DirectiveVisitors() Schema Option
+type Visitor interface {
+	// Before() is always called when the operation includes a directive matching this implementation's name
+	// Errors in Before() will prevent field resolution
```
rudle, line 13: Let's make sure this is the error behaviour we want before we merge this. This behaviour was here in the earlier version of this PR and I left it unchanged.

**Regular Method Resolvers**

Pros:
1) This lets implementations have some control over whether errors short-circuit resolution. For example, in case of an auth directive, that may be nice–graphql servers may want to abandon un-authorized requests to save CPU cycles.

Cons:
1) May cause surprising results in production. In the case of a directive with a network dependency, fields with directives may fail to resolve during a dependency outage.

It would be nice if we could provide the error status of the `Before()` hook to the resolver and provide the aggregate error state to the `After()` hook. The after hook could accept another argument `err` which it could use to decide whether or not to do its work. For the field resolver case, because we can't just change the API for resolvers, using `ctx` is our best option.

** Struct Resolvers **

I think the behaviour we have right now is the best option. Struct resolvers can't make decisions based on the error state of the directive implementation, so we should bail out for them. We could implement the After() hook for aggregate error information (before hook + field.HasError).

internal/exec/exec.go
```diff
@@ -225,7 +267,38 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
...
+							err.Path = path.toSlice()
+							err.ResolverError = visitorErr
+							return err
+						}
+					}
+				}
+			}
 			result = res.FieldByIndex(f.field.FieldIndex)
+			// After hook directive visitor (when no error is returned from resolver)
+			if !f.field.HasError && len(f.field.Directives) > 0 {
```
rudle, line 287: I don't think that struct resolver fields can ever produce an error but I'm not an expert. Shall we delete this, Pavel?

```suggestion
			if len(f.field.Directives) > 0 {
```

rudle:
@pavelnikolov thanks for the review feedback so far. I have some questions about how we should handle errors from directive visitors. Wondering if you've thought this through before.

internal/exec/exec.go
pavelnikolov: Yes, struct fields cannot produce an error. It doesn't make sense to me to check for one.

pavelnikolov:
@rudle I think that it is fine to stop execution if we have an error in the `Before` method of a directive. It makes sense to prevent further execution. We can also make it configurable.
On a separate note, I was thinking that we should make method resolver execution configurable based on logic in the `Before` directive method. Imagine a `@cache` directive which would cache fields for a certain amount of time. And if the value is present in the cache the code should not produce a query to the backend storage. Before merging we should think of practical examples, like `@auth(hasRole="editor")`, `@cache(for="10m")` (such a caching directive would be different to the commonly used `@cacheControl(maxAge="300", scope="PUBLIC")` which is used to set cache-control headers on the response - if something is found in the cache we don't hit the DB).

rudle:
>  Does that make sense or is it too much to think about for now?

I'm not opposed to adding on to this PR nor am I in a rush to have it merge it, but I don't have the time to do the requirements gathering you're describing here.

What I will do as a next action is experiment with different options for how we can pass state between `Before()`, the Method Resolver and `After()` and report back here. I said above that the `context.Context` may be our only choice, but I'd like to convince myself of that.

pavelnikolov:
> What I will do as a next action is experiment with different options for how we can pass state between Before(), the Method Resolver and After() and report back here. I said above that the context.Context may be our only choice, but I'd like to convince myself of that.
Maybe we can have a 3rd method which returns `bool` value indicating whether we need to execute the resolver method or no. Because obviously the directive is a struct and the `Before` and `After` methods can shared state either through the context or using struct fields. However, the library needs to agree on some special method, which indicates whether we need to execute the method resolver between the `Before` and `After` 🤔

rudle:
OK. A 3rd return value from `Before` makes sense to me. I wonder if it need to be a whole method? I guess that's more flexible, but at the cost of a broader API.

I just did a quick code sketch of a directive implements server-side caching via a struct field set in `Before` and a return of that value as the result of resolution in `After` and it felt pretty good to me.

Earlier today, I tried to find a nice way for `Before` to pass data into a resolver, but I didn't like any of the options I came up with. I hope we can avoid this requirement.

rudle:
above commit adds a new test case that demonstrates how to implement server-side caching. It takes the approach of adding a second return value to `Before()`. I'm open to adding a 3rd method to the Visitor interface also.

pavelnikolov:
Thank you @rudle! Happy New Year! I like the 3rd return value! This makes sense to me and looks nice.
I was looking at the [connections specification](https://relay.dev/graphql/connections.htm#sel-DABDFJAAAACLAhkV) and this example looks interesting to me. Imagine this query:

```graphql
{
  user {
    id
    name
    friends(first: 10, after: "opaqueCursor") {
      edges {
        cursor
        node {
          id
          name
        }
      }
      pageInfo {
        hasNextPage
      }
    }
  }
}
```

So imagine we want to cache the friends of a user and when we open their profile get the first 10 friends from cache. The schema declaration would look like this:

```
    ...
    friends(first: Int = 10, after: ID) @cache(ttl: Int = 300, key: String) FriendsConnection!
    ...
```
The interesting thing there is the `key`. This would be our cache key. Since we have many users and the directive is applied to each of them, how do we know which user's friends are we picking up from cache? We need to cache based on the user ID, right? So our "key" would be something like `parent.id` or something similar. I'm not sure how to resolve that problem. I found [this directive](https://www.npmjs.com/package/apollo-server-cache-directive) and it seems that they accept `parent`, `args` and `vars` cacheKeys prefixes, as in:
```
parent.<field-name>
args.<argument-name>
vars.<variable-name>
```
see [their "understanding cacheKey argument section"](https://www.npmjs.com/package/apollo-server-cache-directive#understanding-cachekey-argument).
So my final issue before merging this PR is: do we need to give the directive access to the parent object and query context (like arguments and variables). The field (method) arguments should be accessible I think, however, that is not enough since often times we should cache based on the parent ID. Chances are we have that ID in the `Before` method already. However, in order to create a fully field agnostic directive, we need to work with any type of parent ID/key. Do you think this is important?

Other than that, I am happy with this PR. I have few other thoughts:
* Maybe I should state that the directives API is not stable yet. This would give us an option to iterate on it before we call it final? Thoughts?
* I was thinking about checking out what's needed for something like Apollo Federation. This relies heavily on custom directives. I might take a look. I don't mean fully implementing it. I mean just checking if the current implementation would support such a use-case. Ideally, I don't want to bake such features in the library with special directives that we treat in some different way. It would be nice if the library would allow users to extend it and build anything they want.

graphql_test.go
```diff
@@ -74,6 +74,30 @@ func (v *customDirectiveVisitor) After(ctx context.Context, directive *types.Dir
 	return fmt.Sprintf("Directive '%s' modified result: %s", directive.Name.Name, output.(string)), nil
 }
 
+type cachedDirectiveVisitor struct {
+	cachedValue interface{}
+}
+
+func (v *cachedDirectiveVisitor) Before(ctx context.Context, directive *types.Directive, input interface{}) (bool, error) {
```
pavelnikolov, line 81: Here I feel that we need some info about the field or its parent ID so that we can get the value from cache based on the field path/args or its parent ID.

rudle:
Hello! Happy new year. Thanks for all the feedback so far. Excited that this is maybe getting close!

> Maybe I should state that the directives API is not stable yet. This would give us an option to iterate on it before we call it final? Thoughts?

Definitely agree with this. I appreciate the legwork you have already done to find some specific directives to help guide the API. The graphql-go community will no doubt have more specific feedback for this initial implementation as they try it out on their own applications. I also suspect that once this PR lands, more small contributions and improvements from the community will quickly follow. We should plan to iterate on this API and document that things are unstable.

> Another directive to consider is [a sample query complexity directive](https://github.com/slicknode/graphql-query-complexity/blob/master/src/estimators/directive/README.md). Such a directive would need access to the field args, for example:

This is supported in this PR, but in an untyped way. The 3rd argument to `Before` contains the `traceCtx` and also all args to the field. The input isn't typed, so the resolver would need to know the type of the field resolver method in order to use them. A less sophisticated approach would be: splatting them all together and calling `.String()`. Could the former approach be as easy as passing in `res.Method(f.field.MethodIndex).Type` or something like that?

> Regarding the parent object

This makes some sense to me but it seems hard to support. I don't know of any prior art for this type of thing within graphql-go. It seems to me that each field resolves in isolation of other fields and that walking up and down the graph isn't supported. It seems a bit related to this very old issue #17.

pavelnikolov:
>This makes some sense to me but it seems hard to support. I don't know of any prior art for this type of thing within graphql-go. It seems to me that each field resolves in isolation of other fields and that walking up and down the graph isn't supported. It seems a bit related to this very old issue https://github.com/graph-gophers/graphql-go/issues/17.
Indeed it is similar issue, however, I can imagine people opening issues about it. The selected fields issue is easy to solve, but it is difficult to implement in a type-safe way without changing the existing resolver API. I don't want field resolution to happen through the context if possible. This would require some other argument to the resolvers containing information about the GraphQL request such as the selected fields and maybe some other info in future.

directives/visitor.go
pavelnikolov: Yes, struct fields should be ignored. I also agree with you that the context is how the `Before` and `After` methods can communicate.

graphql_test.go
pavelnikolov: This can happen later when we implement selected fields.

graphql_test.go
```diff
@@ -329,6 +353,34 @@ func TestCustomDirective(t *testing.T) {
...
+				directive @cached(
+					key: String!
+			    ) on FIELD_DEFINITION
+
+				schema {
+					query: Query
+				}
+
+				type Query {
+					hello(full_name: String!): String! @cached(key: "notcheckedintest")
```
pavelnikolov, line 367: It would be cool if this key can come from the resolved entity's ID somehow. In a real-world situation we would need to use the ID of that entity or the ID of its parent, for example, the products of a category or the comments of a post. In both of these examples we need to cache by the ID of the parent entity. I wish we could somehow do that.

pavelnikolov:
@rudle I added a few commits to your branch. I added an example `@hasRole` directive as one of the most popular examples from the community for declarative authorization. I love the custom directive feature overall but here are some topics which might need polishing:

* We pass `ctx` to the `Before` and `After` methods. Instead I believe we need to pass the `traceCtx` this way the activity of the directive visitors will be added to the trace graphs. However, in order to do this in a protocol-agnostic way (e.g. OTEL, OpenTelemetry etc.) we need to create some kind of directive tracer interface. Then check if the provided tracer implements the directive visitor interface and wrap the `Before` and `After` methods with a trace span.
* Alternatively, since we could just pass to them [the `traceCtx`](https://github.com/graph-gophers/graphql-go/blob/master/internal/exec/exec.go#L179) and leave it to the user to start a span inside of these methods.

pavelnikolov:
I just pushed a commit which passes the `traceCtx` to the directive visitors. The users can opt-in to use that.

pavelnikolov:
I'm going to merge this PR the way it is. It already covers basic usage of field directives. I put a note in the readme that the directives API is potentially subject to change in future versions. You are correct that some of my ideas are outside of the scope of this PR. When in future it is possible to check selected fields in a resolver it will be introduced retrospectively into directives as well. Then potentially we will be able to access field arguments as well. That is why I stated that potentially the directives resolver API is subject to change.
Thank you @rudle

rudle:
Thanks @pavelnikolov! I was just checking in on this PR and am happy to see that it's merged.

I'm excited for the next steps that will make this feature more powerful, especially checking for selected fields.

pavelnikolov:
1. There is one thing that I'm not completely happy about in this PR and it's the arguments of the directives. I don't like [the way we extract them](https://github.com/graph-gophers/graphql-go/blob/master/example/directives/authorization/authorization.go#L38) now. I wish we could assign directly to a struct in a type-safe way similar to the way resolvers do it and also get validated during schema parsing. However, this would be tricky.
2. The selected fields in resolvers/directives/context is the next thing on my list. It has been the most requested feature. I've implemented it a few times already. I'm just going to start a new issue with requirements and try to make it user-friendly and without breaking changes. If you are interested - feedback/contributions would be more than appreciated.

Issue #546, opened 2023-01-06
Breaking change caused due to https://github.com/graph-gophers/graphql-go/pull/515


agnivade:
This check here: https://github.com/graph-gophers/graphql-go/pull/515/files#diff-16ec07e3ca6d409f2165fdf6a4b1db83dd05ee0f1da6d4dea1ccbe47a0ec0bdbR898 is incorrect.

We shouldn't be checking for `f >= math.SmallestNonzeroFloat64` because that excludes `0.0` which is a perfectly valid floating point number.

This is breaking our CI pipeline https://github.com/mattermost/mattermost-server/pull/21994:

```
=== RUN   TestGraphQLChannelsLeft/NotLeft
{"timestamp":"2023-01-05 09:24:50.984 Z","level":"error","msg":"Error executing request","caller":"api4/graphql.go:137","operation":"channelsLeft","errors":[{"message":"Variable \"$since\" of type \"Float\" has invalid default value 0.0.\nExpected type \"Float\", found 0.0.","locations":[{"line":2,"column":62}]}]}
{"timestamp":"2023-01-05 09:24:50.984 Z","level":"debug","msg":"Received HTTP request","caller":"web/handlers.go:171","method":"POST","url":"/api/v5/graphql","request_id":"kg91zrrckjni5xoy716ct8gggr","status_code":"200"}
    resolver_test.go:101: 
        	Error Trace:	/tmp/testlib4232044107/resolver_test.go:101
        	Error:      	"[graphql: Variable "$since" of type "Float" has invalid default value 0.0.
        	            	Expected type "Float", found 0.0. (line 2, column 62)]" should have 0 item(s), but has 1
        	Test:       	TestGraphQLChannelsLeft/NotLeft
```

I will send a PR to fix this.

cc @pavelnikolov

pavelnikolov:
PR is more than welcome @agnivade

agnivade:
Yep, busy Monday :)

PetarLotrean:
This doesn't just fix the zero, this also fixes the negative float numbers, which were for some reason invalid before this. However, this is not released yet - version 1.5.0 is still not parsing negative Float numbers. Is no one really using negative Float numbers, so this is not an issue for anyone?

pavelnikolov:
@PetarLotrean, I'm planning to release v1.6 really soon. Probably this or next week.

PR #547, opened 2023-01-10
Fix incorrect validation of zero float value

agnivade:
Float value of 0.0 is valid. Removed the incorrect
if condition. Added a test.

Fixes https://github.com/graph-gophers/graphql-go/issues/546


agnivade:
cc @pavelnikolov

PR #552, opened 2023-01-23
Improve field directives

### LINKED ISSUE #560: Improve field directives design
We have an initial version of custom directives already merged into the master branch. However, it is not ideal because it requires the user to parse arguments from the directive's argument definition list. 
That is why there is another try to improve the design before this feature is released in v1.6.0. See PR [#552](https://github.com/graph-gophers/graphql-go/pull/552)
Tasks left to do:

- ✔ Add executable examples demonstrating the `Directives` schema option
- Add more unit tests 
  - ✔ test for repeatable directives
  - ✔ test adding custom directive alongside `@deprecated` or `@skip`/`@include`
  - multiple custom directives on the same field
- Stretch goal - add validation directives, which are getting evaluated during query validation and don't even try to execute a query in case of validation errors

dackroyd:
This PR redesigns field directives and makes them more like a wrapping middleware for resolvers. 
Fixes #560

### DIFF
--- a/example/directives/authorization/README.md
+++ b/example/directives/authorization/README.md
@@ -92,7 +92,7 @@ $ curl 'http://localhost:8080/query' \
 5. Pay attention to the schmema options. Directive visitors are added as schema option:
     ```go
         opts := []graphql.SchemaOpt{
-            graphql.Directives(map[string]directives.DirectiveVisitor{
+            graphql.Directives(map[string]interface{}{
                 "hasRole": &authorization.HasRoleDirective{},
             }),
             // other options go here
--- a/example/directives/authorization/README.md
+++ b/example/directives/authorization/README.md
@@ -50,7 +50,7 @@ $ curl 'http://localhost:8080/query' \
         }
     ```
 
-3. Define a user Go type which can have a slice of roles where each role is a string:
+3. Define a user Go type which can be assigned to different roles where each role is a string:
     ```go
     type User struct {
         ID    string
@@ -70,32 +70,29 @@ $ curl 'http://localhost:8080/query' \
     }
     ```
 
-4. Define a Go type which implements the DirevtiveVisitor interface:
+4. Define a Go type which implements the ResolverVisitor interface:
     ```go
-    type HasRoleDirective struct{}
-
-    func (h *HasRoleDirective) Before(ctx context.Context, directive *types.Directive, input interface{}) (bool, error) {
-        u, ok := user.FromContext(ctx)
-        if !ok {
-            return true, fmt.Errorf("user not provided in cotext")
-        }
-        role := strings.ToLower((directive.Arguments.MustGet("role").String())
-        if !u.HasRole(role) {
-            return true, fmt.Errorf("access denied, %q role required", role)
-        }
-        return false, nil
+    type HasRoleDirective struct{
+        Role string
     }
 
-    // After is a no-op and returns the output unchanged.
-    func (h *HasRoleDirective) After(ctx context.Context, directive *types.Directive, output interface{}) (interface{}, error) {
-        return output, nil
+    func (h *HasRoleDirective) Resolve(ctx context.Context, args interface{}, next directives.Resolver) (output interface{}, err error) {
+        u, ok := user.FromContext(ctx)
+        if !ok {
+            return nil, fmt.Errorf("user not provided in context")
+        }
+        role := strings.ToLower(h.Role)
+        if !u.HasRole(role) {
+            return nil, fmt.Errorf("access denied, %q role required", role)
+        }
+        return next.Resolve(ctx, args)
     }
     ```
 
 5. Pay attention to the schmema options. Directive visitors are added as schema option:
     ```go
         opts := []graphql.SchemaOpt{
-            graphql.DirectiveVisitors(map[string]directives.Visitor{
+            graphql.Directives(map[string]directives.DirectiveVisitor{
                 "hasRole": &authorization.HasRoleDirective{},
             }),
             // other options go here
--- a/example/directives/authorization/README.md
+++ b/example/directives/authorization/README.md
@@ -76,8 +76,8 @@ $ curl 'http://localhost:8080/query' \
         Role string
     }
 
-   func (h *HasRoleDirective) ImplementsDirective(name string) bool {
-       return "hasRole" == name
+   func (h *HasRoleDirective) ImplementsDirective() string {
+       return "hasRole"
    }
 
     func (h *HasRoleDirective) Resolve(ctx context.Context, args interface{}, next directives.Resolver) (output interface{}, err error) {
--- a/example/directives/authorization/README.md
+++ b/example/directives/authorization/README.md
@@ -70,11 +70,15 @@ $ curl 'http://localhost:8080/query' \
     }
     ```
 
-4. Define a Go type which implements the ResolverVisitor interface:
+4. Define a Go type which implements the `directives.Directive` interface:
     ```go
     type HasRoleDirective struct{
         Role string
     }
+
+   func (h *HasRoleDirective) ImplementsDirective(name string) bool {
+       return "hasRole" == name
+   }
 
     func (h *HasRoleDirective) Resolve(ctx context.Context, args interface{}, next directives.Resolver) (output interface{}, err error) {
         u, ok := user.FromContext(ctx)
@@ -92,9 +96,10 @@ $ curl 'http://localhost:8080/query' \
 5. Pay attention to the schmema options. Directive visitors are added as schema option:
     ```go
         opts := []graphql.SchemaOpt{
-            graphql.Directives(map[string]interface{}{
-                "hasRole": &authorization.HasRoleDirective{},
-            }),
+            graphql.Directives(
+                &authorization.HasRoleDirective{},
+                // additional directives
+            ),
             // other options go here
         }
         schema := graphql.MustParseSchema(authorization.Schema, &authorization.Resolver{}, opts...)
--- a/example/directives/authorization/authorization.go
+++ b/example/directives/authorization/authorization.go
@@ -32,6 +32,10 @@ type HasRoleDirective struct {
 	Role string
 }
 
+func (h *HasRoleDirective) ImplementsDirective(name string) bool {
+	return "hasRole" == name
+}
+
 func (h *HasRoleDirective) Resolve(ctx context.Context, args interface{}, next directives.Resolver) (output interface{}, err error) {
 	u, ok := user.FromContext(ctx)
 	if !ok {
--- a/example/directives/authorization/authorization.go
+++ b/example/directives/authorization/authorization.go
@@ -6,8 +6,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/graph-gophers/graphql-go/directives"
 	"github.com/graph-gophers/graphql-go/example/directives/authorization/user"
-	"github.com/graph-gophers/graphql-go/types"
 )
 
 const Schema = `
@@ -28,22 +28,21 @@ const Schema = `
 	}
 `
 
-type HasRoleDirective struct{}
-
-func (h *HasRoleDirective) Before(ctx context.Context, directive *types.Directive, input interface{}) (bool, error) {
-	u, ok := user.FromContext(ctx)
-	if !ok {
-		return true, fmt.Errorf("user not provided in cotext")
-	}
-	role := strings.ToLower(directive.Arguments.MustGet("role").String())
-	if !u.HasRole(role) {
-		return true, fmt.Errorf("access denied, %q role required", role)
-	}
-	return false, nil
+type HasRoleDirective struct {
+	Role string
 }
 
-func (h *HasRoleDirective) After(ctx context.Context, directive *types.Directive, output interface{}) (interface{}, error) {
-	return output, nil
+func (h *HasRoleDirective) Resolve(ctx context.Context, args interface{}, next directives.Resolver) (output interface{}, err error) {
+	u, ok := user.FromContext(ctx)
+	if !ok {
+		return nil, fmt.Errorf("user not provided in cotext")
+	}
+	role := strings.ToLower(h.Role)
+	if !u.HasRole(role) {
+		return nil, fmt.Errorf("access denied, %q role required", role)
+	}
+
+	return next.Resolve(ctx, args)
 }
 
 type Resolver struct{}
--- a/example/directives/authorization/authorization.go
+++ b/example/directives/authorization/authorization.go
@@ -32,8 +32,8 @@ type HasRoleDirective struct {
 	Role string
 }
 
-func (h *HasRoleDirective) ImplementsDirective(name string) bool {
-	return "hasRole" == name
+func (h *HasRoleDirective) ImplementsDirective() string {
+	return "hasRole"
 }
 
 func (h *HasRoleDirective) Resolve(ctx context.Context, args interface{}, next directives.Resolver) (output interface{}, err error) {
--- a/example_directives_test.go
+++ b/example_directives_test.go
@@ -45,7 +45,7 @@ func ExampleDirectives() {
 		}
 	`
 	opts := []graphql.SchemaOpt{
-		graphql.Directives(map[string]directives.ResolverVisitor{
+		graphql.Directives(map[string]interface{}{
 			"hasRole": &HasRoleDirective{},
 		}),
 		// other options go here
--- a/example_directives_test.go
+++ b/example_directives_test.go
@@ -18,8 +18,8 @@ type HasRoleDirective struct {
 	Role string
 }
 
-func (h *HasRoleDirective) ImplementsDirective(name string) bool {
-	return "hasRole" == name
+func (h *HasRoleDirective) ImplementsDirective() string {
+	return "hasRole"
 }
 
 func (h *HasRoleDirective) Resolve(ctx context.Context, in interface{}, next directives.Resolver) (interface{}, error) {
--- a/example_directives_test.go
+++ b/example_directives_test.go
@@ -18,6 +18,10 @@ type HasRoleDirective struct {
 	Role string
 }
 
+func (h *HasRoleDirective) ImplementsDirective(name string) bool {
+	return "hasRole" == name
+}
+
 func (h *HasRoleDirective) Resolve(ctx context.Context, in interface{}, next directives.Resolver) (interface{}, error) {
 	if ctx.Value(RoleKey) != h.Role {
 		return nil, fmt.Errorf("access deinied, role %q required", h.Role)
@@ -45,9 +49,7 @@ func ExampleDirectives() {
 		}
 	`
 	opts := []graphql.SchemaOpt{
-		graphql.Directives(map[string]interface{}{
-			"hasRole": &HasRoleDirective{},
-		}),
+		graphql.Directives(&HasRoleDirective{}),
 		// other options go here
 	}
 	schema := graphql.MustParseSchema(s, &authResolver{}, opts...)
--- a/example_directives_test.go
+++ b/example_directives_test.go
@@ -1,0 +1,101 @@
+package graphql_test
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/graph-gophers/graphql-go"
+	"github.com/graph-gophers/graphql-go/directives"
+)
+
+type roleKey string
+
+const RoleKey = roleKey("role")
+
+type HasRoleDirective struct {
+	Role string
+}
+
+func (h *HasRoleDirective) Resolve(ctx context.Context, in interface{}, next directives.Resolver) (interface{}, error) {
+	if ctx.Value(RoleKey) != h.Role {
+		return nil, fmt.Errorf("access deinied, role %q required", h.Role)
+	}
+	return next.Resolve(ctx, in)
+}
+
+type authResolver struct{}
+
+func (*authResolver) Greet(ctx context.Context, args struct{ Name string }) string {
+	return fmt.Sprintf("Hello, %s!", args.Name)
+}
+
+// ExampleDirectives demonstrates the use of the Directives schema option.
+func ExampleDirectives() {
+	s := `
+schema {
+query: Query
+}
+
+directive @hasRole(role: String!) on FIELD_DEFINITION
+
+type Query {
+greet(name: String!): String! @hasRole(role: "admin")
+}
+`
+	opts := []graphql.SchemaOpt{
+		graphql.Directives(map[string]directives.ResolverVisitor{
+			"hasRole": &HasRoleDirective{},
+		}),
+		// other options go here
+	}
+	schema := graphql.MustParseSchema(s, &authResolver{}, opts...)
+	query := `
+query {
+greet(name: "GraphQL")
+}
+`
+	cases := []struct {
+		name string
+		ctx  context.Context
+	}{
+		{
+			name: "Unauthorized",
+			ctx:  context.Background(),
+		},
+		{
+			name: "Admin user",
+			ctx:  context.WithValue(context.Background(), RoleKey, "admin"),
+		},
+	}
+	for _, c := range cases {
+		fmt.Println(c.name, "result:")
+		res := schema.Exec(c.ctx, query, "", nil)
+		enc := json.NewEncoder(os.Stdout)
+		enc.SetIndent("", "  ")
+		err := enc.Encode(res)
+		if err != nil {
+			panic(err)
+		}
+	}
+	// output:
+	// Unauthorized result:
+	// {
+	//   "errors": [
+	//     {
+	//       "message": "access deinied, role \"admin\" required",
+	//       "path": [
+	//         "greet"
+	//       ]
+	//     }
+	//   ],
+	//   "data": null
+	// }
+	// Admin user result:
+	// {
+	//   "data": {
+	//     "greet": "Hello, GraphQL!"
+	//   }
+	// }
+}
--- a/example_directives_test.go
+++ b/example_directives_test.go
@@ -34,16 +34,16 @@ func (*authResolver) Greet(ctx context.Context, args struct{ Name string }) stri
 // ExampleDirectives demonstrates the use of the Directives schema option.
 func ExampleDirectives() {
 	s := `
-schema {
-query: Query
-}
+		schema {
+			query: Query
+		}
 
-directive @hasRole(role: String!) on FIELD_DEFINITION
+		directive @hasRole(role: String!) on FIELD_DEFINITION
 
-type Query {
-greet(name: String!): String! @hasRole(role: "admin")
-}
-`
+		type Query {
+			greet(name: String!): String! @hasRole(role: "admin")
+		}
+	`
 	opts := []graphql.SchemaOpt{
 		graphql.Directives(map[string]directives.ResolverVisitor{
 			"hasRole": &HasRoleDirective{},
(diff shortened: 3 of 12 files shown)


graphql_test.go
```diff
@@ -55,19 +54,16 @@ type structFieldResolver struct {
 }
 
 type customDirectiveVisitor struct {
-	// Don't really want this... want to have the visitor be a standard struct with the directive (instance) properties
-	// In most cases, this will be created and bound to each relevant field in the schema once only
-	// Likely need some initialisation, similar to custom schema types to validate the instance on the schema?
-	directive *types.Directive
+	customAttribute *string
```
pavelnikolov, line 57: It needs to be public if we want to access it from somewhere else. But there is another issue. This directive is going to be accessed from multiple goroutines and it is going to be added to multiple fields. So this is not thread-safe 🤔

internal/exec/exec.go
```diff
@@ -250,16 +251,19 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
...
+			res, resolverErr := r.executeWithDirectives(ctx, f, nil, resolver)
 			if resolverErr != nil {
 				err := errors.Errorf("%s", resolverErr)
 				err.Path = path.toSlice()
 				err.ResolverError = resolverErr
 				if ex, ok := resolverErr.(extensionser); ok {
 					err.Extensions = ex.Extensions()
 				}
+				return err
 			}
```
pavelnikolov, line 267: Now this entire block can be taken out of the `if`/`else` statement because it is pretty much the same.

graphql_test.go
pavelnikolov: Technically speaking, GraphQL spec uses case sensitive names. However, I don't think its a good design to include the same arg with names only differing in casing. Therefore just like with method resolvers, args can be mapped to a public struct field (such that has its name starting with an uppercase letter).

pavelnikolov: I absolutely love this idea of yours, it completely hides the `types` package and makes it user-friendly, however, imagine the following schema:
```graphql
schema {
	query: Query
}

type Query {
  field1: ResultOne! @hasRole(role: USER)
  field2: ResultTwo! @hasRole(role: EDITOR)
  field3: ResultThree! @hasRole(role: ADMIN)
}
...
```
Since you are passing a different role for each instance of the directive and each field can execute in parallel with the other fields we need to make sure that we have a per-resolver instance of the arguments. In the above example each there is only one directive for every field being resolved and it means that we might end up with concurrency issues. It would be nice if we received and argument to the ResolverVisitor, which is of type:
```go
// this is used only for directive arguments
args struct{
    CustomAttribute *string
}
```
in addition to the resolver input argument.

internal/exec/exec.go
```diff
@@ -293,10 +297,9 @@ func (r *Request) executeWithDirectives(ctx context.Context, f *fieldToExec, arg
 	currResolver := resolver
 
 	for i := n - 1; i >= 0; i-- {
-		i := i
-		directive := f.field.Directives[i]
+		d := f.field.Directives[i]
 
-		if visitor, ok := r.Visitors[directive.Name.Name]; ok {
+		if visitor, ok := r.Visitors[d.Name.Name]; ok {
```
pavelnikolov, line 302: Maybe we need to instantiate per resolver instance of the visitor here instead of reusing the same one from the map.

pavelnikolov: Ah, `r` in this context is actually `*Request` this means that all visitors are reused across the schema. However, if we add the visitor instance to each `f *fieldToExec`, just like its `PackedArgs`, it might as well have a `Visitors` field. Then your idea with the type-safe arguments directly mapped to struct fields and it will be safe in a multi-threaded environment.

pavelnikolov: Also some directives are repeable For example imagine this schema:
```graphql
schema {
    ...
    mutation: Mutation
}
directive @hasRole(role: String!) repeatable ON FIELD_DEFINITION

type Mutation {
    updatePost(id: ID, patch: UpdatePostInput!): UpdatePostResult! @hasRole(role: Journalist) @hasRole(role: Editor)
    publishPost(id: ID): PublishPostResult! @hasRole(role: Editor)
    ...
}
```
In the `updatePost` resolver example the `hasRole` is even added twice to the same field with a different argument. So the users would get an error if they are not a journalist and not an editor at the same time.

internal/exec/exec.go
```diff
@@ -203,124 +203,70 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
...
-						if visitorErr != nil {
-							err := errors.Errorf("%s", visitorErr)
-							err.Path = path.toSlice()
-							err.ResolverError = visitorErr
-							return err
-						}
-						result = reflect.ValueOf(modified)
-					}
+
+			res, resolverErr := r.executeWithDirectives(ctx, f, nil, resolver)
```
pavelnikolov, line 258: `ctx` -> `traceCtx`

pavelnikolov, line 258:
```suggestion
			res, resolverErr := r.executeWithDirectives(traceCtx, f, nil, resolver)
```

graphql_test.go
pavelnikolov: Now I'm convinced that we need to use `packer.MakeStructPacker` to parse all the directive args and to make them struct fields. This logic can be reused 100%. Then we need an instance of each directive for each field. So each field will have `DirectiveVisitors` or `PackedDirectives`. Naming is to be decided. This way, we'll end up with the proper fields.

pavelnikolov:
Continuation of my comments in the PR. We need to:
1. Add a map of resolvers (e.g. map[string]directive.ResolverVisitor to the schema as an option (already done)
2. Validate each instance of directive.ResolverVisitor against the corresponding [DirectiveDefinition](https://github.com/graph-gophers/graphql-go/blob/cfa4c0057dc40cb166b57e5a534dec46a7d7b1f2/types/directive.go#L16) - to see if it has the appropriate struct fields and if they are of the correct types
3. When we start executing each field, we need to check if it has any directives and if yes, then create an instance of the corresponding directive.ResolverVisitor, then set the appropriate packer.ArgsPacker and [pack/popupate](https://github.com/graph-gophers/graphql-go/blob/cfa4c0057dc40cb166b57e5a534dec46a7d7b1f2/internal/exec/resolvable/resolvable.go#L333-L343) all the struct fields based on the args.
4. Add a new field DirectiveVisitors to the resolvable.Field with the directive and its packed args.

internal/exec/exec.go
```diff
@@ -62,6 +62,71 @@ type fieldToExec struct {
 	out      *bytes.Buffer
 }
 
+// TODO: Don't want to pass request here, but need the directive visitors exposed on the field
+func (f *fieldToExec) resolve(ctx context.Context, r *Request) (output interface{}, err error) {
```
pavelnikolov, line 66: I agree with you, don't pass the request here! The visitors should not even be a field of the `Request` struct. See how the field args are in an `ArgsPacker` of type `reflect.Value`. Calling the `Interface()` method on that returns the actual value. I believe we need "packed directive visitors as well"

pavelnikolov: The code that adds the `ArgsPacker` should create an instance of the directive visitor as well (imagine a field named `DirectivesPacker`). In your code you are reusing the same directive visitor instance for all fields that have this directive. Instead, you need to create a unique instance for every occurrence of the directive for every field. Sometimes, even more than once per field in the case of repeatable directives.

internal/exec/exec.go
```diff
@@ -201,73 +266,19 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
...
-			result = reflect.ValueOf(res)
-		} else {
-			// TODO extract out unwrapping ptr logic to a common place
-			if res.Kind() == reflect.Ptr {
-				res = res.Elem()
-			}
-
-			resolver := func(ctx context.Context, _ interface{}) (output interface{}, err error) {
-				return res.FieldByIndex(f.field.FieldIndex).Interface(), nil
+		res, resolverErr := f.resolve(ctx, r)
```
pavelnikolov, line 269: This code is much cleaner to read and reason about!

pavelnikolov: A slight performance degradation might be observed when executing a query with huge number of fields because [this function will not be inlined](https://stackoverflow.com/questions/45836981/is-it-possible-to-inline-function-containing-loop-in-golang). However, the readability significantly improved, IMHO.

graphql_test.go
dackroyd: Yeah that’s my intention here: an instance of the visitor for each directive declaration in the schema. So yes, where there are multiple of the same directive (in the schema as a whole, or on the same field), these would have separate instances of the visitor in memory, with each being setup with the args relevant to that instance.

As these are directives on the _field definition_ in the schema, a they can be instantiated once when parsing the schema, and reused across requests safely

internal/exec/exec.go
dackroyd: Moved to a receiver function on `fieldToExec`, and which removes the duplication and IMO improves the readability/scope

internal/exec/exec.go
dackroyd: Now implemented

example/directives/authorization/authorization.go
```diff
@@ -30,15 +29,15 @@ const Schema = `
 `
 
 type HasRoleDirective struct {
-	directive *types.Directive
+	Role string
```
pavelnikolov, line 32: Beautiful!

graphql.go
```diff
@@ -75,6 +76,9 @@ type Schema struct {
 	schema *types.Schema
 	res    *resolvable.Schema
 
+	// TODO: don't necessarily want to force directives.ResolverVisitor?
+	// May be acceptable to implement _one_ of the directive interfaces (e.g. add one for the parse/validate step)?
```
pavelnikolov, line 80: Potentially we can solve this problem using different interfaces altogether or implement more methods (e.g. `Type()` etc.). I like your idea about Validation directives that validate before we even try to execute a query.

internal/exec/exec.go
```diff
@@ -62,31 +61,23 @@ type fieldToExec struct {
...
-		return f.resolveField(ctx, args)
-	}
-
 	currResolver := f.resolveField
 
-	for i := n - 1; i >= 0; i-- {
-		d := f.field.Directives[i]
+	ds := f.field.PackedDirectives
+	for i := len(ds) - 1; i >= 0; i-- {
+		d := ds[i]
```
pavelnikolov, line 75: Can we just do? Probably we need to document order of execution of these directives and it'd be fine... The schema authors would have control by changing the directive order in the schema definition.
```suggestion
for _, d := range f.field.PackedDirectives {
}
```

pavelnikolov: I just realized that [maps in Go do not preserve order](https://go.dev/blog/maps#iteration-order). Should we store these in a slice in the schema? Or at least sort the keys in order?

internal/exec/resolvable/resolvable.go
```diff
@@ -363,15 +367,42 @@ func (b *execBuilder) makeFieldExec(typeName string, f *types.FieldDefinition, m
...
 	}
 
+	directivesPackers := make([]*packer.StructPacker, 0, len(f.Directives))
+	for _, d := range f.Directives {
+		n := d.Name.Name
+
+		// FIXME: doesn't have a visitor, so can't be packed
+		// This could cause issues when present with other directives
+		// because we assume every directive gets a packer
+		if n == "deprecated" {
```
pavelnikolov, line 377: Indeed `@skip`, `@include`, `@deprecated` and `@specifiedBy` are the only built-in directives. However, only the `@deprecated` directive is defined on `FIELD_DEFINITION`. I do not expect many more to be added in future. That is why, I think that it's fine to ignore it like this and and treat it in a special way.

graphql.go
```diff
@@ -32,6 +32,7 @@ func ParseSchema(schemaString string, resolver interface{}, opts ...SchemaOpt) (
 		tracer:         noop.Tracer{},
 		logger:         &log.DefaultLogger{},
 		panicHandler:   &errors.DefaultPanicHandler{},
+		directives:     map[string]directives.ResolverVisitor{},
```
pavelnikolov, line 35: I think that we might have to differentiate between different directives. Maybe directives should be a config object:
```go
type Schema struct {
    ...
    directives: &Driectives{...},
    ...
}

// Directives contains maps of directive visitors for different directive locations.
// https://spec.graphql.org/draft/#DirectiveLocation
type Directives struct {
    FieldDefinition map[string]directives.ResolverVisitor{}  // ON FIELD_DEFINITION
    Field           map[string]directives.FieldVisitor{}     // ON FIELD, FRAGMENT_SPREAD, INLINE_FRAGMENT
    Argument        map[string]directives.ArgumentResolver{} // ON ARGUMENT_DEFINITION
    Variable        map[string]directives.VariableVisitor{}  // ON VARIABLE_DEFINITION
    Schema          map[string]directives.SchemaDirective{}
    Scalar          map[string]directives.ScalarDirective{}
    // more directive locations here
}
```
With the above, if we add more fields (i.e. locations) to `schema.Directives` we wouldn't break the API backwards compatibility.
Now that there is at least one directive packer created it would pave the way for more types of directives to be implemented. For example, input validation like `@isEmail`, `@matchRegex(regex: String!)`, `@inRange(min: Int, max: Int)` etc. It makes sense to reuse validation rules. I mean that different directives would do different things based on their location. Obviously, a field definition directive wraps the field resolver, however, a schema or argument directive would do something completely different (their interfaces would differ).

internal/exec/packer/packer.go
```diff
@@ -165,7 +166,33 @@ func (b *Builder) makeNonNullPacker(schemaType types.Type, reflectType reflect.T
...
+
+	return b.makeStructPacker(p, typ)
+}
+
+func (b *Builder) MakeStructPackerArgs(values []*types.Argument, typ reflect.Type) (*StructPacker, error) {
+	p := make([]packable, len(values))
+	for i, v := range values {
+		// TODO: does a default apply here?
+		// FIXME: What should the type actually be?
+		p[i] = packable{Name: v.Name.Name, Type: &types.ScalarTypeDefinition{}}
```
pavelnikolov, line 189: Driective args are of type `types.ArgumentsDefinition` in my opinion. https://spec.graphql.org/draft/#sec-Type-System.Directives

internal/exec/resolvable/resolvable.go
```diff
@@ -61,12 +63,12 @@ func (*Object) isResolvable() {}
 func (*List) isResolvable()   {}
 func (*Scalar) isResolvable() {}
 
-func ApplyResolver(s *types.Schema, resolver interface{}) (*Schema, error) {
+func ApplyResolver(s *types.Schema, resolver interface{}, directives map[string]directives.ResolverVisitor) (*Schema, error) {
```
pavelnikolov, line 66: Again what type of directives? These are only on `FIELD_DEFINITION`. What if we add more types?

pavelnikolov: Technically, this is an internal API, so we can change it/break it.

internal/exec/selected/selected.go
```diff
@@ -234,6 +265,7 @@ func applyField(r *Request, s *resolvable.Schema, e resolvable.Resolvable, sels
 	}
 }
 
+// FIXME: move these to directive visitors?
```
pavelnikolov, line 268: These are applied on `FIELD`, `INLINE_FRAGMENT` and `FRAGMENT_SPREAD` so they are executable directives (applied by the user on the query fields). Therefore, they are properties of the request and not part of the schema directives. I wouldn't worry about them for now. Also, every GraphQL server should implement them - do you want to pass them as Schema option? I'd leave these for now.

internal/exec/selected/selected.go
```diff
@@ -183,6 +192,28 @@ func applySelectionSet(r *Request, s *resolvable.Schema, e *resolvable.Object, s
...
 
+func packDirectives(fe *resolvable.Field, vars map[string]interface{}) ([]directives.ResolverVisitor, error) {
+	packed := make([]directives.ResolverVisitor, len(fe.Directives))
+	for i, d := range fe.Directives {
+		args := make(map[string]interface{})
+		for _, arg := range d.Arguments {
+			args[arg.Name.Name] = arg.Value.Deserialize(vars)
+		}
+
+		// FIXME: mismatch with 'deprecated' present as we don't add a package for that?
```
pavelnikolov, line 203: Is this comment still relevant?

pavelnikolov:
Great job, Dave! It looks much better!

pavelnikolov:
There is one issue I noticed. The directive args fields have to be pointers even if the directive args in the schema are non-nullable. For example in the `@hasRole` example or in the `@customDirective` in the tests.
For the first example try to run the `@hasRole` example locally:
1. `cd ./examples/directives/authorization/`
2. `go run ./server/server.go`
3.  it presents you with a panic:
    ```
    panic: field "Role": string is not a pointer or a nullable type
            used by (*authorization.Resolver).PrivateGreet
    
    goroutine 1 [running]:
    github.com/graph-gophers/graphql-go.MustParseSchema(...)
            /Users/pavelnikolov/go/src/github.com/graph-gophers/graphql-go/graphql.go:69
    main.main()
            /Users/pavelnikolov/go/src/github.com/graph-gophers/graphql-go/example/directives/authorization/server/server.go:22 +0x1ec
    exit status 2
    ```
4. Change the `Role` field of the `HasRoleDirective` to be a pointer to a string instead and it works after running `go server/server.go`
5. navigate to http://localhost:8080
6. try a request:
    ```graphql
    query {
        privateGreet(name: "Dave")
        publicGreet(name: "Dave")
    }
    ```
7. with header:
    ```json
    {"role": "admin"}
    ```
----
For the second example, in [this test](https://github.com/dackroyd/graphql-go/blob/59455bd8a016106dae48c2b9e8f5b1afe104e6c6/graphql_test.go#L318-L345) the custom directive struct also [uses a pointer to its arg](https://github.com/dackroyd/graphql-go/blob/59455bd8a016106dae48c2b9e8f5b1afe104e6c6/graphql_test.go#L57) while [the directive definition requires a non-nullable string](https://github.com/dackroyd/graphql-go/blob/59455bd8a016106dae48c2b9e8f5b1afe104e6c6/graphql_test.go#L321).
----
And finally, [why is this arg (i.e. `hi`) allowed to be a non-quoted string](https://github.com/dackroyd/graphql-go/blob/59455bd8a016106dae48c2b9e8f5b1afe104e6c6/graphql_test.go#L329) 🤔 ?

pavelnikolov:
I decided to give it a try and I found the bug. I put a bug line in the `makePacker` function:
```go
func (b *Builder) makePacker(schemaType types.Type, reflectType reflect.Type) (packer, error) {
	t, nonNull := unwrapNonNull(schemaType)
	fmt.Printf("DEBUG\nschema type: %+v, kind: %v\nnonNull: %v\nreflect type: %v\n", schemaType, schemaType.Kind(), nonNull, reflectType)
	if !nonNull {
```
and here is the output when I run it against the authorization example:
```
➜  authorization git:(59455bd) ✗ go run server/server.go
DEBUG
schema type: String!, kind: NON_NULL
nonNull: true
reflect type: string
DEBUG
schema type: String!, kind: NON_NULL
nonNull: true
reflect type: string
DEBUG
schema type: , kind: SCALAR
nonNull: false
reflect type: string
panic: field "Role": string is not a pointer or a nullable type
        used by (*authorization.Resolver).PrivateGreet

goroutine 1 [running]:
github.com/graph-gophers/graphql-go.MustParseSchema(...)
```
Which leads me to believe that the `schemaType` variable is not populated properly (see 3rd output).

pavelnikolov:
I found a solution and will push a fix in a moment

internal/exec/selected/selected.go
```diff
@@ -183,6 +192,28 @@ func applySelectionSet(r *Request, s *resolvable.Schema, e *resolvable.Object, s
 	return
 }
 
+func packDirectives(fe *resolvable.Field, vars map[string]interface{}) ([]directives.ResolverVisitor, error) {
+	packed := make([]directives.ResolverVisitor, len(fe.Directives))
+	for i, d := range fe.Directives {
+		args := make(map[string]interface{})
```
pavelnikolov, line 198: Maybe we should exit early and only define this variable if the number of directive args is greater than 0.

pavelnikolov:
This is still an issue:
> And finally, [why is this arg (i.e. hi) allowed to be a non-quoted string](https://github.com/dackroyd/graphql-go/blob/59455bd8a016106dae48c2b9e8f5b1afe104e6c6/graphql_test.go#L329) 🤔 ?

pavelnikolov:
Now with the latest changes the unit tests fail 🎉 :
```
testing.go:111: unexpected error: got [graphql: could not unmarshal "hi" (string) into *string: incompatible type], want []
```

internal/exec/packer/packer.go
```diff
@@ -181,17 +181,6 @@ func (b *Builder) MakeStructPacker(values []*types.InputValueDefinition, typ ref
...
-	p := make([]packable, len(values))
-	for i, v := range values {
-		// TODO: does a default apply here?
-		// FIXME: What should the type actually be?
-		p[i] = packable{Name: v.Name.Name, Type: &types.ScalarTypeDefinition{}}
-	}
-
-	return b.makeStructPacker(p, typ)
-}
-
```
pavelnikolov: You don't need this func. The one above works just fine.

internal/exec/packer/packer.go
```diff
@@ -181,17 +181,6 @@ func (b *Builder) MakeStructPacker(values []*types.InputValueDefinition, typ ref
...
-func (b *Builder) MakeStructPackerArgs(values []*types.Argument, typ reflect.Type) (*StructPacker, error) {
-	p := make([]packable, len(values))
-	for i, v := range values {
-		// TODO: does a default apply here?
-		// FIXME: What should the type actually be?
-		p[i] = packable{Name: v.Name.Name, Type: &types.ScalarTypeDefinition{}}
-	}
-
-	return b.makeStructPacker(p, typ)
-}
```
pavelnikolov: You don't need this method. The one above works just fine.

internal/exec/resolvable/resolvable.go
```diff
@@ -385,7 +385,13 @@ func (b *execBuilder) makeFieldExec(typeName string, f *types.FieldDefinition, m
...
 		r := reflect.TypeOf(v)
 
-		p, err := b.packerBuilder.MakeStructPackerArgs(d.Arguments, r)
+		// The directive definition is needed here in order to get the arguments definition list.
+		// d.Arguments wouldn't work in this case because it does not contain args type information.
+		dd, ok := b.schema.Directives[n]
+		if !ok {
+			return nil, fmt.Errorf("directive definition %q is not defined in the schema", n)
+		}
+		p, err := b.packerBuilder.MakeStructPacker(dd.Arguments, r)
```
pavelnikolov, line 394: The above change solved all the bugs from my comments.

dackroyd:
Refactoring of field resolution onto the `fieldToExec` type (receiver functions) has been split out from this work to be applied first, which should reduce the scope of this change. Those functions will still need to change to move from `Before`/`After` to the wrapping function approach applied here, but the changes should be easier to follow and more directly relate to that change, rather than _also_ moving other code around

https://github.com/graph-gophers/graphql-go/pull/568

graphql_test.go
```diff
@@ -79,6 +87,10 @@ type cachedDirectiveVisitor struct {
 	Key string
 }
 
+func (v *cachedDirectiveVisitor) ImplementsDirective(name string) bool {
+	return "cached" == name
```
pavelnikolov, line 91: doesn't it read more naturally if the variable is on the left side of the `==` operator?

internal/exec/resolvable/resolvable.go
```diff
@@ -161,10 +158,50 @@ func ApplyResolver(s *types.Schema, resolver interface{}, registeredDirectives m
...
+			return nil, fmt.Errorf("no visitors have been registered for directive %q", name)
+		}
+	}
+
+	// Validate that all the directive visitors are valid; each must implement *at least 1* of the optional functions
+	for name, v := range byName {
+		// At least 1 of the optional directive functions must be defined for each directive.
+		// For now this is the only valid directive function
+		if _, ok := v.(directives.ResolverInterceptor); ok {
+			continue
```
pavelnikolov, line 192: I want the happy path to be last. So this must change to:
```go
if !ok {
   return nil, fmt.Errorf(...)
}
```
No need to use `continue` here, please.

graphql.go
```diff
@@ -184,9 +184,9 @@ func SubscribeResolverTimeout(timeout time.Duration) SchemaOpt {
 
 // Directives defines the implementation for each directive.
 // Per the GraphQL specification, each Field Directive in the schema must have an implementation here.
-func Directives(directives map[string]interface{}) SchemaOpt {
+func Directives(ds ...directives.Directive) SchemaOpt {
```
pavelnikolov, line 187: What does the `s` in `ds` stand for? Maybe use just `d`?

example/directives/authorization/README.md
```diff
@@ -70,34 +70,36 @@ $ curl 'http://localhost:8080/query' \
...
-4. Define a Go type which implements the DirevtiveVisitor interface:
+4. Define a Go type which implements the `directives.Directive` interface:
     ```go
-    type HasRoleDirective struct{}
+    type HasRoleDirective struct{
+        Role string
+    }
+
+   func (h *HasRoleDirective) ImplementsDirective(name string) bool {
+       return "hasRole" == name
```
pavelnikolov, line 80:
```suggestion
       return name == "hasRole"
```

example_directives_test.go
```diff
@@ -0,0 +1,103 @@
...
+type roleKey string
+
+const RoleKey = roleKey("role")
+
+type HasRoleDirective struct {
+	Role string
+}
+
+func (h *HasRoleDirective) ImplementsDirective(name string) bool {
+	return "hasRole" == name
```
pavelnikolov, line 22:
```suggestion
	return name == "hasRole"
```

graphql_test.go
```diff
@@ -55,38 +54,46 @@ type structFieldResolver struct {
...
-	cachedValue interface{}
+	Key string
 }
 
-func (v *cachedDirectiveVisitor) Before(ctx context.Context, directive *types.Directive, input interface{}) (bool, error) {
-	s := "valueFromCache"
-	v.cachedValue = s
-	return true, nil
+func (v *cachedDirectiveVisitor) ImplementsDirective(name string) bool {
+	return "cached" == name
```
pavelnikolov, line 91:
```suggestion
	return name == "cached"
```

internal/exec/resolvable/resolvable.go
dackroyd: The intent here was to handle each of the supported interface types, so break early on mismatch wasn't going to be possible once validation directives are added (unless I'm misunderstanding an alternative here)

graphql_test.go
```diff
@@ -287,6 +294,35 @@ func TestHelloWorldStructFieldResolver(t *testing.T) {
 	})
 }
 
+type wrapDirective struct {
+	Prefix string
+	Suffix string
+}
+
+func (w *wrapDirective) ImplementsDirective(name string) bool {
+	return "wrap" == name
```
pavelnikolov, line 303:
```suggestion
	return name == "wrap"
```

graphql.go
dackroyd: Was intended be plural, i.e. `d`irective`s`, but in the context `d` would be fine

internal/exec/resolvable/resolvable.go
pavelnikolov: One way to achieve this would be:
```go
	for name, v := range byName {
		if _, ok := v.(directives.ResolverInterceptor); !ok {
			return nil, fmt.Errorf("directive %q (implemented by %T) does not implement a valid directive visitor function", name, v)
		}
	}

	return byName, nil
```
note that `continue` is not used in the above example. 
Another approach would be:
```go
	for name, v := range byName {
		// At least 1 of the optional directive functions must be defined for each directive.
		// For now this is the only valid directive function
		switch t := v.(type) {
		case directives.ResolverInterceptor:
		// case directives.Validator:
		default:
			return nil, fmt.Errorf("directive %q (implemented by %T) does not implement a valid directive visitor function", name, v)
		}
	}

	return byName, nil
```

PR #182, opened 2018-03-15
Separate resolvers for queries and mutations

sanae10001:
See https://github.com/graph-gophers/graphql-go/issues/145

PR #561, opened 2023-02-09
Separate resolvers for each operation

### LINKED ISSUE #145: Separate resolvers for queries and mutations

### LINKED ISSUE #562: Split resolvers for different operations
## Problem
Currently it is not possible to have field with the same name in different GraphQL operations since all resolvers live under the same root resolver and this would result in name collision. Duplicate of #145 .

## Solution
Split resolvers for each operation to allow fields with the same name in different operations.
Imagine a schema SDL like this:
```graphql
schema {
    query: Query
    mutation: Mutation
    subscription: Subscription
}

type Query {
    hello: String!
}

type Mutation {
    hello: String!
}

type Subscription {
    hello: HelloEvent!
}

type HelloEvent {
    msg: String!
}
```
Then the resolvers would look like this:
```go
type RootResolver struct{}
type QueryResolver struct{}
type MutationResolver struct{}
type SubscriptionResolver struct {
	err      error
	upstream <-chan *helloEventResolver
}

func (r *RootResolver) Query() *QueryResolver {
	return &QueryResolver{}
}

func (r *RootResolver) Mutation() *MutationResolver {
	return &MutationResolver{}
}

type helloEventResolver struct {
	msg string
	err error
}

func (r *helloEventResolver) Msg() (string, error) {
	return r.msg, r.err
}

func (r *RootResolver) Subscription() *SubscriptionResolver {
	sr := &SubscriptionResolver{}
	// populate sr with data here
	return sr
}

func (qr *QueryResolver) Hello() string {
	return "Hello world!"
}

func (mr *MutationResolver) Hello() string {
	return "Hello world!"
}

func (sr *SubscriptionResolver) Hello(ctx context.Context) (chan *helloEventResolver, error) {
	c := make(chan *helloEventResolver)
	go func() {
		for r := range sr.upstream {
			select {
			case <-ctx.Done():
				close(c)
				return
			case c <- r:
			}
		}
		close(c)
	}()

	return c, nil
}
```
Note that all three operations (i.e. query, mutation and subscription) have a field named `hello`. Therefore, the corresponding Go struct resolvers have a method named `Hello()`.

pavelnikolov:
This PR allows having fields in the schema with the same name in different operations. For example, consider the schema below:
```graphql
schema {
	query: Query
	mutation: Mutation
	subscription: Subscription
}

type Query {
	hello: String!
}

type Mutation {
	hello: String!
}

type Subscription {
	hello: HelloEvent!
}

type HelloEvent {
	msg: String!
}
```
Before this PR it was not possible to have more than one operation with the same field name. This PR resolves the problem in a backwards compatible way.
Continuation from #182 
Fixes #145
Fixes #562

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,50 @@ func (r *helloWorldResolver) Hello(ctx context.Context) (string, error) {
 }
 ```
 
+### Separate resolvers for different operations
+
+The GraphQL specification allows for fields with the same name defined in different query types. For example, the schema below is a valid schema definition:
+```graphql
+schema {
+  query: Query
+  mutation: Mutation
+}
+
+type Query {
+  hello: String!
+}
+
+type Mutation {
+  hello: String!
+}
+```
+The above schema would result in name collision if we use a single resolver struct because fields from both operations correspond to methods in the root resolver (the same Go struct). In order to resolve this issue, the library allows resolvers for query, mutation and subscription operations to be separated using the `Query`, `Mutation` and `Subscription` methods of the root resolver. These special methods are optional and if defined return the resolver for each opeartion. For example, the following is a resolver corresponding to the schema definition above. Note that there is a field named `hello` in both the query and the mutation definitions:
+
+```go
+type RootResolver struct{}
+type QueryResolver struct{}
+type MutationResolver struct{}
+
+func(r *RootResolver) Query() *QueryResolver {
+  return &QueryResolver{}
+}
+
+func(r *RootResolver) Mutation() *MutationResolver {
+  return &MutationResolver{}
+}
+
+func (*QueryResolver) Hello() string {
+	return "Hello query!"
+}
+
+func (*MutationResolver) Hello() string {
+	return "Hello mutation!"
+}
+
+schema := graphql.MustParseSchema(sdl, &RootResolver{}, nil)
+...
+```
+
 ### Schema Options
 
 - `UseStringDescriptions()` enables the usage of double quoted and triple quoted. When this is not enabled, comments are parsed as descriptions instead.
--- a/example/starwars/starwars.go
+++ b/example/starwars/starwars.go
@@ -286,14 +286,20 @@ var reviews = make(map[string][]*review)
 
 type Resolver struct{}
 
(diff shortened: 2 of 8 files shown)


nicksnyder:
You can update the PR description to say this fixes https://github.com/graph-gophers/graphql-go/issues/145

PR #563, opened 2023-02-11
Fix lint errors

pavelnikolov:


.golangci.yml
```diff
@@ -14,17 +14,14 @@ linters-settings:
...
     - gofmt
     - gosimple
     - govet
     - ineffassign
     - exportloopref
-    - structcheck
     - staticcheck
     - unconvert
     - unused
-    - varcheck
```
pavelnikolov, line 27: These 3 rules have been deprecated after v1.49.0. Now they are handled by the `unused` rule.

Issue #570, opened 2023-02-14
Add validator directives


pavelnikolov:
Add validator directives on `FIELD_DEFINITION`. These directives are executed during the query validation phase and if one of them returns an error the entire query fails and no resolver gets executed.

PR #583, opened 2023-02-18
Add Validating Directives

dackroyd:
Adding support for directives which are evaluated prior to executing any resolvers. This allows validation to be performed on the request and prevent it from executing any significant work by rejecting the request early.

The most obvious case for this is authorization: based on the requested fields, we can tell whether the request is valid given the current user, and reject the entire request. If that were applied at resolution time, the request would have partially resolved, only to return errors for the specific fields which are not authorized.

fixes: #570

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -12,6 +13,7 @@ import (
 	"github.com/graph-gophers/graphql-go"
 	"github.com/graph-gophers/graphql-go/directives"
 	gqlerrors "github.com/graph-gophers/graphql-go/errors"
+	"github.com/graph-gophers/graphql-go/example/social"
 	"github.com/graph-gophers/graphql-go/example/starwars"
 	"github.com/graph-gophers/graphql-go/gqltesting"
 	"github.com/graph-gophers/graphql-go/introspection"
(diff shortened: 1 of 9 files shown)


internal/exec/validate.go
```diff
@@ -0,0 +1,108 @@
...
+
+		for i, err := range vErrs {
+			var ext map[string]interface{}
+			if e, ok := err.(extensionser); ok {
+				ext = e.Extensions()
+			}
+			errs[i] = &errors.QueryError{
+				Err:        err,
+				Message:    err.Error(),
+				Locations:  []errors.Location{f.field.Loc}, // TODO: is this the correct location?
```
pavelnikolov, line 60: LGTM

internal/exec/validate.go
```diff
@@ -0,0 +1,108 @@
...
+		return validateSelections(ctx, sels, path, s)
+	}
+
+	switch t := t.(type) {
+	case *types.List:
+		return validateList(ctx, sels, t, path, s)
+	case *types.ScalarTypeDefinition:
+		// TODO: nothing needed here?
+	case *types.EnumTypeDefinition:
+		// TODO: nothing needed here?
```
pavelnikolov, line 86: If this is the type of the field, then we still need to validate it, right? Although, I don't believe it is possible to have `EnumTypeDefinition` here. More like `EnumValue` if I am not mistaken. Easiest way to validate this is to add a unit tests where a field is of type custom scalar and another one where the result is of type enum. I am a bit confused about this switch case.

dackroyd: Much of the validateX functions here are copied from the equivalent execX function (and adjusted to validate). These types are all handled there.

The types being dealt with here are the _schema_ types for the query selection I believe?

internal/exec/validate.go
dackroyd: I should’ve added more context in the comment here: I’m unclear whether the location should be where the field is defined in the _schema definition_, or where in the _requested query_ the field is

internal/exec/resolvable/resolvable.go
```diff
@@ -241,9 +265,12 @@ func buildDirectivePackers(s *types.Schema, visitors map[string]directives.Direc
 			continue
 		}
 
-		if _, ok = v.(directives.ResolverInterceptor); !ok {
+		switch v.(type) {
+		default:
```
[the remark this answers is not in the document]
dackroyd: Went with default first here since it’s a break early not a success flow, but will change it back to default last (which I had originally before I reordered)

internal/exec/resolvable/resolvable.go
```diff
@@ -47,26 +47,32 @@ type Field struct {
 	HasContext        bool
 	HasError          bool
 	ArgsPacker        *packer.StructPacker
-	DirectiveVisitors []directives.ResolverInterceptor
+	DirectiveVisitors *FieldDirectives
```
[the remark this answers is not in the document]
dackroyd: Will have to think on the naming here. Intent is that they are the directives applied to the schema field. In the `resolvable` package, the schema field is referred to as `Field`

internal/exec/resolvable/resolvable.go
```diff
@@ -78,6 +84,24 @@ func (f *Field) Resolve(ctx context.Context, resolver reflect.Value, args interf
 	return wrapResolver(ctx, args)
 }
 
+func (f *Field) Validate(ctx context.Context, args interface{}) []error {
```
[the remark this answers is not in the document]
dackroyd: Like for execution of requests which could fail across multiple fields and all be reported on the response, the intention here is to report _all_ validation errors from the request. Reporting only a single error can mean playing “whack a mole” resolving 1 error only to then have the next error reported.

Validation skips going deeper and validating child fields, since something like `hasRole` should apply to that entire subtree, but errors across different parts of the tree may be reported

pavelnikolov:
I resolved conflicts in this PR.

internal/exec/validate.go
pavelnikolov: We are validating the query.

graphql_test.go
```diff
@@ -456,6 +458,205 @@ func TestCustomDirective(t *testing.T) {
...
+		},
+		{
+			Schema: graphql.MustParseSchema(
+				`directive @hasRole(role: String!) on FIELD_DEFINITION
+
+				`+strings.ReplaceAll(
+					social.Schema,
+					"role: Role!",
+					`role: Role! @hasRole(role: "ADMIN")`,
+				),
```
pavelnikolov, line 524: (nitpick) You could have used schema extension here : )

I'll add examples how to do that soon.

internal/exec/resolvable/resolvable.go
```diff
@@ -41,15 +41,20 @@ type Object struct {
...
+	HasContext  bool
+	HasError    bool
+	ArgsPacker  *packer.StructPacker
+	Visitors    *FieldVisitors
+	ValueExec   Resolvable
+	TraceLabel  string
+}
+
+type FieldVisitors struct {
+	Interceptors []directives.ResolverInterceptor
```
pavelnikolov, line 56: I wonder if we should call these `ResolverMiddleware` or `ResolverInterceptor` 🤔

internal/exec/resolvable/resolvable.go
```diff
@@ -78,6 +84,24 @@ func (f *Field) Resolve(ctx context.Context, resolver reflect.Value, args interf
...
 
+func (f *Field) Validate(ctx context.Context, args interface{}) []error {
+	d := f.Visitors
+	if d == nil {
+		// Meta schema fields don't include directives on those fields
+		return nil
+	}
+
+	var errs []error
+
```
pavelnikolov, line 95: No need for empty line here.

internal/exec/validate.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/graph-gophers/graphql-go/internal/exec/selected"
+	"github.com/graph-gophers/graphql-go/types"
+)
+
+func collectFieldsToValidate(sels []selected.Selection, s *resolvable.Schema, fields *[]*fieldToValidate, fieldByAlias map[string]*fieldToValidate) {
+	for _, sel := range sels {
+		switch sel := sel.(type) {
+		case *selected.SchemaField:
+			field, ok := fieldByAlias[sel.Alias]
+			if !ok { // validation already checked for conflict (TODO)
```
pavelnikolov, line 18: What does `// validation already checked for conflict (TODO)` mean in this context?

internal/exec/validate.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func validateFieldSelection(ctx context.Context, s *resolvable.Schema, f *fieldToValidate, path *pathSegment) []*errors.QueryError {
+	if f.field.FixedResult.IsValid() {
+		// Skip fixed result meta fields like __TypeName
+		return nil
+	}
+
+	var args interface{}
+
```
pavelnikolov, line 42: (nitpick) Empty line not needed.

internal/exec/validate.go
```diff
@@ -0,0 +1,106 @@
...
+	if f.field.ArgsPacker != nil {
+		args = f.field.PackedArgs.Interface()
+	}
+
```
pavelnikolov, line 46: Same as above

pavelnikolov:
Two important things to consider before merging this code:
1. How to pass directives to the schema. The current option to pass those using a schema option, for example:
    ```go
    		graphql.Directives(&HasRoleDirective{}, &UpperDirective{}),
    ```
    looks nice, however you could also pass these as a method of the root resolver struct. We might check if it implements an interface like this:
    ```go
    type DirectivesResolver interface {
        func() Directives() []directives.Directive
    }
    ```
    Even if the customers have a `query` or `mutation` field named `directives`, we might split the resolvers so that our root resolver can have methods for `Query()`, `Mutation()`, `Subscription()` and `Directives()`. This would allow the users to avoid name collisions. Still directives would be opt-in. This makes me think if the schema option for directives at all.
2. Should we require all directives in the schema to have a Go visitor counterpart.  There might directives that are there just for decorative purposes to be used by generators or schema validators but not by the GraphQL server at runtime. I think that the library should allow arbitrary directives to be added to the GraphQL (i.e. SDL) schema and they should be ignored if there isn't a Go visitor counterpart registered for them.

pavelnikolov:
After thinking a little bit more, directives should not be part of the resolver. Instead, they should indeed be passed as a schema option. The reason for that is the possibility to parse a schema without actually passing any resolver. This would allow us to inspect the schema and to call [Schema.AST] on it. In future we might have some directives on `schema` for example or on one of the types which would give some schema parsing instructions. Therefore, directives should not be tied to the resolver in order to make them future-proof. 
I also created a separate issue for point 2. above, so this PR is ready to merge, IMO.

internal/exec/validate.go
dackroyd: This is the location in the schema being applied here, rather than the query. I'm not sure that we have access to the other

internal/exec/validate.go
dackroyd: Copy-pasted from the execX equivalent, I believe would apply here as well, but not sure what the intended long-term resolution is

internal/exec/exec.go
```diff
@@ -54,6 +54,13 @@ func (r *Request) Execute(ctx context.Context, s *resolvable.Schema, op *ast.Ope
 		default:
 			panic("unknown query operation")
 		}
+
+		if errs := validateSelections(ctx, sels, nil, s); errs != nil {
+			r.Errs = errs
+			out.Write([]byte("null"))
+			return
+		}
```
pavelnikolov, line 62: We can add a global short circuit here. If the schema doesn't have any validation directives, then this entire block can be skipped.

internal/exec/validate.go
pavelnikolov: I guess we are checking that the alias doesn't exist as a field. This is fine I guess. I'll add a unit test to confirm that. Not related to your PR, though.

Issue #588, opened 2023-02-21
Allow directives without a Go visitor counterpart


pavelnikolov:
There might directives that are there just for decorative purposes to be used by generators or schema validators but not by the GraphQL server at runtime. Therefore, the library should allow arbitrary directives to be added to the GraphQL (i.e. SDL) schema and they should be ignored if there isn't a Go visitor counterpart registered for some of them. Currently if there is a directive without a Go visitor counterpart, for example, in this schema:
    ```graphql
    directive @awesome on FIELD_DEFINITION

    type Query {
        hello: String! @awesome
    }
    ```
    the following error is returned:
    ```
    panic: no visitors have been registered for directive "awesome" [recovered]
  	  panic: no visitors have been registered for directive "awesome"
    ```
    Maybe there needs to be a schema option to define the behavior, for example, `graphql.AllowDirectivesWithoutVisitors()` or `graphql.DisallowDirectivesWithoutVisitors()`.
    An example of a directive which might need that is the `@key` directive added on `FIELD_DEFINITION` which shows the primary key of an object. It will be used by code generation tools, other directives or a super graph. However, the server would generally ignore this directive so it doesn't need to define a visitor for it.
    Another idea to solve this problem is to require visitors for all directives except of a list of special directive exceptions, e.g. `graphql.DirectivesWithoutVisitors("key", "awesome")` etc.

PR #591, opened 2023-02-22
Allow directives without a Go visitor counterpart

pavelnikolov:
fixes: #588

### DIFF
--- a/internal/exec/resolvable/resolvable.go
+++ b/internal/exec/resolvable/resolvable.go
@@ -283,32 +283,6 @@ func applyDirectives(s *ast.Schema, visitors []directives.Directive) (map[string
 		byName[name] = v
 	}
 
-	for name, def := range s.Directives {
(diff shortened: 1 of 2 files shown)


internal/exec/resolvable/resolvable.go
```diff
@@ -283,32 +283,6 @@ func applyDirectives(s *ast.Schema, visitors []directives.Directive) (map[string
...
-		if _, ok := byName[name]; !ok {
-			if name == "include" || name == "skip" || name == "deprecated" || name == "specifiedBy" {
-				// Special case directives, ignore
-				continue
-			}
-
-			return nil, fmt.Errorf("no visitors have been registered for directive %q", name)
-		}
-	}
-
```
pavelnikolov, line 311: We might add some schema option which adds an exception list from the above rule, however, until then I'm removing this restriction.

pavelnikolov: I'm still considering if we need to restrict directive locations or not 🤔

PR #609, opened 2023-03-14
Directives: Allow Argument Validation

dackroyd:
Proof of concept for implementing directive validation for field arguments. Validation for many cases applies better on the arguments than it does against the field itself.

### DIFF
--- a/directives/visitor.go
+++ b/directives/visitor.go
@@ -28,3 +28,8 @@ type ResolverInterceptor interface {
 type Validator interface {
 	Validate(ctx context.Context, args interface{}) error
 }
+
+// InputValidator directive which executes before anything is resolved, allowing the request to be rejected.
+type InputValidator interface {
+	ValidateArg(ctx context.Context, value interface{}) error
+}
--- a/directives/visitor.go
+++ b/directives/visitor.go
@@ -10,6 +10,7 @@ import (
 //
 // See the graphql.Directives() Schema Option.
 type Directive interface {
+	AllowLocation(l string) bool
 	ImplementsDirective() string
 }
 
@@ -26,10 +27,5 @@ type ResolverInterceptor interface {
 
 // Validator directive which executes before anything is resolved, allowing the request to be rejected.
 type Validator interface {
-	Validate(ctx context.Context, args interface{}) error
-}
-
-// InputValidator directive which executes before anything is resolved, allowing the request to be rejected.
-type InputValidator interface {
-	ValidateArg(ctx context.Context, value interface{}) error
+	Validate(ctx context.Context, typ, args interface{}) error
 }
--- a/example/directives/authorization/authorization.go
+++ b/example/directives/authorization/authorization.go
@@ -31,11 +31,15 @@ type HasRoleDirective struct {
 	Role string
 }
 
+func (h *HasRoleDirective) AllowLocation(l string) bool {
+	return l == "FIELD_DEFINITION"
+}
+
 func (h *HasRoleDirective) ImplementsDirective() string {
 	return "hasRole"
 }
 
-func (h *HasRoleDirective) Validate(ctx context.Context, _ interface{}) error {
+func (h *HasRoleDirective) Validate(ctx context.Context, _, _ interface{}) error {
 	u, ok := user.FromContext(ctx)
 	if !ok {
 		return fmt.Errorf("user not provided in cotext")
--- a/example_directives_test.go
+++ b/example_directives_test.go
@@ -19,11 +19,15 @@ type HasRoleDirective struct {
 	Role string
 }
 
+func (h *HasRoleDirective) AllowLocation(l string) bool {
+	return l == "FIELD_DEFINITION"
+}
+
 func (h *HasRoleDirective) ImplementsDirective() string {
 	return "hasRole"
 }
 
-func (h *HasRoleDirective) Validate(ctx context.Context, _ interface{}) error {
+func (h *HasRoleDirective) Validate(ctx context.Context, _, _ interface{}) error {
 	if ctx.Value(RoleKey) != h.Role {
 		return fmt.Errorf("access denied, role %q required", h.Role)
 	}
@@ -44,6 +48,10 @@ func (_ *WithNullableArgumentDirective) ImplementsDirective() string {
 }
 
 type UpperDirective struct{}
+
+func (d *UpperDirective) AllowLocation(l string) bool {
+	return l == "FIELD_DEFINITION"
+}
 
 func (d *UpperDirective) ImplementsDirective() string {
 	return "upper"
--- a/example_directives_test.go
+++ b/example_directives_test.go
@@ -39,7 +39,11 @@ type WithNullableArgumentDirective struct {
 	ANullableArgument *string
 }
 
-func (_ *WithNullableArgumentDirective) Validate(_ context.Context, _ interface{}) error {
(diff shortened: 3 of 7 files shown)


graphql_test.go
```diff
@@ -654,6 +654,37 @@ func TestCustomValidatingDirective(t *testing.T) {
...
+						... on Human { id name }
+						... on Droid { id name }
+						... on Starship { id name }
+					}
+				}
+			`,
+			Variables:      map[string]interface{}{"text": "Lu"},
+			ExpectedResult: "null",
+			ExpectedErrors: []*gqlerrors.QueryError{
+				// FIXME: would be helpful to include the name of the field? The context of the type / query around that also??
```
pavelnikolov, line 684: The errors support extensions. You can include the name of the field and the name of the variable in the extensions. right?

pavelnikolov: Actually, wouldn't the path include the arg name 🤔? I have to read the spec.

internal/exec/resolvable/resolvable.go
```diff
@@ -86,7 +87,7 @@ func (f *Field) Resolve(ctx context.Context, resolver reflect.Value, args interf
 	return wrapResolver(ctx, args)
 }
 
-func (f *Field) Validate(ctx context.Context, args interface{}) []error {
+func (f *Field) Validate(ctx context.Context, rawArgs map[string]interface{}, args interface{}) []error {
```
pavelnikolov, line 90: I don't like that we pass both args and rawArgs here. Should the args validator be called before the field validator? This way the field is always called with valid args. 🤔

graphql_test.go
dackroyd: We've got the `path` which I think is the _aliased_ path to the field, while I think I meant the name of the argument(s) here, which we don't know at all. We have the directive implementation, but not what it is attached to. `search` here takes `text`, but the directive implementation doesn't see that

pavelnikolov: I agree with you on this one. Some solution will be needed for this problem.

dackroyd: Latest commit should have sorted this out, as the function takes the ast type and can get the arg name from there

directives/visitor.go
```diff
@@ -26,5 +27,5 @@ type ResolverInterceptor interface {
 
 // Validator directive which executes before anything is resolved, allowing the request to be rejected.
 type Validator interface {
```
dackroyd, line 29: Thinking that the type (& function) may need a rename before a release is made. I believe there are 3 different times where parts of the schema could be “visited” (are there more?):

* When the schema is parsed (during parsing? After successful?)
* While handling a request _before_ executing any resolvers ie validation currently, part of query complexity scoring soon
* While executing resolvers for a request ie as resolver middleware, currently applied in the interceptor types

Unsure how each would ideally be named for clarity, and for consistency with other GraphQL implementations

pavelnikolov: Yes, you are right. I think that the middleware makes sense to be called interceptors as they are definitely not visitors. A visitor implies visiting all fields in the schema or the current executable document. So I would call the existing directives interceptors and I would allow for a pluggable visitor plugin which has access to the field directives. These directives, however, would be just decorative and would not have any callback themselves. The author of the visitor should decide whether and how to use them when it makes sense.

example/directives/authorization/authorization.go
```diff
@@ -31,11 +31,15 @@ type HasRoleDirective struct {
 	Role string
 }
 
+func (h *HasRoleDirective) AllowLocation(l string) bool {
+	return l == "FIELD_DEFINITION"
```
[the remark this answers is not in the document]
pavelnikolov: I was thinking about the same thing. Currently location is part of the errors but we would need to use location in many other places as we start using AST more extensively.

dackroyd:
Nearly a year on, and life has been busy. Perhaps we can pick this back up soon @pavelnikolov ?

pavelnikolov:
I'll get back on this either this or next week.

example/directives/authorization/authorization.go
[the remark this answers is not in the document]
pavelnikolov: These definitely need to become part of ast package. Users should not repeat strings like "FIELD_DEFINITION" in their code.

example/directives/authorization/authorization.go
```diff
@@ -31,11 +31,15 @@ type HasRoleDirective struct {
...
+}
+
 func (h *HasRoleDirective) ImplementsDirective() string {
 	return "hasRole"
 }
 
-func (h *HasRoleDirective) Validate(ctx context.Context, _ interface{}) error {
+func (h *HasRoleDirective) Validate(ctx context.Context, _, _ interface{}) error {
```
pavelnikolov, line 42: Just thinking out loud - should this method be called `Validate`? Can these do something than validation? I know that as per the PR title and your intentions is to add validation, however, I'm trying just to think if this can be more general. Still don't have a good use case, though 🤔

example_directives_test.go
```diff
@@ -19,11 +19,15 @@ type HasRoleDirective struct {
 	Role string
 }
 
+func (h *HasRoleDirective) AllowLocation(l string) bool {
+	return l == "FIELD_DEFINITION"
```
pavelnikolov, line 23: As per above, this needs to be a constant, however, we need to probably export only the field definitions that are supported by directives at the moment.

graphql_test.go
```diff
@@ -85,10 +94,38 @@ func (v *customDirectiveVisitor) Resolve(ctx context.Context, args interface{},
...
+
+func (v *customArgDirectiveVisitor) AllowLocation(l string) bool {
+	return l == "ARGUMENT_DEFINITION"
+}
+
+func (v *customArgDirectiveVisitor) ImplementsDirective() string {
+	return "customDirective"
+}
+
+func (v *customArgDirectiveVisitor) Resolve(ctx context.Context, args interface{}, next directives.Resolver) (output interface{}, err error) {
```
pavelnikolov, line 109: Let's drop the `Visitor` suffix from this type.

internal/exec/resolvable/resolvable.go
```diff
@@ -86,6 +87,34 @@ func (f *Field) Resolve(ctx context.Context, resolver reflect.Value, args interf
 	return wrapResolver(ctx, args)
 }
 
+func (f *Field) ValidateArgs(ctx context.Context, args map[string]interface{}) []error {
+	d := f.Visitors
```
pavelnikolov, line 91: Still unsure about naming (thinking about `Visitors`).

internal/exec/resolvable/resolvable.go
```diff
@@ -55,8 +55,9 @@ type Field struct {
 }
 
 type FieldVisitors struct {
-	Interceptors []directives.ResolverInterceptor
-	Validators   []directives.Validator
+	ArgValidators map[string][]directives.Validator
+	Interceptors  []directives.ResolverInterceptor
+	Validators    []directives.Validator
```
pavelnikolov, line 60: I should document these 3 use cases with appropriate example tests.
* ArgValidators is easy to understand
* Validators is for things that are not argument specific, e.g. `hasRole` (applied to fields).
* I can't see an example of resolver interceptor in this PR, unless I'm missing something. Did you mean a field directive to be applied to a (non-trivial) resolver? Can you give me an example?

dackroyd:
I've opted to close this, as we've discussed taking directives in a different direction. I've started on #655 to strip out the current directives implementation, with the intention to replace it with something simpler and more flexible

Issue #429, opened 2021-01-06
Cannot Represent Optional Fields with Defaults


rudle:
# Overview

We encountered a bug in production that has uncovered a mismatch in the relationship between optional fields with defaults and the Go types used to represent them. The parser should be corrected to add support for setting defaults on optional InputValues. 

## Current State

Consider this simple schema with one object:

```graphql
schema {
  query: Query
}

type Query {
  buyIceCream(args: QueryInput): Boolean!
}

input QueryInput {
  flavor: String = "VANILLA"
}
```

and the corresponding resolver:

```go
func (r *resolver) BuyIceCream(args struct{ Args *struct{ Flavor *string } }) bool {
	return true
}
```

The parsing method will reject this schema with this error `could not unmarshal "VANILLA" (string) into *string: incompatible type`.

## Expected Behaviour

The expectation is that the parser is able to represent this schema and resolver combination and that all users of this library are able to set defaults on optional InputValues.

### Further Notes & Workarounds

The common way to work around this is to use a non-pointer type to represent the optional argument. While the parser allows this, it can lead to runtime errors if a client supplies a (legal) null value for the field. This error message in this case is `graphql: got null for non-null`.

This also hopefully explains why we'd want optional fields with defaults in the first place. Including `null` as a possible input is sometimes the best way to represent our business domain.

I hope this write-up is sufficient. I can open a PR with some failing tests as a next step if that would be helpful. I'm going to continue working on solving the issue, but I thought that having the issue described here might help someone else who is seeing the same problem.

panikgaul:
@rudle Did you tried to use `string` instead of `*string` in your corresponding resolver?

I tested this behavior now and it works perfectly with following setup:

> 	input QueryInput {
>   	flavor: String = "VANILLA"
> 	}

> func (r *Resolver) BuyIceCream(args struct{ Args *struct{ Flavor string } }) bool {
> 	if args.Args.Flavor == "VANILLA" {
> 		return true
> 	}
> 	return false
> }

I think there is no bug in this library and the issue can be closed.

tonyghita:
The issue here is that the `flavor` input field is defined as a nullable `String` type, which means `null` or `nil` is a valid value for this field, but is not representable with the current implementation of the library.

This behavior is not fixed, re-opening the issue.

EverWinter23:
@pavelnikolov Could you point me to the lines where you parse the schema and check if the field can be a nullable or a non-nullable?

[...]

PR #618, opened 2023-06-19
Support for Setting Defaults on Optional Input Values

yamad07:
Fix https://github.com/graph-gophers/graphql-go/issues/429 

I have resolved the issue that prevented the inclusion of default values in optional fields. With this change, it's now possible to specify default values for optional fields, and when null is explicitly passed, it is treated as `nil`, not the default value, within the resolver.

In order to implement this change, I modified the packer code. I added a conditional branch for when the type passed to unmarshalInput is `reflect.Pointer`, ensuring it is properly unmarshalled. If this kind of conditional branching is inappropriate according to the design intentions of the packer, please leave a comment.

Also, I've adjusted the `starwars` schema for testing this logic. If this method of testing is deemed inappropriate, please provide feedback.

### DIFF
--- a/example/starwars/introspect.json
+++ b/example/starwars/introspect.json
@@ -928,17 +928,23 @@
           {
             "args": [
               {
-                "defaultValue": null,
+                "defaultValue": "\"\"",
                 "description": null,
                 "name": "text",
                 "type": {
-                  "kind": "NON_NULL",
-                  "name": null,
-                  "ofType": {
-                    "kind": "SCALAR",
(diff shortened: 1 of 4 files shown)


pavelnikolov:
I'm closing this PR as I believe that this issue is not an issue with the library per se but rather an incorrect interpretation of the GraphQL spec. If you can show me that the GraphQL spec allows this, then I'd be happy to re-consider such a PR. 
Thank you!

PR #625, opened 2023-10-16
refactored execFieldSelection to return multiple errors on the same path if err instance of `interface { Unwrap() []error }`

savaki:
I opened this discussion, [but thou](https://github.com/graph-gophers/graphql-go/discussions/623), but thought it might be useful to provide the code change being proposed.  This allows joined errors to be returned as:

```
{
  "errors": [
    {
      "message": "You cannot do this because of X.",
      "path": [
        "field"
      ],
      "extensions": {
        "code": "failed_x"
      }
    },
    {
      "message": "You cannot do this because of Y.",
      "path": [
        "field"
      ],
      "extensions": {
        "code": "failed_y"
      }
    },
    {
      "message": "You cannot do this because of Z.",
      "path": [
        "field"
      ],
      "extensions": {
        "code": "failed_z"
      }
    }
  ],
  "data": {
    "field": null
  }
}
```

rather than being using extensions and a more nested:

```
{
  "errors": [
    {
      "message": "failed_x: You cannot do this because of X., failed_y: You cannot do this because of Y., failed_z: You cannot do this because of Z.",
      "path": [
        "field"
      ],
      "extensions": {
        "details": [
          {
            "code": "failed_x",
            "message": "You cannot do this because of X."
          },
          {
            "code": "failed_y",
            "message": "You cannot do this because of Y."
          },
          {
            "code": "failed_z",
            "message": "You cannot do this because of Z."
          }
        ]
      }
    }
  ],
  "data": {
    "field": null
  }
}
```

### DIFF
--- a/internal/exec/exec.go
+++ b/internal/exec/exec.go
@@ -201,55 +201,71 @@ func execFieldSelection(ctx context.Context, r *Request, s *resolvable.Schema, f
 	}
 
 	var result reflect.Value
-	var err *errors.QueryError
+	var errs []*errors.QueryError
 
 	traceCtx, finish := r.Tracer.TraceField(ctx, f.field.TraceLabel, f.field.TypeName, f.field.Name, !f.field.Async, f.field.Args)
 	defer func() {
-		finish(err)
+		if len(errs) > 0 {
+			finish(errs[0]) // unfortunately the signature of finish is not flexible enough to allow us to pass multiple errors
+			return
+		}
+		finish(nil)
 	}()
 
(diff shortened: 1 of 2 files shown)


pavelnikolov:
This is against the GraphQL spec, which clearly states:

> If the field returns null because of a field error which has already been added to the "errors" list in the response, the "errors" list must not be further affected. That is, only one error should be added to the errors list per field.

https://spec.graphql.org/October2021/#sel-EANRNDLAACNAn7V

PR #627, opened 2023-12-13
Add NullID nullable type

Slessi:
Following up on #210 - This adds the `NullID` type to handle the nullable `ID` case

### DIFF
--- a/nullable_types.go
+++ b/nullable_types.go
@@ -5,6 +5,36 @@ import (
 	"math"
 )
 
+// NullID is an ID that can be null. Use it in input structs to
+// differentiate a value explicitly set to null from an omitted value.
+// When the value is defined (either null or a value) Set is true.
+type NullID struct {
+	Value *string
+	Set   bool
+}
+
+func (NullID) ImplementsGraphQLType(name string) bool {
+	return name == "ID"
+}
+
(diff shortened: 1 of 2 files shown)


nullable_types.go
```diff
@@ -5,6 +5,36 @@ import (
...
+type NullID struct {
+	Value *string
+	Set   bool
+}
+
+func (NullID) ImplementsGraphQLType(name string) bool {
+	return name == "ID"
+}
+
+func (s *NullID) UnmarshalGraphQL(input interface{}) error {
```
pavelnikolov, line 20: After `s.Set = true` it would be better to reuse the UnmarshalGraphQLType from the `graphql.ID` type.
https://github.com/graph-gophers/graphql-go/blob/master/id.go#L15
What do you think?

pavelnikolov (changes requested):

nullable_types.go
Slessi: I had to replace `*string` with `*graphql.ID` to do that, LMK if you think its more correct this way

[53cca00](https://github.com/graph-gophers/graphql-go/pull/627/commits/53cca00c1f52830edcbb91e89acc563461b3ea8d)

PR #633, opened 2024-03-25
Upgrade graphql-js Testdata Process

dackroyd:
The process for pulling in test cases from graphql-js has been broken for a long time. The project has made an array of changes which prevent the existing approach from being functional, including migration to Typescript, changes to assertion helper functions and more.

With this change, the approach for integration is shifted. Rather than needing to clone the graphql-js repo, instead we pull it in as a dependency of the Node.js project under testdata. This is patched via patch-package in a postinstall hook, which causes the test functions to write the cases out, and those are collected and written to JSON.

As expected, the test cases have a significant number of updates, and the Go tests driven by them fail in their current state, so further work is required.

Parity with test files pulled in from graphql-js prior to this change is maintained. A number of test files were excluded due to failures, and others have been introduced since: these are left alone for now, as it is expected that enabling those will require larger changes to the implementation to satisfy them.

Tests that have been added manually to the tests.json file have also been abandoned here. The JSON file is generated entirely by the process defined, and additional test cases will need to be recovered and re-implemented in other ways if they are still valuable.

Ensure consistent behaviour with graphql-js, matching expected test outputs. In many cases this has been a change to quoting and formatting of identifiers, however other changes to application of rules have also been applied to satisfy they required behaviours.

Rule names exercised by these tests are updated to match what has changed in graphql-js. Other rules however have been left as-is, with none of the tests from upstream validating those. As additional tests are enabled and behaviour brought inline with that, these should be updated at that time.


internal/validation/validation_test.go
```diff
@@ -40,31 +53,47 @@ func TestValidate(t *testing.T) {
...
+		s := schema.New()
+
+		s.Directives["oneOf"] = &ast.DirectiveDefinition{
+			// graphql-js includes support for @oneOf, currently in RFC
+			// This is not available in graphql-go, nor is it expected to be unless the RFC is accepted
+			// See https://github.com/graphql/graphql-js/pull/3513 & https://github.com/graphql/graphql-spec/pull/825/
+			Name:      "oneOf",
+			Desc:      "Indicates exactly one field must be supplied and this field must not be `null`.",
+			Locations: []string{"INPUT_OBJECT"},
+		}
```
pavelnikolov, line 65: This is just a validation, IMO, and we can easily include that when I make the `query` package public and allow a middleware with access to AST before actually executing the query. This way every user can validate whatever they want without including this in the library. This can be an example added to the tests/docs.

internal/validation/validation_test.go
```diff
@@ -75,4 +104,7 @@ func sortLocations(errs []*errors.QueryError) {
 		locs := err.Locations
 		sort.Slice(locs, func(i, j int) bool { return locs[i].Before(locs[j]) })
 	}
+
+	// FIXME: Sorting errors seems necessary, as mutations being validated before queries, but order reversed in expectations?
```
pavelnikolov, line 108: Then we can't really avoid it, I guess

internal/validation/validation_test.go
```diff
@@ -40,31 +53,47 @@ func TestValidate(t *testing.T) {
...
+				if rule := strings.TrimSuffix(test.Rule, "Rule"); err.Rule == rule {
 					err.Rule = ""
 					got = append(got, err)
 				}
 			}
 			sortLocations(test.Errors)
 			sortLocations(got)
 			if !reflect.DeepEqual(test.Errors, got) {
-				t.Errorf("wrong errors\nexpected: %v\ngot:      %v", test.Errors, got)
+				t.Errorf("wrong errors for rule %q\nexpected: %v\ngot:      %v", test.Rule, test.Errors, got)
```
pavelnikolov, line 96: nitpick: this code was written long time ago and now the convention is [got before want](https://go.dev/wiki/TestComments#got-before-want).

internal/validation/validation_test.go
```diff
@@ -40,31 +53,47 @@ func TestValidate(t *testing.T) {
...
 			d, err := query.Parse(test.Query)
 			if err != nil {
-				t.Fatal(err)
+				t.Fatalf("failed to parse query: %s", err)
 			}
 			errs := validation.Validate(schemas[test.Schema], d, test.Vars, 0)
 			got := []*errors.QueryError{}
 			for _, err := range errs {
-				if err.Rule == test.Rule {
+				// graphql-js adjusted its rule naming. Preserve the existing naming to minimise changes for now
```
pavelnikolov, line 87: Before we merge the PR we should probably use the graphql-js names.

internal/validation/validation_test.go
```diff
@@ -25,6 +26,18 @@ type Test struct {
...
+		// The meta schema always includes the standard types, so this isn't applicable
+		"Validate: Known type names/references to standard scalars that are missing in schema": {},
+		// Ignore tests using experimental @stream
+		"Validate: Overlapping fields can be merged/different stream directive label":               {},
+		"Validate: Overlapping fields can be merged/different stream directive initialCount":        {},
+		"Validate: Overlapping fields can be merged/different stream directive first missing args":  {},
+		"Validate: Overlapping fields can be merged/different stream directive second missing args": {},
+		"Validate: Overlapping fields can be merged/different stream directive extra argument":      {},
+		"Validate: Overlapping fields can be merged/mix of stream and no stream":                    {},
+	}
```
pavelnikolov, line 39: It is awesome that only the stream tests are being skipped. This is expected for now.

internal/validation/validation_test.go
dackroyd: This was a note to come back to / discuss once better understood. Could be the order doesn't really matter here, and sorting before comparison is right, or could be changing the order things are evaluated for validation is needed. Will try to dig in here

internal/validation/validation_test.go
dackroyd: Happy to switch that over, but does mean the change will be larger, and if any consumers make use of the `Rule` value in errors (e.g. testing comparisons) they may be impacted

internal/validation/validation_test.go
dackroyd: Not quite as good as this might suggest... check `export.ts`. I've maintained parity with the old implementation there in terms of which tests are imported (which get added to the `tests.json`). Many test files had been excluded, and I believe additional ones have also been added to `graphql-js` since, which also aren't being imported here.

I expect getting those to behave correctly and pass the tests is going to quite a bit more work, so have left that out of this change, but will start having a go at that afterwards.

internal/validation/validation_test.go
dackroyd: Resolving this was a case of changing _when_ the directives were validated for the operations from _before_ validating the variables, to _after_. Given it was a simple enough change, I've opted to do that, and remove the sorting of the errors (sorting of locations is left as it was)

internal/validation/validation_test.go
dackroyd: Will leave this alone here. I expect that its necessary to go through all the tests in the repo to apply this convention, and makes sense to do that all at once.

internal/validation/validation_test.go
dackroyd: I've updated to use the rule names from the tests.json pulled from graphql-js tests where applicable. Since there are a number of other rules *not* validated by these tests (currently...), those have been left alone for now, and will likely need to be updated when the relevant tests are enabled.

internal/validation/validation_test.go
pavelnikolov: I'd leave it here since it validates that we can add this directive here but it doesn't enforce any behavior.

pavelnikolov:
LGTM. I believe that this is a good start and improvements and new test cases can be added incrementally. Now at least we have some automated validation test coverage. 
Do you believe that there is something important left to do for this PR?

dackroyd:
In its current state, I believe this change is ready to go. Updates pushed earlier tonight should address the outstanding issues, and ensure all tests run successfully (or are skipped)

PR #637, opened 2024-04-03
Correctly apply arg names validation rule

dackroyd:
Updating implementation of KnownArgumentNamesRule to follow the graphql-js test cases. Changes required here are to correct messaging and suggestions, along with pulling in the actual test cases.

The minimum threshold is updated for the suggestion levenshteinDistance to allow suggestions for short names. From the test cases, this allows 'if' to be suggested when 'iff' is used with the `@skip` directive.


dackroyd:
This is built on top of #636 and needs that to be reviewed(/merged?) first. That should bring the change here down to a single commit, which is fairly small

PR #654, opened 2024-10-19
Support leading pipe in union definition

camdencheek:
The GraphQL spec [supports a leading pipe operator for union type definitions](https://github.com/graphql/graphql-spec/pull/320), but the parser in graphql-go fails on schemas defined that way. 

Notably, in [prettier has started using this format in v3](https://github.com/prettier/prettier/pull/15870), so many GraphQL schemas formatted by prettier will fail when used by this library.

This PR adds support for a leading pipe in union types. I added a unit test. I looked into why this wasn't captured by the validation tests that are generated from `graphql-js`, and it seems that this language feature [isn't represented](https://github.com/graphql/graphql-js/pull/907/files) in the validation tests.

### DIFF
--- a/internal/schema/schema.go
+++ b/internal/schema/schema.go
@@ -488,6 +488,9 @@ func parseUnionDef(l *common.Lexer) *ast.Union {
 
 	union.Directives = common.ParseDirectives(l)
 	l.ConsumeToken('=')
+	if l.Peek() == '|' {
+		l.ConsumeToken('|')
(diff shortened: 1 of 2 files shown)


eseliger (approved):
nice!

keegancsmith:
cc @pavelnikolov this seems like a straight-forward change. In particular this unblocks anyone who is using prettier v3.

PR #655, opened 2024-11-10
Remove Custom Directives Support

dackroyd:
Support for custom directives is desired to allow extension of the capabilities and behaviour of the GraphQL server. The current implementation would have allowed this, however the approach has become complex and limiting. In retrospect this would be better to be done in other ways to better support the future of the library.


pavelnikolov:
@dackroyd I'm thinking about merging this PR and removing the directives for now until we settle on a better approach. This would allow me to release a bunch of other useful features in v1.6.

PR #662, opened 2025-04-21
Err skip

blessedvictim:


pavelnikolov:
What is the idea behind error skip?

PR #666, opened 2025-05-24
test: fuzz test schema exec

pavelnikolov:
Add fuzz testing for schema execution and error early for executable documents without any operations.

### DIFF
--- a/graphql.go
+++ b/graphql.go
@@ -233,6 +233,10 @@ func (s *Schema) ValidateWithVariables(queryString string, variables map[string]
 		return []*errors.QueryError{qErr}
 	}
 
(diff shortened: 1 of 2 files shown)


graphql.go
```diff
@@ -233,6 +233,10 @@ func (s *Schema) ValidateWithVariables(queryString string, variables map[string]
 		return []*errors.QueryError{qErr}
 	}
 
+	if len(doc.Operations) == 0 {
+		return []*errors.QueryError{errors.Errorf("executable document must contain at least one operation")}
+	}
```
pavelnikolov, line 238: This helps to exit early when testing with an executable document without any operations.

Issue #441, opened 2021-03-25
ObjectType, Input, Enum, and Interface Definitions should not be allowed with zero fields.


ssko1:
## Current
```graphql
type Query {}
type Foo {}
input Bar {}
interface Qux {}
enum Quux {}
```

fails the following test
```golang
func TestMustHaveOneField(t *testing.T) {
	_, err := graphql.ParseSchema("type Query {} input Bar {} interface Qux {} enum Quux {}", nil)
	if err == nil {
		t.Fatal("types, inputs, interfaces and enums must have at least one field")
	}
}
```

## Expected
Types, inputs, interfaces, and enums should be validated to ensure that there is at least one field. 

Relevant specification links
ObjectTypeDefinition validation: https://spec.graphql.org/draft/#sel-HAHZhCFBABABuBq9O
InterfaceTypeDefinition validation: https://spec.graphql.org/draft/#sel-HAHbnBFBABABxB4a
EnumTypeDefinition validation: https://spec.graphql.org/draft/#sel-HAHfVFBABAB4B_6J
InputObjectTypeDefinition validation: https://spec.graphql.org/draft/#sel-HAHhBXDBABAB6BpkU

## Environment
* tested against the `master` branch

I'd be more than happy to contribute a fix.

ssko1:
There is a PR open for this issue: https://github.com/graph-gophers/graphql-go/pull/127, but it hasn't been touched in 2 years.

pavelnikolov:
@ssko1 feel free to take it and resolve the conflicts

PR #676, opened 2025-08-22
fix: disallow empty field lists

pavelnikolov:
Make the library compliant with the spects and disallow empty fields in [objects](https://spec.graphql.org/October2021/#sec-Objects.Type-Validation), [interfaces](https://spec.graphql.org/October2021/#sec-Interfaces.Type-Validation) and [input objects](https://spec.graphql.org/October2021/#sec-Input-Objects.Type-Validation).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+[Unreleased]
+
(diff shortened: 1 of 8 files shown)


pavelnikolov:
fixes #441

PR #687, opened 2025-09-16
docs(meta): add description for Time

michaelknowles:


pavelnikolov:
The Time scalar is not predefined in the schema. As its documentation say:
```go
// Time is a custom GraphQL type to represent an instant in time. It has to be added to a schema
// via "scalar Time" since it is not a predeclared GraphQL type like "ID".
```

PR #689, opened 2025-09-27
Split schema parsing

dackroyd:
Allow the schema definition parsing to be applied independently of making it executable.

This enables multiple executable schemas to be created from a single schema parse. This also gives access to the AST without attaching the resolver immediately.

The change in type for `SchemaOpt` isn't strictly a 100% backward compatible change. It should however be non-breaking for typical cases. Breakages may exist if usages define their own equivalent type from `func(*graphql.Schema)`.


pavelnikolov:
> Allow the schema definition parsing to be applied independently of making it executable.

> This also gives access to the AST without attaching the resolver immediately.

This is already available if you pass `nil` for the resolver, isn't it? It's documented in the `ParseSchema` comment.

dackroyd:
> This is already available if you pass `nil` for the resolver, isn't it? It's documented in the `ParseSchema` comment.

It's possible to parse a schema without a resolver, yes. That result cannot then be attached to a resolver later on, without re-parsing the schema from scratch, however.

By splitting the phases, it becomes possible to parse the schema and access the AST without immediately attaching a resolver, and being able to use the already parsed schema to create any number of executable schemas. These may use identical resolvers and schema options, or use different ones.

An example case might be to apply strict depth and query length limits on a publicly exposed schema, whilst allowing larger limits to the same schema when exposed internally and/or behind authentication.

pavelnikolov:
@dackroyd, I finally found time to look into this. I'm thinking about this version #697. I copied your unit test but decided not to split the opts into separate structs. Because it creates more confusion (maybe I'm just used to the existing code...)
Let me know what you think about my implementation.

PR #750, opened 2026-05-13
fix: Resolver cycles where one of the types is a union

preng69:
Version 1.10.0 introduced a breaking change for schemas with resolver cycles where one or more of the resolved types are a union. Here is an example error message from trying to generate our schema:

`*query.LinkedObjectResolver does not resolve "ParsedObject": type assertion "CargoQuote" did not resolve to an object`

In our schema `ParsedObject` is a union type while `CargoQuote` is one of the types in the union.

This PR is meant as a minimal fix to avoid the breaking change. Most of the diff is a new unit test of a cycle-with-union schema that fails in 1.10.0 but passes when the fix is applied.

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -3903,6 +3903,108 @@ func TestTypeAssertions(t *testing.T) {
 	})
 }
 
+type unionCycleQueryResolver struct{}
+
+func (*unionCycleQueryResolver) Wrapper() *unionCycleWrapperResolver {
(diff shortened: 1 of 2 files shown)


pavelnikolov:
Thank you!

pavelnikolov:
Released https://github.com/graph-gophers/graphql-go/releases/tag/v1.10.1

preng69:
Thanks!

PR #755, opened 2026-05-19
feature: New Parse opt AllowDeprecatedUsage

preng69:
This PR adds the schema parse option `AllowDeprecatedUsage`. With this option, the server will resolve fields and
arguments that have been annotated as `@deprecated` in `.graphql`, which is the pre-v1.10.0 behavior. This PR does not change the default behavior.

The purpose of the PR is to make `graphql-go` convenient (again) for situations where a grace period is required between deprecation and removal.

I have not found specific wording in the GraphQL specification (at https://spec.graphql.org/September2025/#sec--deprecated) either way; it says neither "deprecated fields must be resolved" nor "resolving a deprecated fields must return an error". In this situation I think the reasonable, least intrusive shape of this PR is to leave current default behavior as is, while making the legacy behavior optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # CHANGELOG
 
+* [FEATURE] Add `AllowDeprecatedUsage()` schema option to restore pre-v1.10.0 validation behavior for deprecated fields, arguments, enum values, input fields, and directive arguments.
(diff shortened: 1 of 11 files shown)


pavelnikolov:
Thank you for bringing this to my attention - [released v1.10.2](https://github.com/graph-gophers/graphql-go/releases/tag/v1.10.2).

Issue #763, opened 2026-06-22
`reflect: Call using X as type Y` panic when Go interface resolver has concrete type with extra methods


janisz:
## Description

When a resolver method returns a **Go interface** (not a GraphQL `interface`/`union`) as the resolver for a regular GraphQL `type`, and the underlying concrete type has additional exported methods beyond those in the interface, field resolution panics with:

```yaml
reflect: Call using bug_test.QueryArgs as type bug_test.PaginatedArgs
```

The root cause is that `MethodIndex` values are computed against the Go interface type at schema-build time, but applied to the concrete type at runtime — and the two have different method orderings when the concrete type has extra methods.

## Versions affected

- **Last working:** v1.5.0
- **First broken:** v1.6.0
- **Still broken in:** v1.10.2 (latest)

## Introduced by

- https://github.com/graph-gophers/graphql-go/pull/568
- refactored inline resolver dispatch into a helper returning `(interface{}, error)`. This introduced an `any` round-trip that loses the Go interface type:

**v1.5.0 (working)** — `exec.go` kept the `reflect.Value` from the method call:

```go
callOut = res.Method(f.field.MethodIndex).Call(in)
result = callOut[0]  // reflect.Value preserves declared interface type
```

**v1.6.0+ (broken)** — `resolve()` returns `any`, `exec.go` re-wraps it:

```go
res, resolverErr := f.resolve(ctx)   // returns (any, error)
result = reflect.ValueOf(res)         // unwraps to concrete type
```

The critical line is [`resolvable.go:129`](https://github.com/graph-gophers/graphql-go/blob/v1.10.2/internal/exec/resolvable/resolvable.go#L129): `return result.Interface(), nil` — this converts the `reflect.Value` (which preserves the interface type) into `any` (which unwraps to the concrete type). When [`exec.go:285`](https://github.com/graph-gophers/graphql-go/blob/v1.10.2/internal/exec/exec.go#L285) then does `reflect.ValueOf(res)`, it gets `*concreteType` instead of `MyInterface`.

https://github.com/graph-gophers/graphql-go/blob/945316a96f2b309c04584a8f30f6b33422313669/internal/exec/resolvable/resolvable.go#L129
https://github.com/graph-gophers/graphql-go/blob/945316a96f2b309c04584a8f30f6b33422313669/internal/exec/resolvable/resolvable.go#L285

## Root cause

`reflect.ValueOf()` on an `any` holding an interface value always unwraps to the concrete type:

```go
[...]

PR #764, opened 2026-06-22
fix: return reflect.Value from Field.resolve to preserve Go interface

janisz:
Fixes:
- #763

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


pavelnikolov (dismissed):

pavelnikolov:
fixes: #763

PR #773, opened 2026-09-02
Fix TraceField finish never receiving the resolver's error

### LINKED ISSUE #772: Tracer.TraceField's finish callback never sees field resolver errors
I have implemented a custom `tracer.Tracer` to export errors count into Prometheus metrics like this:

```go
func (t *prometheusTracer) TraceField(ctx context.Context, label, typeName, fieldName string, trivial bool, args map[string]any) (context.Context, func(*gqlerrors.QueryError)) {
	return ctx, func(err *gqlerrors.QueryError) {
		if err != nil {
			prometheusErrorsMetric.WithLabelValues(typeName, fieldName).Inc()
		}
	}
}
```

However the `err` is always `nil`, because of following line in the library:
https://github.com/graph-gophers/graphql-go/blob/6ab2b74415f5415862b64c6a844d65bff9735ac3/internal/exec/exec.go#L251

Go evaluates a deferred call's arguments when the defer statement executes, not when the deferred call itself later runs. Finish is bound to the nil value of err from before the resolver ran.

An easy fix is to use a closure instead:
```go
defer func() { finish(err) }()
```

thaarok:
Fixes #772 - Tracer.TraceField's finish callback never sees field resolver errors

### DIFF
--- a/graphql_test.go
+++ b/graphql_test.go
@@ -6422,3 +6422,55 @@ func TestIssue763InterfaceMethodIndexInListUnion(t *testing.T) {
 		ExpectedResult: `{"search":[{"label":"label:x"}]}`,
 	})
 }
+
+type issue772ErrorRecordingTracer struct {
+	mu       sync.Mutex
+	fieldErr map[string]*gqlerrors.QueryError
+}
+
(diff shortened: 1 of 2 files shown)


pavelnikolov (approved):

pavelnikolov:
Thank you!

